PR #3, opened 2017-06-16
Generated Client with Auth code

krabhishek8260:
Generated the client with Auth boilerplate code. Will add the wrapper code to set auth from kubeconfig in next PR.


brendandburns:
blind LGTM.

PR #11, opened 2017-07-16
Added Travis-CI

sesispla:
Added Travis-CI for current netcoreapp1.1

Build example from my [Travis CI](https://travis-ci.org/sesispla/kubernetes-client-csharp)

NOTE: Current build looks broken. I thought it was something related to my machine, but it looks like an issue with dotnet-sdk-1.0.4 and netcoreapp1.1

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,11 @@
+language: csharp
+solution: tests/tests.csproj
+sudo: required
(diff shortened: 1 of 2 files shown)


brendandburns:
Thanks for this!  I will merge this and then I will try to fix the build :)

PR #13, opened 2017-07-24
Update credentials for more valid values.

brendandburns:
@sesispla @krabhishek8260 

This makes it possible to use csharp with the `minikube` tool.


src/KubeConfigModels/UserCredentials.cs
```diff
@@ -8,9 +9,15 @@ public class UserCredentials
         [YamlMember(Alias = "client-certificate-data")]
         public string ClientCertificateData { get; set; }
 
+        [YamlMember(Alias = "client-certificate")]
+        public string ClientCertificate { get; set; }
```
sesispla: Need to review unit tests. Thought that client-certifidate-data hold the route to a X509 certificate, but concluded now it is the certificate itself. Additional tests should be written to cover the client-certificate scenario.

src/KubeConfigModels/UserCredentials.cs
```diff
@@ -8,9 +9,15 @@ public class UserCredentials
...
         public string ClientCertificateData { get; set; }
 
+        [YamlMember(Alias = "client-certificate")]
+        public string ClientCertificate { get; set; }
+
         [YamlMember(Alias = "client-key-data")]
         public string ClientKeyData { get; set; }
 
+        [YamlMember(Alias = "client-key")]
+        public string ClientKey { get; set; }
```
sesispla: Same as client-certificate. Thought client-key-data stored a path, but they are the certificate itself. Now several tests are broken.

brendandburns:
@sesispla I fixed the tests. Would you like me to add new tests to this PR or as part of a later commit?

sesispla:
@brendanburns Maybe we can add some basic tests for client-certificate-data, and then create a more intensive test plan in the future. Not perfect, but it will the job for now.

For the future plan, I am looking for a Code Coverage tool in #16 which will make our life easier.

brendandburns:
@sesispla sorry for the delay. I added some tests, and then finally got them passing :)

Please take another look.

Thanks!

sesispla:
@brendanburns can't see any new commits to this PR. Maybe they are in another branch? Or are they 6333784?

sesispla:
6333784 LGTM

brendandburns:
@sesispla yeah, I squashed them down to one commit....

Thanks for the reivew!

PR #19, opened 2017-08-24
simplifying certificate chain handling

allantargino:
Hey guys,
I've recently needed to call Kube API from my C# application (just 2 methods, not all Kube API is covered like yours) and my server validation used the same X509Chain object from WebRequestHandler (it doesn't create an additional one). Here is my implementation:
```cs
private bool ServerCertificateValidationCallback(object sender, X509Certificate x509Certificate, X509Chain x509Chain, SslPolicyErrors sslPolicyErrors)
{
    if (sslPolicyErrors.Equals(SslPolicyErrors.None))
        return true;

    if ((sslPolicyErrors & SslPolicyErrors.RemoteCertificateChainErrors) > 0)
    {
        x509Chain.ChainPolicy.ExtraStore.Add(_caCertificate);
        x509Chain.ChainPolicy.VerificationFlags = X509VerificationFlags.AllowUnknownCertificateAuthority;
        return x509Chain.Build((X509Certificate2)x509Certificate);
    }

    return false;
}
```

### DIFF
--- a/src/Kubernetes.Auth.cs
+++ b/src/Kubernetes.Auth.cs
@@ -105,13 +105,12 @@
             // If there are errors in the certificate chain, look at each error to determine the cause.
             if ((sslPolicyErrors & SslPolicyErrors.RemoteCertificateChainErrors) != 0)
             {
-                X509Chain chain0 = new X509Chain();
-                chain0.ChainPolicy.RevocationMode = X509RevocationMode.NoCheck;
+                chain.ChainPolicy.RevocationMode = X509RevocationMode.NoCheck;
 
                 // add all your extra certificate chain
-                chain0.ChainPolicy.ExtraStore.Add(this.CaCert);
-                chain0.ChainPolicy.VerificationFlags = X509VerificationFlags.AllowUnknownCertificateAuthority;
-                var isValid = chain0.Build((X509Certificate2)certificate);
+                chain.ChainPolicy.ExtraStore.Add(this.CaCert);
+                chain.ChainPolicy.VerificationFlags = X509VerificationFlags.AllowUnknownCertificateAuthority;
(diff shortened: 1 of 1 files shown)


krabhishek8260:
LGTM

brendandburns:
@allantargino what are the implications (if any) of modifying the chain in the parameter? Would this install the cert globally by any chance?  Basically this could introduce side effects outside of this method, I want to make sure that it doesn't in practice...

Thanks!

allantargino:
hi @brendanburns!

When looking at the [documentation](https://msdn.microsoft.com/en-us/library/system.security.cryptography.x509certificates.x509chainpolicy.extrastore(v=vs.110).aspx), it says:

> "Protocols such as Secure Sockets Layer (SSL) and Secure MIME (S/MIME) allow the inclusion of additional certificates for use in validating a certificate chain; these certificates can prove to be useful in the validation process. This property allows these certificates to be passed to the chain validation engine. By default, this collection is empty."

It doesn't explicit says that is just an in-memory object for validation, but really looks like for me. I've monitored changes in certmgr.exe (Windows 10) if any time my kube CA root was added at some Store - it was not. Also, every time the HTTP call is made again, it also needs to validate the CA programmatically - another positive signal that ExtraStore is just a temporary object for validation.

As soon as the validation callback method is triggered, the ExtraStore only contains the "apiserver.crt" sent from Kube. So adding the CA certificate on the same chain, we do not need to create another object for validation.

This is a really simple change on the code - but I really liked the discussion here. Windows does not seem to be affected by ExtraStore. Do you think is valid to perform some investigation on Linux as well?

Thanks!

brendandburns:
@allantargino good enough for me, I'd bet the code to handle it is the same in both Linux & Windows...

LGTM, thanks for the pr.

PR #23, opened 2017-09-14
move to .NET Standard 2.0

colemickens:
Looks like the X509Certificate2 API is slightly different...

### DIFF
--- a/src/KubernetesClient.csproj
+++ b/src/KubernetesClient.csproj
@@ -1,6 +1,6 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>netstandard1.6</TargetFramework>
+    <TargetFramework>netstandard2.0</TargetFramework>
   </PropertyGroup>
   <ItemGroup>
     <Compile Remove="GlobalSuppressions.cs" />
--- a/examples/simple/PodList.cs
+++ b/examples/simple/PodList.cs
@@ -1,4 +1,4 @@
-﻿namespace simple
+namespace simple
 {
     using System;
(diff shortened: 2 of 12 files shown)


colemickens:
(It seems that .NET Standard 2.0 still doesn't seem to support non-PFX scenarios around x509)

colemickens:
When running the example, I get this:

```
Unhandled Exception: System.AggregateException: One or more errors occurred. (The handler does not support client authentication certificates with this combination of libcurl (7.55.1) and its SSL backend ("OpenSSL/1.1.0f").) ---> System.PlatformNotSupportedException: The handler does not support client authentication certificates with this combination of libcurl (7.55.1) and its SSL backend ("OpenSSL/1.1.0f").
```

colemickens:
It appears the error I've listed is unrelated to moving it to .NETStandard2 and is rather just a deficiency in corefx on Linux/OSX. If I replace my libcurl4 with one built against openssl1.0.0, then it works. Not sure how common that is among modern distros...

I think this can be merged.

src/KubernetesClient.csproj
```diff
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>netstandard1.6</TargetFramework>
+    <TargetFramework>netstandard2</TargetFramework>
```
krabhishek8260: nit: netstandard2.0, just to be consistent

colemickens:
Done. Added another commit that removes all the UTF-8 BOMs I could find. It appears VS Code also did some other small cleanups as I touched those files. Hopefully that's okay to bundle in this PR.

brendandburns:
LGTM, thanks!

PR #25, opened 2017-09-22
Fix up versions and improve config loading...

### LINKED ISSUE #24: kubeconfig handling is overly sensitive
For example, this is a valid config file that kubectl accepts, but this client can not parse into KubernetesClientCredentials:

```
apiVersion: v1
kind: Config
clusters:
- name: colemick1
  cluster:
    server: https://colemick1-api.westus.cloudapp.azure.com:443
    certificate-authority-data: <redacted>
users:
- name: kubelet
  user:
    client-certificate-data: <redacted>
    client-key-data: <redacted>
contexts:
- context:
    cluster: colemick1
    user: kubelet
```

I had to:
1. name the context
2. actually set the active context in the kubeconfig

I'm not sure it's technically a well-formed kubeconfig, but I figure if `kubectl` accepts it, that this client should too.

brendandburns:
Fixes #24 

@colemickens

### DIFF
--- a/examples/simple/PodList.cs
+++ b/examples/simple/PodList.cs
@@ -8,7 +8,11 @@ namespace simple
     {
         static void Main(string[] args)
         {
-            var k8sClientConfig = new KubernetesClientConfiguration();
+            var k8sClientConfig = KubernetesClientConfiguration.defaultConfiguration();
+            if (k8sClientConfig.CurrentContext == null) {
+                Console.WriteLine("No current context");
+                return;
+            }
             IKubernetes client = new Kubernetes(k8sClientConfig);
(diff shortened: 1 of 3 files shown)


brendandburns:
@krabhishek8260 @colemickens any chance I can get a quick LGTM?

colemickens:
1. I found it a bit weird that `getDefaultConfiguration` doesn't throw and instead the caller must check for `null`.

3. It looks like it still won't handle the kubeconfig I posted in #24, but instead will silently fall out without good credentials and avoid the NPE at least?

brendandburns:
closing this, it's out of date.

Issue #26, opened 2017-09-22
Remove the dependency of OpenSSL executable


ArchangelSDY:
Hi,

I see currently this library depends on an OpenSSL executable to do some key format conversion. We are building a web service so opening a new process seems a bit heavy. Do you have any plan to drop this dependency? Some libraries such as BouncyCastle might help.

Thanks.

brendandburns:
Until very recently, dotnet core didn't support the right features https://github.com/dotnet/corefx/issues/14284

But this is now supported in dotnet 2.0, we need to write the code...

krabhishek8260:
This has still not been resolved with dotnet core 2.0. The 2.0 release has a new method CopyWithPrivateKey(RSA privatekey) to set the private key but there is no way to create the RSA private ey object from the key file. I confirmed this with .net core team.

qmfrederik:
@krabhishek8260 Did you look at Bouncy Castle? I'm pretty sure you can use Bouncy Castle to create the RSA private key; you can then pass that info to CopyWithPrivateKey.

PR #27, opened 2017-09-23
Remove the dependency on the openssl executable.

brendandburns:
@sesispla 

Fixes https://github.com/kubernetes-client/csharp/issues/26


brendandburns:
@firazs 
1) The FxCop was cargo-cult, I think.

2) Tests added 

3) Removed.

Please take another look.

src/KubernetesClient.csproj
```diff
@@ -1,15 +1,14 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
```
krabhishek8260: Moving back to netcore from netstandard will make the library incompatible with .net.
We should target .net standard.

brendandburns:
When I target netstandard I don't see the function we need. Any ideas about how to work around that?

--brendan

src/KubernetesClient.csproj
krabhishek8260: Looks like new method CopyWithPrivateKey has not been made part of the netstandard APIs. :(

krabhishek8260: Tagging @bartonjs who can shed some more light on this.

bartonjs: CopyWithPrivateKey and the CertificateRequest API are not part of .NET Standard and do not work on .NET Framework.

krabhishek8260: Thanks Jeremy @bartonjs. Is there any plan to add it in later releases.
@brendandburns We can then target dotnetcore and the library won't work for .net :( . 
@firazs for .net we will still need to invoke openssl to create the pfx for the client cert.

bartonjs: > @bartonjs. Is there any plan to add it in later releases.

I'm not personally aware of how we're going to rev netstandard.  But once someone informs me of how it's supposed to work, I'll be arguing that this stuff belongs :smile:.  Doin' what I can, and hoping that things go fairly fast.

brendandburns:
@krabhishek8260 I think this is ready to merge, can I get an LGTM?

Once it is merged, I'll look into using reflection to make this netstandard compatible in a separate PR...

krabhishek8260:
LGTM 👍

tg123:
Hi

I found a workaround for GeneratePfx in dotstanard

```
        public static X509Certificate2 GeneratePfx(KubernetesClientConfiguration config)
        {
            var keyData = new byte[]{};
            var certData = new byte[]{};

            if (!string.IsNullOrWhiteSpace(config.ClientCertificateKey))
            {
                keyData = Convert.FromBase64String(config.ClientCertificateKey);
            }
            if (!string.IsNullOrWhiteSpace(config.ClientKey))
            {
                keyData = File.ReadAllBytes(config.ClientKey);
            }

            if (!string.IsNullOrWhiteSpace(config.ClientCertificateData))
            {
                certData = Convert.FromBase64String(config.ClientCertificateData);
            }
            if (!string.IsNullOrWhiteSpace(config.ClientCertificate))
            {
                certData = File.ReadAllBytes(config.ClientCertificate);
            }

            var cert = new X509CertificateParser().ReadCertificate(new MemoryStream(certData));
            object obj;

            using (var reader = new StreamReader(new MemoryStream(keyData)))
            {
                obj = new PemReader(reader).ReadObject();
                var key = obj as AsymmetricCipherKeyPair;
                if (key != null)
                {
                    var cipherKey = key;
                    obj = cipherKey.Private;
                }
            }

            var rsaKeyParams = (RsaPrivateCrtKeyParameters) obj;

            var store = new Pkcs12StoreBuilder().Build();
            store.SetKeyEntry("K8SKEY", new AsymmetricKeyEntry(rsaKeyParams), new[] {new X509CertificateEntry(cert)});

            using (var pkcs = new MemoryStream())
            {
                store.Save(pkcs, null, new SecureRandom());
                return new X509Certificate2(pkcs.ToArray());
            }
        }
``

tg123:
hi

I have added same testcases to verify https
and the code above works

https://github.com/tg123/csharp/blob/mock_svc/tests/AuthTests.cs#L156

I will cherry pick my testcases to a separate PR after this got merged

brendandburns:
@tg123 cool!  this is merged, I'd love to see your workaround get in.

krabhishek8260:
@tg123 Is the Pkcs12Store secure and does this approach have any security implications?

tg123:
Hi @krabhishek8260 

`Pkcs12Store` is something works like `openssl` command. which just generate the `pfx` in memory instead of writing it to disk.

since the input already contains private key in memory, I think the security level would be no worse and better than put the `pfx` on the disk.

the memory (byte[]) of the generated `pfx` is totally destroyed after `X509Certificate2` was loaded.

PR #28, opened 2017-09-27
Correct the content type in the Kubernetes client.

brendandburns:
This is a temporary fix, we need a fix in upstream swagger, AutoRest or post-processing.


krabhishek8260:
With this change too the library is still not usable for POST calls without fixing https://github.com/kubernetes-client/csharp/issues/9

brendandburns:
I did a kubernetes PR, so the code is there now:

https://raw.githubusercontent.com/kubernetes/kubernetes/master/api/openapi-spec/swagger.json

We need to re-run AutoRest to regenerate.

krabhishek8260:
Great. I will try to rerun AutoRest and regenerate the client.
Thanks!

krabhishek8260:
The generator Docker image does not have autorest. I added autorest to the Dockerfile but I have not been able to successfully generate the client. I am talking to the autorest team to find out what's wrong.

PR #31, opened 2017-10-11
update test sdk to release version

tg123:
1. update xunit to stable version
1. change ci from `dotnet xunit` to more general way `dotnet test`

some of those changes are already in #27, but this xunit version is newer


brendandburns:
Looks great, but needs a rebase.

Thanks!

Issue #32, opened 2017-10-11
Support lower runtime version netstandard 1.0+ netframework 4.0


tg123:
Hi

Is there a plan that this library would support lower dotnet runtime? like:
*  netstandard 1.0
* netframework 4.0

brendandburns:
I don't know what is needed? I'm in favor of supporting as compatible a standard as we can as long as we can get the library to do what we need.

tg123:
We want to run this client on `.net45`. Benefit from `netstandard1.3`, it now can target on `net46`. 

This issue wants to define the minimal version of netstandard version and netframework version this library support.

in other words: will this library support `net45` or any other lower versions

felixfbecker:
I would like to use this with targeting .NET Standard 2.0
```
error: Package KubernetesClient 0.1.0-beta is not compatible with netstandard2.0 (.NETStandard,Version=v2.0). Package KubernetesClient 0.1.0-beta supports: netcoreapp1.1 (.NETCoreApp,Version=v1.1)
```

tg123:
please update to higher version > 0.3.0-beta

felixfbecker:
Ah, that fixed it. Thanks.

PR #33, opened 2017-10-12
change lib target to netstandard13

tg123:
1. make lib work in netstandard 1.3 #32 
1. add root namespace


tg123:
seems need to test on mac.

krabhishek8260:
We have to stick to dotnet core 2.0 for now because the cert apis to set private key is only avaibale in dotnetcore 2.0.

tg123:
hi @krabhishek8260 
I have found the dotnet standard way to generate the cert see 8e9ed00fc3d4917cbf841e3119101035fe87469a and #27

krabhishek8260:
LGTM. This is great. Thanks!

PR #35, opened 2017-10-16
introduce watch api

### LINKED ISSUE #8: Support watch calls

tg123:
fix #8 

The API is like

```
var listTask = client.ListNamespacedPodWithHttpMessagesAsync("default", watch: true).Result;
listTask.Watch<Corev1Pod>((type, item) =>
{
    Console.WriteLine("==on watch event==");
    Console.WriteLine(type);
    Console.WriteLine(item.Metadata.Name);
    Console.WriteLine("==on watch event==");
});
```

since autorest generated code use `ReadAsString` to receive the data from api server, json stream would be hang.

I tried to hack autorest first, but it is hard to change the template for k8s client.

The implementation used an HttpDelegatingHandler to rewrite the response and return first line to autorest client, then use WatchExt to create a watch object which interact with the replaced http response to get watch works.

### DIFF
--- a/src/Watcher.cs
+++ b/src/Watcher.cs
@@ -22,8 +22,10 @@ namespace k8s
 
     public class Watcher<T> : IDisposable
     {
-        private readonly StreamReader _streamReader;
+        public bool Watching { get; private set; }
+
         private readonly CancellationTokenSource _cts;
+        private readonly StreamReader _streamReader;
 
         public Watcher(StreamReader streamReader, Action<WatchEventType, T> onEvent, Action<Exception> onError)
         {
@@ -37,24 +39,39 @@ namespace k8s
 
             Task.Run(async () =>
             {
-                while (!streamReader.EndOfStream)
+                try
                 {
-                    if (token.IsCancellationRequested)
-                    {
-                        return;
-                    }
+                    Watching = true;
 
-                    try
+                    while (!streamReader.EndOfStream)
                     {
+                        if (token.IsCancellationRequested)
+                        {
+                            return;
+                        }
+
                         var line = await streamReader.ReadLineAsync();
-                        var @event = SafeJsonConvert.DeserializeObject<WatchEvent>(line);
 
-                        OnEvent?.Invoke(@event.Type, @event.Object);
+                        try
+                        {
+                            var @event = SafeJsonConvert.DeserializeObject<WatchEvent>(line);
+                            OnEvent?.Invoke(@event.Type, @event.Object);
+                        }
+                        catch (Exception e)
+                        {
+                            // error if deserialized failed or onevent throws
+                            OnError?.Invoke(e);
+                        }
                     }
-                    catch (Exception e)
-                    {
-                        OnError?.Invoke(e);
-                    }
+                }
+                catch (Exception e)
+                {
+                    // error when transport error, IOException ect
+                    OnError?.Invoke(e);
+                }
+                finally
+                {
+                    Watching = false;
                 }
             }, token);
         }
--- a/src/Watcher.cs
+++ b/src/Watcher.cs
@@ -22,12 +22,15 @@ namespace k8s
 
     public class Watcher<T> : IDisposable
     {
+        /// <summary>
+        /// indicate if the watch object is alive
+        /// </summary>
         public bool Watching { get; private set; }
 
         private readonly CancellationTokenSource _cts;
         private readonly StreamReader _streamReader;
 
-        public Watcher(StreamReader streamReader, Action<WatchEventType, T> onEvent, Action<Exception> onError)
+        internal Watcher(StreamReader streamReader, Action<WatchEventType, T> onEvent, Action<Exception> onError)
         {
             _streamReader = streamReader;
             OnEvent += onEvent;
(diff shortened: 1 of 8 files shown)


krabhishek8260:
Let me know if you need some help with the k8s client template to make this less hacky.

src/Watcher.cs
```diff
@@ -0,0 +1,100 @@
...
+        [EnumMember(Value = "MODIFIED")] Modified,
+
+        [EnumMember(Value = "DELETED")] Deleted,
+
+        [EnumMember(Value = "ERROR")] Error
+    }
+
+    public class Watcher<T> : IDisposable
+    {
+        private readonly StreamReader _streamReader;
```
brendandburns: I don't think we use the underscore convention elsewhere, do we? I think we should be consistent.

src/Watcher.cs
```diff
@@ -0,0 +1,100 @@
...
+using k8s.Exceptions;
+using Microsoft.Rest;
+using Microsoft.Rest.Serialization;
+
+namespace k8s
+{
+    public enum WatchEventType
+    {
+        [EnumMember(Value = "ADDED")] Added,
+
```
brendandburns: I'm definitely not a csharp style expert, but it feels weird to have newlines between these values.

src/Watcher.cs
```diff
@@ -0,0 +1,100 @@
...
+                    }
+
+                    try
+                    {
+                        var line = await streamReader.ReadLineAsync();
+                        var @event = SafeJsonConvert.DeserializeObject<WatchEvent>(line);
+
+                        OnEvent?.Invoke(@event.Type, @event.Object);
+                    }
+                    catch (Exception e)
```
brendandburns: Isn't catching all exceptions an anti-pattern? Maybe just the parsing and IO exceptions?

src/WatcherDelegatingHandler.cs
```diff
@@ -0,0 +1,61 @@
+﻿using System.IO;
+using System.Net;
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace k8s
+{
+    internal class WatcherDelegatingHandler : DelegatingHandler
```
brendandburns: Perhaps document what this class does here?

src/WatcherDelegatingHandler.cs
```diff
@@ -0,0 +1,61 @@
...
+    {
+        protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage request,
+            CancellationToken cancellationToken)
+        {
+            var originResponse = await base.SendAsync(request, cancellationToken);
+
+            if (originResponse.IsSuccessStatusCode)
+            {
+                if ($"{request.RequestUri.Query}".Contains("watch=true"))
```
brendandburns: We should probably do better parsing, so that `?baywatch=true-ish` won't trigger this handling.

brendandburns:
Some small-ish comments. Many thanks for the PR!

src/Watcher.cs
tg123: I really love `go fmt`

we might introduce something like https://github.com/dotnet/codeformatter and contribution guide. I will create another issue to track this.

src/Watcher.cs
tg123: https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md

```
We use _camelCase for internal and private fields and use readonly where possibl
```

Actually, I use resharper out of box coding style formatter
https://www.jetbrains.com/resharper/features/code_formatting.html

src/WatcherDelegatingHandler.cs
tg123: Thanks, 
this must be something I am missing, cloud you please referrer any full doc about watch in k8s.

brendandburns: I'm just saying that a user can attach random query strings (possibly accidentally, possibly on purpose) and this code will trigger a watch, even if a watch isn't actually being requested.

We should parse the query string into a map, and then check `map['watch'] == 'true'`

src/WatcherDelegatingHandler.cs
tg123: fixed

src/Watcher.cs
tg123: hi please have a look at my testcases which would better illustrate what I want to do.

* TestSuriveBadLine: the watch object will still alive when json parse error or exception from onevent. the `onerror` will be triggered
* TestWatchServerDisconnect: the watch object stops as of transport level error. the `onerror` will be triggered

src/WatcherDelegatingHandler.cs
tg123: I introduce a dependency `Microsoft.AspNetCore.WebUtilities`  to parse the query. I am not sure you might consider it too heavy.

another approach is 

```
if ($"&{request.RequestUri.Query}&".Contains("&watch=true&"))
|| ($"&{request.RequestUri.Query}&".Contains("&?watch=true&"))
```

seems too ugly

tg123:
travis ci stucks if `sudo: required`. 
so I changed it to `false`

I also tested on my Ubuntu 16.04 and Windows, seems a travis ci problem.

brendandburns:
LGTM. Thanks!

PR #42, opened 2017-11-02
Update example for new config file loading.

brendandburns:
The example didn't work (at least for me) it crashed with:

```
Unhandled Exception: System.ArgumentNullException: Value cannot be null.
Parameter name: uriString
   at System.Uri..ctor(String uriString)
   at k8s.Kubernetes..ctor(KubernetesClientConfiguration config) in /home/bburns/kubernetes-client/csharp/src/Kubernetes.Auth.cs:line 25
   at simple.PodList.Main(String[] args) in /home/bburns/kubernetes-client/csharp/examples/simple/PodList.cs:line 12
```

This fixes it for the new config file approach.
@tg123

### DIFF
--- a/examples/simple/PodList.cs
+++ b/examples/simple/PodList.cs
@@ -8,7 +8,7 @@ namespace simple
     {
         static void Main(string[] args)
         {
-            var k8sClientConfig = new KubernetesClientConfiguration();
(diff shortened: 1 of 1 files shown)


tg123:
LGTM

Issue #44, opened 2017-11-03
Methods that return ObjectOrStatus don't work...


brendandburns:
This is a general problem with our current Kubernetes Swagger, Create and Delete can return an object or a `Status`, the Swagger isn't set up to handle that properly.  It tries to parse and it dies:

```
Unhandled Exception: System.AggregateException: One or more errors occurred. (Unable to deserialize the response.) ---> Microsoft.Rest.SerializationException: Unable to deserialize the respo
nse. ---> Newtonsoft.Json.JsonReaderException: Unexpected character encountered while parsing value: {. Path 'status', line 1, position 305.
   at Newtonsoft.Json.JsonTextReader.ReadStringValue(ReadType readType)
   at Newtonsoft.Json.JsonTextReader.ReadAsString()
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.ReadForType(JsonReader reader, JsonContract contract, Boolean hasConverter)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.PopulateObject(Object newObject, JsonReader reader, JsonObjectContract contract, JsonProperty member, String id)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateObject(JsonReader reader, Type objectType, JsonContract contract, JsonProperty member, JsonContainerContract containerC
ontract, JsonProperty containerMember, Object existingValue)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.Deserialize(JsonReader reader, Type objectType, Boolean checkAdditionalContent)
   at Newtonsoft.Json.JsonSerializer.DeserializeInternal(JsonReader reader, Type objectType)
   at Microsoft.Rest.Serialization.SafeJsonConvert.DeserializeObject[T](String json, JsonSerializerSettings settings)
   at k8s.Kubernetes.<DeleteNamespaceWithHttpMessagesAsync>d__173.MoveNext() in /mnt/c/Users/bburns/kubernetes-client/csharp/src/generated/Kubernetes.cs:line 29720
   --- End of inner exception stack trace ---
   at k8s.Kubernetes.<DeleteNamespaceWithHttpMessagesAsync>d__173.MoveNext() in /mnt/c/Users/bburns/kubernetes-client/csharp/src/generated/Kubernetes.cs:line 29729
   --- End of inner exception stack trace ---
   at System.Threading.Tasks.Task`1.GetResultCore(Boolean waitCompletionNotification)
[...]

PR #45, opened 2017-11-03
Add an example that creates and deletes a namespace.

brendandburns:
Simple example of create/list/delete.

Sadly it currently throws at the end because of #44 but still worth adding as an example, I think.

@tg123 @krabhishek8260

### DIFF
--- a/examples/namespace/namespace.csproj
+++ b/examples/namespace/namespace.csproj
@@ -1,0 +1,12 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <ItemGroup>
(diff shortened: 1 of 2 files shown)


examples/namespace/namespace.csproj
```diff
@@ -0,0 +1,14 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Rest.ClientRuntime" Version="3.0.3" />
```
tg123: can be removed

tg123:
LGTM

PR #46, opened 2017-11-04
Add some utilities for handling delete better.

brendandburns:
Hacky, but necessary for now.

@mbohlool @krabhishek8260 @tg123 

Partially addresses #44


tg123:
I suggest as a temporary fix below until swagger and autorest can handle it 

* comment out all fields in `V1Status` (go client did not use the response)
or
* remove content of failed api using `DelegatingHandler` like watch api

brendandburns:
Closing this in favor of #47

PR #57, opened 2017-11-16
Add IntOrString support

### LINKED ISSUE #48: API call to submit a deployment gets back HTTP 422 (Unprocessable Entity) error
I am getting an HTTP 422 (Unprocessable Entity) error for an API call I am making to submit a deployment (json attached, submitting the same json through kubectl works fine), the API call is
k8sClient.CreateNamespacedDeploymentWithHttpMessagesAsync

The only relevant line in the APIServer logs is the following:
I1104 00:08:00.847824       1 wrap.go:42] POST /apis/apps/v1beta1/namespaces/default/deployments?pretty=true: (45.627034ms) 422 [[FxVersion/4.7.2114.0 OSName/WindowsServer2012R2Datacenter OSVersion/6.3.9600 k8s.Kubernetes/0.0.0.0] 167.220.1.198:34012]

[deployment-gateway.json.txt](https://github.com/kubernetes-client/csharp/files/1450891/deployment-gateway.json.txt)

tg123:
this is to fix #48


tg123:
@brendandburns @krabhishek8260 

I add a 
```
sed -i '/BaseUri = new System.Uri(\"\");/ d' ${OUTPUT_DIR}/Kubernetes.cs
```
to `csharp.sh` in order to remove empty url which will cause an exception when init.

this is due to the template used by autorest.
And I tested code in master, it will yield the wrong `BaseUri = new System.Uri("");` too.

https://github.com/Azure/autorest.csharp/blob/master/src/vanilla/Templates/Rest/Client/ServiceClientTemplate.cshtml#L47

brendandburns:
LGTM, many thanks!

PR #60, opened 2017-11-30
generate from V1.8.4

tg123:
generate client from upstream tag 1.8.4

1. `IntOrString` changed to `IntstrIntOrString` which is support by `gen`
1. `ResourceQuantity` generated the implementation will be in separate PR (for better code review)

many object names are changed, `Corev1Namespace` -> `V1Namespace`
do this client need to be compatible with earlier generated version?

@brendandburns


brendandburns:
I think we should rev. the version from `0.1.0` to `0.2.0` since this is a breaking change...

tg123:
Regenerate using https://github.com/kubernetes-client/gen/pull/39

brendandburns:
LGTM, merging.

PR #51, opened 2017-11-08
cleanup unused code and functions and fix unittest naming style

tg123:
1. rename `Utils` to `CertUtils` , a more proper name
1. remove base64 functions from `Utils` since no long used
1. remove `DotNetUtilities.cs` since no long used
1. clean up unit test naming style to `verb` from `TestVerb` or `VerbTest`

PR #63, opened 2017-12-02
Add support for executing commands within a container

qmfrederik:
This PR adds support for the equivalent of `kubectl exec` and allows you to run commands within a container.

I've added a `exec` sample application which is a very basic attempt to replicate `kubectl exec` using the C# Kubernetes client.

From a code point of view:
- Adds a `IKubernetes.WebSocketNamespacedPodExecWithHttpMessagesAsync` method which is very similar to `IKubernetes.ConnectGetNamespacedPodExecWithHttpMessagesAsync` but returns a `ClientWebSocket`, allowing callers to interact with the process running on the container
- Adds a `ExecClient` convenience class which wraps around `ClientWebSocket` and provides access to the stdin, stdout and stderr streams.

There are important gotchas related to `ClientWebSocket`, the WebSocket client which ships as part of .NET Core:
* ClientWebSocket is not fully implemented in .NET Core 2.0. Most importantly, the `ClientWebSocket.Options` are ignored - this is where you set the authorization header or specify the client credentials. Long story short: you need the latest preview bits of .NET Core for this to work. See dotnet/corefx#5120
* There was a bug in `ClientWebSocket` which prevented the client & server on agreeing on the WebSockets subprotocol. This was fixed recently in .NET Core - see dotnet/corefx#25645
* The `ClientWebSocket` implementation on Linux is fairly complete; this is not the case on Windows. The Linux implementation is a managed implementation which works on Windows, too. See dotnet/corefx#5120 for how you can use the managed implementation on Windows.

I think it's still safe to merge this code as:
- This PR doesn't introduce any dependency on a pre-production NuGet packages
- Users using the latest corefx bits (me!) would be able to execute commands, something they can't do now.

### DIFF
--- a/src/IKubernetes.WebSocket.cs
+++ b/src/IKubernetes.WebSocket.cs
@@ -73,5 +73,58 @@ namespace k8s
         /// The cancellation token.
         /// </param>
         Task<WebSocket> WebSocketNamespacedPodPortForwardAsync(string name, string @namespace, IEnumerable<int> ports, Dictionary<string, List<string>> customHeaders = null, CancellationToken cancellationToken = default(CancellationToken));
+
+        /// <summary>
+        /// connect GET requests to attach of Pod
+        /// </summary>
+        /// <param name='name'>
+        /// name of the Pod
+        /// </param>
+        /// <param name='namespace'>
+        /// object name and auth scope, such as for teams and projects
+        /// </param>
+        /// <param name='container'>
+        /// The container in which to execute the command. Defaults to only container
+        /// if there is only one container in the pod.
+        /// </param>
+        /// <param name='stderr'>
+        /// Stderr if true indicates that stderr is to be redirected for the attach
+        /// call. Defaults to true.
+        /// </param>
+        /// <param name='stdin'>
+        /// Stdin if true, redirects the standard input stream of the pod for this
+        /// call. Defaults to false.
+        /// </param>
+        /// <param name='stdout'>
+        /// Stdout if true indicates that stdout is to be redirected for the attach
+        /// call. Defaults to true.
+        /// </param>
+        /// <param name='tty'>
+        /// TTY if true indicates that a tty will be allocated for the attach call.
+        /// This is passed through the container runtime so the tty is allocated on the
+        /// worker node by the container runtime. Defaults to false.
+        /// </param>
+        /// <param name='customHeaders'>
+        /// Headers that will be added to request.
+        /// </param>
+        /// <param name='cancellationToken'>
+        /// The cancellation token.
+        /// </param>
+        /// <exception cref="HttpOperationException">
+        /// Thrown when the operation returned an invalid status code
+        /// </exception>
+        /// <exception cref="SerializationException">
+        /// Thrown when unable to deserialize the response
+        /// </exception>
+        /// <exception cref="ValidationException">
+        /// Thrown when a required parameter is null
+        /// </exception>
+        /// <exception cref="System.ArgumentNullException">
+        /// Thrown when a required parameter is null
+        /// </exception>
+        /// <return>
+        /// A response object containing the response body and response headers.
+        /// </return>
+        Task<WebSocket> WebSocketNamespacedPodAttachAsync(string name, string @namespace, string container = default(string), bool stderr = true, bool stdin = false, bool stdout = true, bool tty = false, Dictionary<string, List<string>> customHeaders = null, CancellationToken cancellationToken = default(CancellationToken));
     }
 }
--- a/src/IKubernetes.WebSocket.cs
+++ b/src/IKubernetes.WebSocket.cs
@@ -52,6 +52,26 @@ namespace k8s
         /// <return>
         /// A <see cref="ClientWebSocket"/> which can be used to communicate with the process running in the pod.
         /// </return>
-        Task<WebSocket> WebSocketNamespacedPodExecWithHttpMessagesAsync(string name, string @namespace = "default", string command = "/bin/bash", string container = null, bool stderr = true, bool stdin = true, bool stdout = true, bool tty = true, Dictionary<string, List<string>> customHeaders = null, CancellationToken cancellationToken = default(CancellationToken));
+        Task<WebSocket> WebSocketNamespacedPodExecAsync(string name, string @namespace = "default", string command = "/bin/bash", string container = null, bool stderr = true, bool stdin = true, bool stdout = true, bool tty = true, Dictionary<string, List<string>> customHeaders = null, CancellationToken cancellationToken = default(CancellationToken));
+
+        /// <summary>
+        /// Start port forwarding one or more ports of a pod.
+        /// </summary>
+        /// <param name='name'>
+        /// The name of the Pod
+        /// </param>
+        /// <param name='namespace'>
+        /// The object name and auth scope, such as for teams and projects
+        /// </param>
+        /// <param name='ports'>
+        /// List of ports to forward.
+        /// </param>
+        /// <param name='customHeaders'>
+        /// The headers that will be added to request.
+        /// </param>
+        /// <param name='cancellationToken'>
+        /// The cancellation token.
+        /// </param>
+        Task<WebSocket> WebSocketNamespacedPodPortForwardAsync(string name, string @namespace, IEnumerable<int> ports, Dictionary<string, List<string>> customHeaders = null, CancellationToken cancellationToken = default(CancellationToken));
     }
 }
--- a/src/IKubernetes.WebSocket.cs
+++ b/src/IKubernetes.WebSocket.cs
@@ -1,0 +1,57 @@
+using System.Collections.Generic;
+using System.Net.WebSockets;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace k8s
+{
+    public partial interface IKubernetes
+    {
+        /// <summary>
+        /// Executes a command in a pod.
+        /// </summary>
+        /// <param name='name'>
+        /// name of the Pod
+        /// </param>
+        /// <param name='namespace'>
+        /// object name and auth scope, such as for teams and projects
+        /// </param>
+        /// <param name='command'>
+        /// Command is the remote command to execute. argv array. Not executed within a
+        /// shell.
+        /// </param>
+        /// <param name='container'>
+        /// Container in which to execute the command. Defaults to only container if
+        /// there is only one container in the pod.
+        /// </param>
+        /// <param name='stderr'>
+        /// Redirect the standard error stream of the pod for this call. Defaults to
+        /// <see langword="true"/>.
+        /// </param>
+        /// <param name='stdin'>
+        /// Redirect the standard input stream of the pod for this call. Defaults to
+        /// <see langword="true"/>.
+        /// </param>
+        /// <param name='stdout'>
+        /// Redirect the standard output stream of the pod for this call. Defaults to
+        /// <see langword="true"/>.
+        /// </param>
+        /// <param name='tty'>
+        /// TTY if true indicates that a tty will be allocated for the exec call.
+        /// Defaults to <see langword="true"/>.
+        /// </param>
+        /// <param name='customHeaders'>
+        /// Headers that will be added to request.
+        /// </param>
+        /// <param name='cancellationToken'>
+        /// The cancellation token.
+        /// </param>
+        /// <exception cref="ArgumentNullException">
+        /// Thrown when a required parameter is null
+        /// </exception>
+        /// <return>
+        /// A <see cref="ClientWebSocket"/> which can be used to communicate with the process running in the pod.
+        /// </return>
+        Task<WebSocket> WebSocketNamespacedPodExecWithHttpMessagesAsync(string name, string @namespace = "default", string command = "/bin/bash", string container = null, bool stderr = true, bool stdin = true, bool stdout = true, bool tty = true, Dictionary<string, List<string>> customHeaders = null, CancellationToken cancellationToken = default(CancellationToken));
+    }
+}
--- a/src/Kubernetes.WebSocket.cs
+++ b/src/Kubernetes.WebSocket.cs
@@ -65,20 +65,16 @@ namespace k8s
 
             uriBuilder.Path += $"api/v1/namespaces/{@namespace}/pods/{name}/exec";
 
-            List<string> _queryParameters = new List<string>();
-            _queryParameters.Add(string.Format("command={0}", Uri.EscapeDataString(command)));
 
-            if (container != null)
+            uriBuilder.Query = QueryHelpers.AddQueryString(string.Empty, new Dictionary<string, string>
             {
-                _queryParameters.Add(string.Format("container={0}", Uri.EscapeDataString(container)));
-            }
-
-            _queryParameters.Add(string.Format("stderr={0}", stderr ? 1 : 0));
-            _queryParameters.Add(string.Format("stdin={0}", stdin ? 1 : 0));
-            _queryParameters.Add(string.Format("stdout={0}", stdout ? 1 : 0));
-            _queryParameters.Add(string.Format("tty={0}", tty ? 1 : 0));
-
-            uriBuilder.Query = string.Join("&", _queryParameters);
+                { "command", command},
+                { "container", container},
+                { "stderr", stderr ? "1": "0"},
+                { "stdin", stdin ? "1": "0"},
+                { "stdout", stdout ? "1": "0"},
+                { "tty", tty ? "1": "0"}
+            });
 
             return this.StreamConnectAsync(uriBuilder.Uri, _invocationId, customHeaders, cancellationToken);
         }
--- a/src/Kubernetes.WebSocket.cs
+++ b/src/Kubernetes.WebSocket.cs
@@ -1,0 +1,144 @@
+using Microsoft.Rest;
+using System;
+using System.Collections.Generic;
+using System.Net.Http;
+using System.Net.WebSockets;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace k8s
+{
+    public partial class Kubernetes
+    {
+        /// <summary>
+        /// Gets a function which returns a <see cref="WebSocketBuilder"/> which <see cref="Kubernetes"/> will use to
+        /// create a new <see cref="WebSocket"/> connection to the Kubernetes cluster.
+        /// </summary>
+        public Func<WebSocketBuilder> CreateWebSocketBuilder { get; set; } = () => new WebSocketBuilder();
+
+        /// <inheritdoc/>
+        public Task<WebSocket> WebSocketNamespacedPodExecWithHttpMessagesAsync(string name, string @namespace = "default", string command = "/bin/sh", string container = null, bool stderr = true, bool stdin = true, bool stdout = true, bool tty = true, Dictionary<string, List<string>> customHeaders = null, CancellationToken cancellationToken = default(CancellationToken))
+        {
+            if (name == null)
+            {
+                throw new ArgumentNullException(nameof(name));
+            }
+
+            if (@namespace == null)
+            {
+                throw new ArgumentNullException(nameof(@namespace));
+            }
+
+            if (command == null)
+            {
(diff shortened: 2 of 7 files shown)


examples/exec/NuGet.config
```diff
@@ -0,0 +1,5 @@
+<configuration>
```
brendandburns: Do we want to add this file?

examples/exec/exec.csproj
```diff
@@ -0,0 +1,19 @@
...
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <LangVersion>latest</LangVersion>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\src\KubernetesClient.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
```
brendandburns: What does this do? Do we want this in here?

brendandburns: What if that file doesn't exist?

src/ExecClient.cs
```diff
@@ -0,0 +1,116 @@
...
+
+        public async Task Write(string command, CancellationToken cancellationToken = default(CancellationToken))
+        {
+            // Send:
+            //  - [uint32: stream index = 0 for stdin]
+            //  - [byte[]: command as UTF8 value]
+            var commandLength = Encoding.UTF8.GetByteCount(command);
+            var buffer = ArrayPool<byte>.Shared.Rent(commandLength + 4);
+
+            try
```
brendandburns: nit: may as well move this try down to around the send, right?

src/ExecClient.cs
```diff
@@ -0,0 +1,116 @@
...
+
+        protected async Task RunLoop(CancellationToken cancellationToken)
+        {
+            // Get a 1KB buffer
+            byte[] buffer = ArrayPool<byte>.Shared.Rent(1024 * 1024);
+            var segment = new ArraySegment<byte>(buffer);
+
+            while (!cancellationToken.IsCancellationRequested && this.socket.CloseStatus == null)
+            {
+                // We always get data in this format:
```
brendandburns: Ultimately we're going to want to pull this logic out so that we can use it for Attach and PortForward, but we can do that in a different PR...

src/ExecClient.cs
```diff
@@ -0,0 +1,116 @@
...
+                }
+
+                var streamType = buffer[0];
+
+
+                var value = Encoding.UTF8.GetString(buffer, 1, result.Count - 1);
+
+                switch (streamType)
+                {
+                    case 1:
```
brendandburns: Can you make `1` and `2` constants?

brendandburns:
This looks great! Thanks for doing it (and the thorough investigation...)

Some minor nits, and you need to sign the CLA, then I think this is good to merge!

Thanks again!

examples/exec/NuGet.config
qmfrederik: If you want the 'F5' experience to work with this example, then yes.

`ClientWebSocket` only works properly with corefx 2.1 which has dotnet/corefx#25645; so very recent nightly builds. If you don't target `netcoreapp2.1` the code will compile but you'll hit a bug in .NET Core so the sample won't work.

Once there's a public preview of .NET Core 2.1 this can be removed.

examples/exec/exec.csproj
qmfrederik: We don't need it, so I removed the reference.

src/ExecClient.cs
qmfrederik: The try/finally protects `buffer` which comes from the shared array pool. The code that initializes the first 4 bytes of the buffer won't fail, but `Encoding.UTF8.GetBytes` may throw an exception. So if we move the `try` down to `SendAsync`, we may leak a buffer if `GetBytes` fails.
(It's very unlikely it will since we already call `Encoding.UTF8.GetByteCount`)

I can move it down; but because the `try` is there make sure we return `buffer` once we're done with it, I prefer to keep it close to `buffer`. 

I don't have a strong preference, though, so can always move it down if you prefer it that way.

src/ExecClient.cs
qmfrederik: Done, thx.

src/ExecClient.cs
qmfrederik: Yeah, I didn't have the time to look at the protocol for Attach yet (not sure whether it uses stream multiplexing or not).

I had a cursory look at `PortForward` and I believe it's simpler - there is no multiplexing so you can just use `ReceiveAsync`/`SendAsync` to talk with the remote socket.
You may want to add code which wraps the `ClientWebSocket` in `Stream`-like class, though.

That said, one thing at a time 😄

qmfrederik:
@brendanburns You're welcome, I updated the code/responed to your comments, the CLA status should also be OK now.

Let me know if you need anything else!

src/ExecClient.cs
```diff
@@ -9,6 +9,10 @@ namespace k8s
 {
     public class ExecClient : IDisposable
     {
+        private const int StdInStreamIndex = 0;
```
tg123: const from py client
https://github.com/kubernetes-client/python-base/blob/master/stream/ws_client.py#L24

tg123:
Cloud you please also attach some testcases and also have a look at my suggestion about channel length

qmfrederik:
@tg123 I'll try to add some unit tests.

I'm sorry - I may have missed your comment about channel length. Can you let me know where you left that comment (or repeat it here)?

tg123:
@qmfrederik really sorry I did not click submit review posted yesterday but reviews were shown on my page. you can now see it.

examples/exec/Program.cs
```diff
@@ -0,0 +1,41 @@
...
+namespace exec
+{
+    class Program
+    {
+        static async Task Main(string[] args)
+        {
+            var pod = args[0];
+            var container = args.Length > 1 ? args[1] : null;
+
+            var k8sClientConfig = KubernetesClientConfiguration.BuildConfigFromConfigFile(kubeconfigPath: "kube-config.yml");
```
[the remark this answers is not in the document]
qmfrederik: Thx, I fixed this.

src/ExecClient.cs
qmfrederik: Thx, I added the other channel indices as well.

examples/exec/NuGet.config
qmfrederik: Well, turns out Travis doesn't have .NET Core 2.1 (which is reasonable 😄 ), this caused the build to break so I reverted this change for now.

qmfrederik:
@tg123 Thanks for the feedback! I updated the PR.

As for unit tests, I added unit tests for `Kubernetes.WebSocketNamespacedPodExecWithHttpMessagesAsync`.
In my unit test I wanted to make sure the constructed URL, the message headers and client certificates are OK.

The easiest way I found to do this was by introducing a `WebSocketBuilder` class.
You pass your WebSocket configuration to this class and this class gives you a `ClientWebSocket` in return.
I've modified the `Kubernetes` class so that you can pass your own `WebSocketBuilder`. I pass a mock one in the unit tests to intercept this.

Let me see what I can do about the ExecClient (or whatever its future shape may be).

qmfrederik:
@tg123 @brendandburns Thanks, I updated the PR.

Based on your feedback, would it be better to rename `ExecClient` to `ConsoleClient` or `TerminalClient`? It's a generic class which adds support for multiplexing stdout/stderr so I guess it can be reused accross Attach/Exec.

Alternatively, we can split `ExecClient` off into a separate PR if that needs more design work.

.gitignore
```diff
@@ -2,6 +2,7 @@
 .vs
 obj/
 bin/
+kube-config.yml
```
tg123: can be removed

src/ExecClient.cs
```diff
@@ -0,0 +1,124 @@
...
+            // Send:
+            //  - [byte: stream index = 0 for stdin]
+            //  - [byte[]: command as UTF8 value]
+            var commandLength = Encoding.UTF8.GetByteCount(command);
+            var buffer = ArrayPool<byte>.Shared.Rent(commandLength + 1);
+
+            try
+            {
+                // The first byte represent the stream index. For stdin this is 0
+                buffer[0] = 0;
```
tg123: cloud be const

examples/exec/Program.cs
```diff
@@ -0,0 +1,40 @@
...
+    {
+        static async Task Main(string[] args)
+        {
+            var pod = args[0];
+            var container = args.Length > 1 ? args[1] : null;
+
+            var k8SClientConfig = KubernetesClientConfiguration.BuildConfigFromConfigFile();
+            var client = new Kubernetes(k8SClientConfig);
+
+            var webSocket = await client.WebSocketNamespacedPodExecWithHttpMessagesAsync(pod, container: container, command: "/bin/bash").ConfigureAwait(false);
```
tg123: Maybe the API cloud return `ExecClient` directly, make `websocket with channel` for intern use only

src/ExecClient.cs
tg123: I think it would be better to hide channel in the `websocket` class which can be reuse in attach. but I think it is OK to impl them in next PR.

tests/ExecClientTests.cs
```diff
@@ -0,0 +1,41 @@
...
+using System.Text;
+using System.Threading;
+using Xunit;
+
+namespace k8s.tests
+{
+    public class ExecClientTests
+    {
+        [Fact]
+        public void ReadWriteTest()
```
tg123: cloud you please align with naming style only `verb` 
like https://github.com/kubernetes-client/csharp/pull/51

src/ExecClient.cs
```diff
@@ -0,0 +1,124 @@
...
+
+                // Copy the command to the buffer, starting at offset 1
+                Encoding.UTF8.GetBytes(command, 0, command.Length, buffer, 1);
+
+                ArraySegment<byte> segment = new ArraySegment<byte>(buffer, 0, commandLength + 1);
+                await this.socket.SendAsync(segment, WebSocketMessageType.Binary, true, cancellationToken).ConfigureAwait(false);
```
tg123: send/recy websocket in C# seems did not thread safe.

src/ExecClient.cs
```diff
@@ -0,0 +1,124 @@
...
+        private Task runLoop;
+
+        public ExecClient(WebSocket socket)
+        {
+            this.socket = socket ?? throw new ArgumentNullException(nameof(socket));
+            this.cts = new CancellationTokenSource();
+        }
+
+        public event EventHandler ConnectionClosed;
+        public event EventHandler<string> StandardOutputReceived;
```
tg123: I prefer java client style

ExecClient with 3 streams Stdin, Stdout, Stderr

event string does not seem to be friendly if you want to readline or something else

qmfrederik:
@tg123 Thanks!

I split off the exec-related functionality for now; we can do this in a separate PR.

I'd really prefer the Kubernetes C# client to expose a "normal" `WebSocket` class. It gives any user of the C# Kubernetes client the freedom to communicate over the WebSocket class the way they want.

It would be fine for the C# client to add default 'wrappers' around WebSocket which allow you to interact with the data in a specific way (e.g. by demuxing the streams); but as a user I'd really like to have access to the raw WebSocket object, too.
I've been bitten a couple of times in the past by APIs which tried to do "the right thing", but whatever they were doing wasn't the right thing for me, and I was left with no option but to compile from source. That'd be unfortunate.

Let me know if the WebSocket-related changes look good to me.

src/IKubernetes.Exec.cs
```diff
@@ -0,0 +1,57 @@
...
+        /// <param name='cancellationToken'>
+        /// The cancellation token.
+        /// </param>
+        /// <exception cref="ArgumentNullException">
+        /// Thrown when a required parameter is null
+        /// </exception>
+        /// <return>
+        /// A <see cref="ClientWebSocket"/> which can be used to communicate with the process running in the pod.
+        /// </return>
+        Task<WebSocket> WebSocketNamespacedPodExecWithHttpMessagesAsync(string name, string @namespace = "default", string command = "/bin/bash", string container = null, bool stderr = true, bool stdin = true, bool stdout = true, bool tty = true, Dictionary<string, List<string>> customHeaders = null, CancellationToken cancellationToken = default(CancellationToken));
```
tg123: FileName to Ixxx.Websocket.cs or Ixxx.Stream.cs

Cloud we name this StreamConnectAsync(path, ...) or something else, which just return plain websocket
since this PR has nothing to do with `exec`

qmfrederik:
@tg123 I tried to update the PR in line with your feedback:
- I renamed the files to `.WebSocket.cs` to make it clear the files contain the WebSocket-related functionality
- Code which I think can be shared across attach, port forward,... is split off to a `StreamConnectAsync`
- I kept `WebSocketNamespacedPodExecWithHttpMessagesAsync` as this could will return a `WebSocket` which connects to the `exec` endpoint for a specific pod.

src/Kubernetes.WebSocket.cs
```diff
@@ -0,0 +1,144 @@
...
+            {
+                _queryParameters.Add(string.Format("container={0}", Uri.EscapeDataString(container)));
+            }
+
+            _queryParameters.Add(string.Format("stderr={0}", stderr ? 1 : 0));
+            _queryParameters.Add(string.Format("stdin={0}", stdin ? 1 : 0));
+            _queryParameters.Add(string.Format("stdout={0}", stdout ? 1 : 0));
+            _queryParameters.Add(string.Format("tty={0}", tty ? 1 : 0));
+
+            uriBuilder.Query = string.Join("&", _queryParameters);
```
tg123: take look at `QueryHelpers.AddQueryString` I already imported this.
it might help work encode strings

tg123:
LGTM

looking forward friendly api in #64

tg123:
by the way 
cloud you please squash you commits?

qmfrederik:
@tg123 Sure, rebased & squashed.

qmfrederik:
@tg123, @brendandburns Are we good to merge this one? Anything else you need?

tg123:
please wait for a bit 

because

http://events.linuxfoundation.org/events/kubecon-and-cloudnativecon-north-america/program/featured-speakers#brendan-burns

qmfrederik:
Ah, makes sense. Thanks for letting me know!

src/IKubernetes.WebSocket.cs
```diff
@@ -0,0 +1,77 @@
...
+        /// <param name='ports'>
+        /// List of ports to forward.
+        /// </param>
+        /// <param name='customHeaders'>
+        /// The headers that will be added to request.
+        /// </param>
+        /// <param name='cancellationToken'>
+        /// The cancellation token.
+        /// </param>
+        Task<WebSocket> WebSocketNamespacedPodPortForwardAsync(string name, string @namespace, IEnumerable<int> ports, Dictionary<string, List<string>> customHeaders = null, CancellationToken cancellationToken = default(CancellationToken));
```
brendandburns: Why is this in the `WebSocket` class? I feel like this should be in the port forward class...

src/Kubernetes.WebSocket.cs
```diff
@@ -0,0 +1,195 @@
...
+            // Send Request
+            cancellationToken.ThrowIfCancellationRequested();
+
+            WebSocket webSocket = null;
+
+            try
+            {
+                webSocket = await webSocketBuilder.BuildAndConnectAsync(uri, CancellationToken.None).ConfigureAwait(false);
+            }
+            catch (Exception ex)
```
brendandburns: Please don't catch bare `Exception` only catch expected Exceptions.

src/WebSocketBuilder.cs
```diff
@@ -0,0 +1,43 @@
...
+namespace k8s
+{
+    /// <summary>
+    /// The <see cref="WebSocketBuilder"/> creates a new <see cref="WebSocket"/> object which connects to a remote WebSocket.
+    /// </summary>
+    /// <remarks>
+    /// By default, this uses the .NET <see cref="ClientWebSocket"/> class, but you can inherit from this class and change it to
+    /// use any class which inherits from <see cref="WebSocket"/>, should you want to use a third party framework or mock the requests.
+    /// </remarks>
+    public class WebSocketBuilder
```
brendandburns: This class doesn't seem to add much imho.  Why not just inline the various calls in the class in the code above...

qmfrederik: The calls are not inlined because the unit tests use a mock `WebSocketBuilder` which traces all the parameters which have been passed and returns a dummy `WebSocket`.

I used a Builder-pattern becuase the code also uses the `UriBuilder` class.

I can change it a function/property like `GetWebSocket(Dictionary<string,string> headers, Collection<X509Certificate2> clientCertificates, Uri uri)`, though, if that's what you'd prefer. As long as the unit tests can intercept the call it's fine with ime.

src/Kubernetes.WebSocket.cs
qmfrederik: The catch catch statement is there so we can trace errors; the exception itself is rethrown.

.NET doesn't have the concept of checked exceptions and the MSDN documentation doesn't list which exceptions `ClientWebSocket.ConnectAsync` can throw.

src/IKubernetes.WebSocket.cs
qmfrederik: It's in the `IKubernetes` interface, this file just contains manual extensions to `IKubernetes` for API calls which returns WebSockets.
I can move it to something like `IKubernetes.PortForward`, if that's what you'd prefer - so one file per method?

src/Kubernetes.WebSocket.cs
qmfrederik: Like this:

```csharp
string queryString = string.Empty;
queryString = QueryHelpers.AddQueryString(queryString, "command", command);

if (container != null)
{
    queryString = QueryHelpers.AddQueryString(queryString, "container", container);
}

queryString = QueryHelpers.AddQueryString(queryString, "stderr", stderr ? "1" : "0");
queryString = QueryHelpers.AddQueryString(queryString, "stdin", stdin ? "1" : "0");
queryString = QueryHelpers.AddQueryString(queryString, "stdout", stdout ? "1" : "0");
queryString = QueryHelpers.AddQueryString(queryString, "tty", tty ? "1" : "0");

uriBuilder.Query = queryString;
```
?

src/KubernetesClient.csproj
```diff
@@ -23,5 +23,6 @@
     <PackageReference Include="Newtonsoft.Json" Version="10.0.2" />
     <PackageReference Include="System.ValueTuple" Version="4.4.0" />
     <PackageReference Include="YamlDotNet.NetCore" Version="1.0.0" />
+    <PackageReference Include="System.Net.WebSockets.Client" Version="4.3.2"/>
```
tg123: seems netstandard can use websocket without this dependeny

src/Kubernetes.WebSocket.cs
tg123:
```
            var queryString = QueryHelpers.AddQueryString(string.Empty, new Dictionary<string, string>
            {
                { "command", "aaaa"},
                { "tty", "1"}
            });
```

src/IKubernetes.WebSocket.cs
tg123: why not also add attach

src/KubernetesClient.csproj
qmfrederik: I think you can on netstandard2.0 but it doesn't work on `netstandard1.4`. Did you run `dotnet restore` when you tested this?

src/IKubernetes.WebSocket.cs
qmfrederik: Done ;-)

src/Kubernetes.WebSocket.cs
qmfrederik: Done, thanks!

src/WebSocketBuilder.cs
brendandburns: Injectable for testing is fair...

Issue #71, opened 2017-12-08
JSON deserialization of non-JSON content in proxy-related methods


philliphoff:
The generated methods for the proxy-related REST calls (e.g. `ProxyGETNamespacedServiceWithHttpMessagesAsync()`) expect responses to be JSON-serialized strings; this is often not the case (e.g. when a web site is being proxied and raw HTML, CSS, and JavaScript are returned).

Those proxy-related methods should probably just pass through the raw response content.

brendandburns:
Yeah, this is tricky to do correctly. It will be improved when Port forwarding is supported.

PR #75, opened 2017-12-21
generate with stream support

tg123:
let log return Stream instead of string

some problem also happens in #71, I will update all related methods to stream in another PR due to I did not the pattern at the time,.


brendandburns:
LGTM

PR #76, opened 2017-12-22
 Strongly-name the KubernetesClient assembly

### LINKED ISSUE #70: Signed assemblies?
As the library assemblies are unsigned, they cannot be referenced by assemblies that are (required to be) signed, which then excludes its use in certain applications.

It'd be great to offer a signed flavor of the assemblies.

(Alternatively, one could fork the repo, build, and sign a forked version, but that's a good deal of overhead to take on.)

qmfrederik:
[Strongly-naming is a deprecated .NET technology which is actively being discouraged.](https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/strong-name-signing.md#1-microsoft-strong-names-their-assemblies-should-i)

However, some projects strongly-name their assembly and this means they can only use strongly-named dependencies. From a compatibility perspective, it sometimes make sense to strongly-name open source projects.

This PR adds a strong name to the KubernetesClient assembly.

It also updates the dependencies:
* From BouncyCastle.NetCore to Portable.BouncyCastle, the later being the more popular variant
* YamlDotNet.NetCore to YamlDotNet, the later being the upstream project and now publishing packages for netstandard1.3

This PR checks in the private key used for signing in the Git repository. That's in [accordance with the guidelines](https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/strong-name-signing.md#2-i-really-really-need-to-strong-name-what-kinds-of-issues-will-i-run-into):

> Firstly, we would recommend that these open source projects check-in their private key (remember, strong-names are used for identity, and not for security).

YamlDotNet and Fractions do not strongly name their assemblies.

Fixes #70


brendandburns:
Travis build is broken, but otherwise LGTM.

qmfrederik:
It looks like this is blocked by .NET Core not supporting strongly signing assemblies on non-Windows platforms. dotnet/roslyn#8210 tracks the feature request and it seems this was merged for the next release (VS 15 Update 16, which is currently under development); the PR is dotnet/roslyn#23061.

This is blocked on Linux until VS 15.6 is updated and the Travis images use a newer version of the .NET SDK.

Alternatively, we can use AppVeyor to build on Windows.

brendandburns:
@qmfrederik given that we're requiring 2.1 these days, we can probably revisit this if you want to...

qmfrederik:
@brendanburns This seems to be OK now. The Linux leg of the build is failing because of the ` k8s.Tests.WatchTests.WatchServerDisconnect` test, which seems to be flaky.

PR #77, opened 2018-01-06
Use mainstream and strongly-named versions of BouncyCastle, YamlDotNet

qmfrederik:
- Use Portable.BouncyCastle instead of BouncyCastle.NetCore. The portable version has > 1.4M downloads on NuGet and tracks upstream BouncyCastle
- Use YamlDotNet.Signed instead of YamlDotNet.NetCore, YamlDotNet.Signed is the package maintained by the YamlDotNet author.

Split off from #76


brendandburns:
LGTM, thanks.

Issue #80, opened 2018-01-14
NuGet package does not support .NET Standard 4.5.2


shaneqld:
When attempting to add the package in a .NET Framework 4.5.2 project, nuget reports the following error:

Could not install package 'KubernetesClient 0.3.0-beta'. You are trying to install this package into a project that targets '.NETFramework,Version=v4.5.2', but the package does not contain any assembly references or content files that are compatible with that framework. For more information, contact the package author.

I believe this is because the package specifically targets only .NET Standard 1.4.

tg123:
#32 

4.5 should work well but not yet tested

shaneqld:
.NET 4.6 does not work from the public NuGet package.  Error below:

Could not install package 'KubernetesClient 0.3.0-beta'. You are trying to install this package into a project that targets '.NETFramework,Version=v4.6', but the package does not contain any assembly references or content files that are compatible with that framework. For more information, contact the package author.

tg123:
https://docs.microsoft.com/en-us/dotnet/standard/net-standard

sorry please try net461, 3.0-beta complies with netstandard 1.4

shaneqld:
Thanks, was just reading about that.  Perhaps add this to the README.

60 packages needed to get this installed.  That's a lot! :)

Any plans to make a vanilla .NET Framework version that does away with .NET Standard?

tg123:
I will try to prune dependencies and make multi-target working

brendandburns:
Closing this since #86 was merged.

PR #86, opened 2018-01-22
support target net452

tg123:
this is for #80, #32 

however, target net452 cannot be built on travis ci.
a replacement is wanted. 

I suggest migrate to VSTS which can provide Windows agent and can build multiple targets net452 and netstandard1.4

### DIFF
--- a/src/Kubernetes.ConfigInit.cs
+++ b/src/Kubernetes.ConfigInit.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Diagnostics.CodeAnalysis;
+using System.Net;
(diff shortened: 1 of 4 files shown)


qmfrederik:
@tg123 If you want to build for .NET 4.5.2 as well, you can also consider [AppVeyor](https://www.appveyor.com/), it's free for open source projects just like Travis but runs on Windows instead of Linux.

tg123:
Thanks for the information @qmfrederik
 It might be a good candidate.

brendandburns:
LGTM.

Issue #87, opened 2018-01-23
SDK is triggering HttpOperationException when creating any entity.


allantargino:
My team and I was trying to create a namespace using the following code:

```cs
[...]
var KubeNamespaceBody = new V1Namespace
{
	Metadata = new V1ObjectMeta
	{
		Name = nameSpace
	}
};
var KubeNamespace = await _kubernetes.CreateNamespaceAsync(KubeNamespaceBody);
[...]
```

We're using Kubernetes 1.7.7, hosted on Azure (ACS).
The Kubernetes API is able to create the namespace with success, returning a HTTP status of 202 (CREATED).

Unfortunately,  on line 2235 of [Kubernetes.cs](https://github.com/kubernetes-client/csharp/blob/master/src/generated/Kubernetes.cs), it contains the following piece, expecting a HTTP 200:
```cs
[...]
if ((int)_statusCode != 200 && (int)_statusCode != 401)
{
	var ex = new HttpOperationException(string.Format("Operation returned an invalid status code '{0}'", _statusCode));
```
and on line 2262:
```cs
// Deserialize Response
if ((int)_statusCode == 200)
{
```

The new reference is shown here: https://kubernetes.io/docs/api-reference/v1.9/#namespace-v1-core
![image](https://user-images.githubusercontent.com/13934447/35287919-9c9e97e4-004a-11e8-97b1-7eb22fd456b2.png)

The old one here: https://v1-7.docs.kubernetes.io/docs/api-reference/v1.7/#namespace-v1-core
![image](https://user-images.githubusercontent.com/13934447/35287937-a7682bc2-004a-11e8-8b9a-8b3c35c6dffc.png)

As a temporary alternative, we've changed the mentioned code for the following, and it's working perfectly.
```cs
[...]
if ((int)_statusCode != 200 && (int)_statusCode != 201 && (int)_statusCode != 202 && (int)_statusCode != 401)
{
```

I believe this behavior is due to AutoRest was used with an older version of Kubernetes API. 
Should we sent a PR? Or you have plans to generated that code with a more recently version?

allantargino:
@viniciussouza

tg123:
seems bug in swagger.json

I regenerate the sdk with 1.9.2

![image](https://user-images.githubusercontent.com/170430/35292014-285917be-00aa-11e8-94f9-b0917fa24f0c.png)

tg123:
[...]

PR #84, opened 2018-01-18
make V1.Patch strong type of JsonPatch

tg123:
fix #78

PR #88, opened 2018-01-23
generate using 1.9.2

tg123:
regenerate with 1.9.2 for #87 

might be conflict with #84  and #85 
those patches should be regenerate using 1.9.2


brendandburns:
LGTM but now needs regen/rebase.

PR #92, opened 2018-02-07
connect to GCP GKE from local machine

### LINKED ISSUE #91: BuildConfigFromConfigFile explanation missing
Hello, I'm trying to run code copied from examples but got `Unhandled Exception: k8s.Exceptions.KubeConfigException: User: X does not have appropriate auth credentials in kubeconfig`

Can not find instructions in readme how should we run code locally agains google cloud kubernetes

I do have `~/.kube` and commands like `kubectl get pods` are working as expected

mac2000:
explanation of how to connect from lacal machine to Google Cloud Kubernetes without headache, close #91

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,16 +46,23 @@ cd csharp\examples\simple
 dotnet run
 ```
 
-## Connecting to GCP GKE from local machine
+## Connecting to a cluster from a local machine
 
-Start proxy to access kubernetes cluster:
+Preferred way of connecting to a cluster:
+
+```
+var config = KubernetesClientConfiguration.BuildConfigFromConfigFile();
+var client = new Kubernetes(config);
+```
+
+Not all auth providers are supported at moment [#91](https://github.com/kubernetes-client/csharp/issues/91#issuecomment-362920478), but you still can connect to cluster by starting proxy:
 
 ```bash
 $ kubectl proxy
 Starting to serve on 127.0.0.1:8001
 ```
 
-In code use following config:
+and changing config:
 
 ```csharp
 var config = new KubernetesClientConfiguration {  Host = "http://127.0.0.1:8001" };
--- a/README.md
+++ b/README.md
@@ -46,9 +46,9 @@ cd csharp\examples\simple
 dotnet run
 ```
 
-## Connecting to a cluster from a local machine
+## Known issues
 
-Preferred way of connecting to a cluster:
+While preferred way of connecting to a remote cluster from local machine is:
 
 ```
 var config = KubernetesClientConfiguration.BuildConfigFromConfigFile();
@@ -67,6 +67,8 @@ and changing config:
 ```csharp
 var config = new KubernetesClientConfiguration {  Host = "http://127.0.0.1:8001" };
 ```
+
+Notice that this is a workaround and is not recommended for production use
 
 ## Testing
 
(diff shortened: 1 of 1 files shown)


mac2000:
Hm, I have signed PR, should I do something next with it? or it will mark as ok in a while?

tg123:
you can try push -f to force it rerun on your side. project admin may retrigger it too.

README.md
```diff
@@ -46,6 +46,21 @@ cd csharp\examples\simple
 dotnet run
 ```
 
+## Connecting to GCP GKE from local machine
```
brendandburns: These instructions are generic to any cloud provider, so it likely should be:

`Connecting to a cluster from a local machine`

or somesuch.

README.md
```diff
@@ -46,6 +46,21 @@ cd csharp\examples\simple
...
+
+Start proxy to access kubernetes cluster:
+
+```bash
+$ kubectl proxy
```
brendandburns: I don't think connecting through the proxy is the preferred way to sugest that people connect.

README.md
```diff
@@ -46,6 +46,21 @@ cd csharp\examples\simple
...
+Starting to serve on 127.0.0.1:8001
+```
+
+In code use following config:
+
+```csharp
+var config = new KubernetesClientConfiguration {  Host = "http://127.0.0.1:8001" };
```
brendandburns: The preferred way to connect is to use the config file on disk as follows:
```cs
            var k8SClientConfig = KubernetesClientConfiguration.BuildConfigFromConfigFile();
            IKubernetes client = new Kubernetes(k8SClientConfig);
```

https://github.com/kubernetes-client/csharp/blob/master/examples/namespace/Namespace.cs#L52

brendandburns:
Thanks for the PR. I made some comments. Can you adapt the example to use the preferred way of connecting to a cluster:

```cs
var k8SClientConfig = KubernetesClientConfiguration.BuildConfigFromConfigFile();
IKubernetes client = new Kubernetes(k8SClientConfig);
```

Then I think we can merge it in.

Thanks!

mac2000:
I do agree with your point about `Connecting to a cluster from a local machine` reason why I called it like so is just because GKE is only one cloud provider who has stable K8S not in preview, but it gonna change fast so you are right

As about preferred way - yep it definitely should be like you have written, but unfortunately it will not always work - this PR is all about work around so people can still play with K8S api without headache

So I have changed description to "here is how you should do it normally and here is workaround"

Thank you in advance

brendandburns:
Can you explain the situations where it won't work? If 'kubectl' works then you have a valid kubeconfig. The c# library should pick that up and connect.

If there is a kubeconfig that doesn't work, please file a bug with an example config, that's something we should fix.

I don't want people to rely on proxy to use this client library

Thanks!
--brendan

mac2000:
Yep, it was first thing I did, here it is: https://github.com/kubernetes-client/csharp/issues/91#issuecomment-362920478

I totally agree with your point, especially for something running in production, but it is a alternative for a temporary playground, e.g. in my case simple hello world application

brendandburns:
Ah thanks for the clarification.

Can you retitle this section "known issues" and describe this as a work around?

Thanks

--brendan

brendandburns:
LGTM, many thanks for the patience.

PR #95, opened 2018-02-14
Update README

### LINKED ISSUE #81: README instructions mistake
There is a small mistake in README.md.  Specifically where it says:

```bash
$ Where REPO_DIR points to the root of the csharp repository
cd ${REPO_DIR}/csharp/src
${GEN_DIR}/openapi/csharp.sh generated csharp.settings
```

`csharp.settings` is located in $REPO_DIR/csharp, but in the example we are in $REPO_DIR/csharp/src.  The last line of this block should instead say:

```bash
${GEN_DIR}/openapi/csharp.sh generated ../csharp.settings
```

Also:

- it might be helpful to list dependencies as: git, nodejs >= 7.10.0, docker
- it might be helpful to state that the instructions are intended to be run on a Linux system (since it's a C# library, people might assume they are Windows instructions)

qmfrederik:
- Document the requirement for a Linux machine with nodejs
- Fix a relative path in the instructions

Fixes #81

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,11 @@ dotnet add package KubernetesClient
 
 ## Prerequisites
 
+The client code is generated using [AutoRest](https://github.com/azure/autorest).
+You'll need a Linux machine and [node.js](https://nodejs.org/en/) 7.10.0 or greater.
+
+The generated code works on all platforms supported by .NET or .NET Core.
+
 Check out the generator project into some other directory
 (henceforth `$GEN_DIR`)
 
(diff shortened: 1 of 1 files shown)


tg123:
generator runs in docker, I think the only dependency is docker. no nodejs needed.

qmfrederik:
Sounds good! I removed the `npm install autorest` step as well, as it seems that's not a requirement anymore.

README.md
```diff
@@ -14,6 +14,10 @@ dotnet add package KubernetesClient
 
 ## Prerequisites
 
+You'll need a Linux machine with Docker.
```
brendandburns: it _might_ work on OS X...

brendandburns:
LGTM, thanks.

PR #98, opened 2018-03-08
Add a utility for YAML loading.

### LINKED ISSUE #59: Simplify loading models from YAML files
I've realized that trying to generate all my Kubernetes objects in C# code may be a bit too cumbersome and loading them from YAML instead may be a better way.

So I ended up with something like this:

```csharp
var deserializer = 
    new DeserializerBuilder()
    .WithNamingConvention(new CamelCaseNamingConvention())
    .Build();

var deployment = deserializer.Deserialize<Apiappsv1beta1Deployment>(File.ReadAllText("deployment.yml"));
```

While this works, it would be neat if there was an easier way to do this; say:

```csharp
var deployment = Apiappsv1beta1Deployment.FromYamlFile("deployment.yml");
```

brendandburns:
Fixes #59 

@qmfrederik

### DIFF
--- a/src/Yaml.cs
+++ b/src/Yaml.cs
@@ -1,0 +1,34 @@
+using System.IO;
+using System.Threading.Tasks;
+using YamlDotNet.Serialization;
+using YamlDotNet.Serialization.NamingConventions;
+
+namespace k8s
+{
+    /// <summary>
+    /// This is a utility class that helps you load objects from YAML files.
+    /// </summary>
+
+    public class Yaml {
+        public static async Task<T> LoadFromStreamAsync<T>(Stream stream) {
+            var reader = new StreamReader(stream);
+            var content = await reader.ReadToEndAsync();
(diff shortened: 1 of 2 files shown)


src/Yaml.cs
```diff
@@ -0,0 +1,34 @@
...
+
+    public class Yaml {
+        public static async Task<T> LoadFromStreamAsync<T>(Stream stream) {
+            var reader = new StreamReader(stream);
+            var content = await reader.ReadToEndAsync();
+            return LoadFromString<T>(content);
+        }
+
+        public static async Task<T> LoadFromFileAsync<T> (string file) {
+            using (FileStream fs = File.Open(file, FileMode.Open, FileAccess.Write, FileShare.None)) {
```
tg123: there is a shortcut `File.OpenRead`

qmfrederik:
Awesome, thanks!

Issue #102, opened 2018-03-11
WebSockets - custom validation of server certificate


tintoy:
Hey, @brendanburns, as previously discussed in [tintoy/dotnet-kube-client#6](https://github.com/tintoy/dotnet-kube-client/issues/6#issuecomment-371984010) I'm happy to contribute my WebSocket functionality (which includes support for custom verification of the server certificate).

But I figured it'd be worth discussing here first :)

The big gotcha (currently) is that it has a dependency on `netcoreapp2.1` because in previous versions of corefx WebSockets used WinHTTP on Windows (which didn't honour a subset of the certificate options on `ClientWebSocketOptions`) and something else on OSX / Linux (which didn't honour a _different_ subset of the certificate options). In 2.1, the WebSockets implementation is entirely implemented in managed code and fully supports customisation of SSL certificate usage / validation.

As it stands, here's how I implemented it:

https://github.com/tintoy/dotnet-kube-client/blob/develop/src/KubeClient.Extensions.WebSockets/K8sWebSocketOptions.cs#L13
https://github.com/tintoy/dotnet-kube-client/blob/develop/src/KubeClient.Extensions.WebSockets/K8sWebSocket.cs#L31

I borrowed some of the original `ClientWebSocket` code from .NET Core 2.0 and modified it to work with `2.1-preview1` (but from `2.1-preview2` onwards, the `ManagedWebSocket` types will be publicly available and much of this code can be eliminated).

The latest code from corefx is a fair bit simpler because the required managed types are available internally:

https://github.com/dotnet/corefx/blob/f1f3b3f8af21c0f782f7f0699d505a0cdb9026ba/src/System.Net.WebSockets.Client/src/System/Net/WebSockets/WebSocketHandle.Managed.cs#L71

tintoy:
So it might be worth placing in a separate package (as I did in my client) that [extends the client](https://github.com/tintoy/dotnet-kube-client/blob/develop/src/KubeClient.Extensions.WebSockets/ResourceClientWebSocketExtensions.cs#L56) to add operations that are WebSocket-dependent.

tintoy:
CC: @qmfrederik

qmfrederik:
It's a pitty the WebSockets implementation was not complete in .NET Core 2.0 but the good news is it is being fixed in 2.1.

My 2 cents would be that it's:
[...]

PR #103, opened 2018-03-13
Custom validation of server certificate for WebSockets

tintoy:
Hi.

This PR adds improved support for customisation of SSL-related options for WebSocket connections. It depends on `netcoreapp2.1` so these features are only available when consumers are targeting that. If they're targeting `netstandard`, the full .NET framework, or an earlier version of `netcoreapp` then the existing behaviour and API surface still apply.

**Note:** this PR (temporarily) adds some code (that I originally adapted for KubeClient) from .NET Core that sets up the WebSocket connection (once .NET Core 2.1 is RTM most of it can be removed).
.NET Core is MIT-licensed, which is more permissive than the Apache license so as far as I am aware this is acceptable (but if anyone else thinks differently, let me know). I have included an appropriate copyright notice and link to the .NET Core license text in the file (`CoreFX.cs`) that contains this code.

I also had to make a few small changes to the Travis build process; we need the .NET Core `2.1.300-preview1` SDK to build and run tests, so this is now installed as part of the build (Travis doesn't know how to install the preview SDK so we do it manually).

There is now a mock server that can handle WebSockets requests and the first test (exec-in-pod) that exercises WebSocket connection and communication.

Would probably be good to (at some stage) add some tests that wrap the WebSockets in a `StreamDemuxer`.

I've hooked up the test server's log to Xunit's test output (to make it easier to diagnose deadlocks etc) but haven't tried to integrate the `Microsoft.Rest` logging into that yet (figured that's probably a separate PR).

A fair bit of this code (especially the test and test-infrastructure) has been copied / adapted from [KubeClient](https://github.com/tintoy/dotnet-kube-client); hopefully it'll be useful to this project :)

Closes kubernetes-client/csharp#102

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -4,9 +4,12 @@
 set -e
 
 # Ensure no compile errors in all projects
-find . -name *.csproj -exec dotnet build {} \;
+dotnet restore
+dotnet build --no-restore
 
 # Execute Unit tests
 cd tests
-dotnet restore
-dotnet test
+dotnet test --no-restore --no-build
+if [[ $? != 0 ]]; then
+    exit 1
+fi
--- a/src/WebSocketBuilder.NetCoreApp2.1.cs
+++ b/src/WebSocketBuilder.NetCoreApp2.1.cs
@@ -1,0 +1,143 @@
+#if NETCOREAPP2_1
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Net.Security;
+using System.Net.WebSockets;
+using System.Security.Authentication;
+using System.Security.Cryptography.X509Certificates;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace k8s
+{
+    /// <summary>
+    /// The <see cref="WebSocketBuilder"/> creates a new <see cref="WebSocket"/> object which connects to a remote WebSocket.
+    /// </summary>
+    public sealed class WebSocketBuilder
+    {
+        public KubernetesWebSocketOptions Options { get; } = new KubernetesWebSocketOptions();
+
+        public WebSocketBuilder()
+        {
+        }
+
+        public WebSocketBuilder SetRequestHeader(string headerName, string headerValue)
+        {
+            Options.RequestHeaders[headerName] = headerValue;
+
+            return this;
+        }
+
+        public WebSocketBuilder AddClientCertificate(X509Certificate2 certificate)
+        {
+            Options.ClientCertificates.Add(certificate);
+
+            return this;
+        }
+
+        public WebSocketBuilder ExpectServerCertificate(X509Certificate2 serverCertificate)
+        {
+            Options.ServerCertificateCustomValidationCallback = (sender, certificate, chain, sslPolicyErrors) =>
+            {
+                if (sslPolicyErrors != SslPolicyErrors.RemoteCertificateChainErrors)
+                    return false;
+
+                try
+                {
+                    using (X509Chain certificateChain = new X509Chain())
+                    {
+                        certificateChain.ChainPolicy.ExtraStore.Add(serverCertificate);
+                        certificateChain.ChainPolicy.VerificationFlags = X509VerificationFlags.AllowUnknownCertificateAuthority;
+                        certificateChain.ChainPolicy.RevocationMode = X509RevocationMode.NoCheck;
+
+                        return certificateChain.Build(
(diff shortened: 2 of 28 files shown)


tintoy:
CC: @qmfrederik @brendanburns

tintoy:
It be a good idea to split out the `netcoreapp2.1` tests into a separate test project.

tintoy:
Ugh, ok - looks like the test fails on Linux; I think for preview1, I need to explicitly opt into using the managed WebSocket implementation (by setting an environment variable).

tintoy:
Also, looks like the CI script doesn't fail if tests fail. I'll see if I can fix that.

tintoy:
Oh, it's actually 2 unrelated tests (`WatchTests.SuriveBadLine` and `WatchTests.TestWatchWithHandlers`) that are failing; interestingly, they only fail intermittently when I run them locally :)

tintoy:
CC: @tg123 - is it possible that the build server is running the tests too slowly?

tintoy:
They seem to be a little racy - they're passing now :)

tintoy:
I've modified `MockKubeApiServer` to send logs to the test output (when configured to do so) so that the logs are only shown if a test fails. Individual tests can opt into this logging by injecting `ITestOutputHelper` into the test class constructor and passing it to the `MockKubeApiServer` constructor.

tintoy:
I've tweaked some of the `WatcherTests` so they use a `CountDownEvent` to wait for incoming events.
They pass now, but one of the `ByteBufferTests` is failing instead (seems like the same issue). I'll see what I can do there as well.

tintoy:
Ok, there seem to be a couple of racy tests left, but I think I'll leave it for now; there are already more changes than I'm comfortable making in a single PR :wink:

brendandburns:
Thanks! Will review tomorrow, I kicked the CI/CD again...

tintoy:
Sorry in advance for the amount of code in this PR - once I broke the build, I just got stubborn :grin:

tintoy:
@brendanburns - FYI, it looks like the change for custom validation of server certificate may wind up making it into 2.1 after all: https://github.com/dotnet/corefx/issues/12038#issuecomment-374303655

If this is still merged (the separate code for `netstandard` vs `netcoreapp2.1` is still required), I guess it can be refactored when 2.1 goes RTM?

ci.sh
```diff
@@ -4,9 +4,12 @@
...
-find . -name *.csproj -exec dotnet build {} \;
+dotnet restore
+dotnet build --no-restore
 
 # Execute Unit tests
 cd tests
-dotnet restore
-dotnet test
+dotnet test --no-restore --no-build
+if [[ $? != 0 ]]; then
```
tg123: already set -e

tintoy: Yeah, I was wondering about that - it's supposed to stop on error (if you have set -e) right? But if I take that out then the build doesn't fail if tests fail :-/

tintoy: I'm no bash expert so it's entirely possible I've got it wrong though.

tintoy: I can't remember where, but I remember reading something a while back that said "set -e" is unreliable in some cases and you're better off using "trap" but again, I could be wrong :)

src/WebSocketBuilder.NetCoreApp2.1.cs
```diff
@@ -0,0 +1,143 @@
...
+            Options.ClientCertificates.Add(certificate);
+
+            return this;
+        }
+
+        public WebSocketBuilder ExpectServerCertificate(X509Certificate2 serverCertificate)
+        {
+            Options.ServerCertificateCustomValidationCallback = (sender, certificate, chain, sslPolicyErrors) =>
+            {
+                if (sslPolicyErrors != SslPolicyErrors.RemoteCertificateChainErrors)
```
brendandburns: nit: curly braces around all `if` ...

brendandburns:
LGTM.
This looks good enough to me and this PR has grown sufficiently large that I'd like to get it merged and then iterate...

tintoy:
Thanks! I'll open a new PR to address your feedback :)

src/WebSocketBuilder.NetCoreApp2.1.cs
tintoy: Gotcha, will make sure I do from now on :)

PR #108, opened 2018-03-21
Address feedback from kubernetes-client/csharp#103

tintoy:
Always use braces for `if` / `else`.


brendandburns:
Thanks!

tintoy:
NP - I should have paid closer attention to the style used in existing files :)

PR #114, opened 2018-03-22
Add an 'exec' example. Clean up some websockets code.

brendandburns:
@tintoy 

Thanks!

### DIFF
--- a/examples/exec/Exec.cs
+++ b/examples/exec/Exec.cs
@@ -1,0 +1,34 @@
+using System;
+using System.Threading.Tasks;
+using k8s;
+using k8s.Models;
+
+namespace exec
+{
+    internal class Exec
+    {
+        private static async Task Main(string[] args)
+        {
+            var config = KubernetesClientConfiguration.BuildConfigFromConfigFile();
+            IKubernetes client = new Kubernetes(config);
+            Console.WriteLine("Starting Request!");
+
+            var list = client.ListNamespacedPod("default");
+            var pod = list.Items[0];
+            await ExecInPod(client, pod);
+        }
+
+        private async static Task ExecInPod(IKubernetes client, V1Pod pod) {
+            var webSocket = await client.WebSocketNamespacedPodExecAsync(pod.Metadata.Name, "default", "ls", pod.Spec.Containers[0].Name);
+
+            var demux = new StreamDemuxer(webSocket);
+            demux.Start();
+
+            var buff = new byte[4096];
+            var stream = demux.GetStream(1, 1);
+            var read = stream.Read(buff, 0, 4096);
+            var str = System.Text.Encoding.Default.GetString(buff);
+            Console.WriteLine(str);
+        }
+    }
+}
--- a/src/Kubernetes.ConfigInit.cs
+++ b/src/Kubernetes.ConfigInit.cs
@@ -38,6 +38,7 @@ namespace k8s
             }
 
             CaCert = config.SslCaCert;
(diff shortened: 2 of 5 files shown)


examples/exec/Exec.cs
```diff
@@ -0,0 +1,33 @@
+using System;
+using System.Threading.Tasks;
+using k8s;
+using k8s.Models;
+
+namespace exec
+{
+    internal class Exec
+    {
+        private static void Main(string[] args)
```
tintoy: Note sure what version of C# this is built with but latest supports async main if that makes the code cleaner?

src/Kubernetes.ConfigInit.cs
```diff
@@ -59,10 +60,15 @@ public Kubernetes(KubernetesClientConfiguration config, params DelegatingHandler
                     }
 
 #if NET452
-                    ((WebRequestHandler) HttpClientHandler).ServerCertificateValidationCallback =
-                        CertificateValidationCallBack;
+                    ((WebRequestHandler) HttpClientHandler).ServerCertificateValidationCallback = (sender, certificate, chain, sslPolicyErrors) =>
```
tintoy: Noted for next time :)

src/Kubernetes.WebSocket.cs
```diff
@@ -226,18 +226,20 @@ public Task<WebSocket> WebSocketNamespacedPodAttachAsync(string name, string @na
 
 #if NETCOREAPP2_1
             if (this.CaCert != null)
+            {
                 webSocketBuilder.ExpectServerCertificate(this.CaCert);
-            else
```
tintoy: Aren't these mutually exclusive? I suppose throwing an exception would have been clearer when I first wrote this...

src/WebSocketBuilder.NetCoreApp2.1.cs
```diff
@@ -39,34 +39,10 @@ public WebSocketBuilder AddClientCertificate(X509Certificate2 certificate)
...
-                        );
-                    }
-                }
-                catch (Exception chainException)
-                {
-                    Debug.WriteLine(chainException);
-
-                    return false;
-                }
+                return Kubernetes.CertificateValidationCallBack(sender, serverCertificate, certificate, chain, sslPolicyErrors);
```
tintoy: Oops! Sorry, must have missed that :)

src/Kubernetes.WebSocket.cs
brendandburns: Maybe? I sort of lean on "the customer is always right" or "the caller is always right"

If they ship us a config with a ca Cert and noVerifyTLS who are we to judge? We should obey their wishes.

examples/exec/Exec.cs
brendandburns: Done (had to move to c# 7.1)

src/Kubernetes.WebSocket.cs
tintoy: Yeah fair enough, that reasoning makes sense to me :)

My personal approach to API design is "forgiving, but also willing to point out when the caller may be making a mistake". If someone specifies both `CaCert` and `SkipTlsVerify`, which one did they really intend to specify? Is it intuitive to them which one we'd select when there's a conflict like this, and do we document this assumption anywhere?

**But** in this case, correct behaviour is dictated by the K8s client config spec which doesn't prevent them specifying both (so I agree with you that there's not much point in stopping them from doing so).

tintoy:
(sorry, review stuff doesn't show up well on mobile :grin:)

PR #123, opened 2018-03-27
WebSocketNamespacedPodExecAsync: Support specifying command arguments

qmfrederik:
In the current implementation, `WebSocketNamespacedPodExecAsync` only allows a single command (`command` is a single string). 

The Kubernetes API allows you to pass a command with arguments, such as `/bin/bash -c "echo Hello, World!\nexit 0\n"`.

This PR changes the `command` argument to be an `IEnumerable<string>`, allowing you to pass parameters.

### DIFF
--- a/src/IKubernetes.WebSocket.cs
+++ b/src/IKubernetes.WebSocket.cs
@@ -52,6 +52,53 @@ namespace k8s
         /// <return>
         /// A <see cref="ClientWebSocket"/> which can be used to communicate with the process running in the pod.
         /// </return>
+        Task<WebSocket> WebSocketNamespacedPodExecAsync(string name, string @namespace = "default", string command = null, string container = null, bool stderr = true, bool stdin = true, bool stdout = true, bool tty = true, Dictionary<string, List<string>> customHeaders = null, CancellationToken cancellationToken = default(CancellationToken));
+
+        /// <summary>
+        /// Executes a command in a pod.
+        /// </summary>
+        /// <param name='name'>
+        /// name of the Pod
+        /// </param>
+        /// <param name='namespace'>
+        /// object name and auth scope, such as for teams and projects
+        /// </param>
+        /// <param name='command'>
+        /// Command is the remote command to execute. argv array. Not executed within a
+        /// shell.
+        /// </param>
+        /// <param name='container'>
+        /// Container in which to execute the command. Defaults to only container if
+        /// there is only one container in the pod.
+        /// </param>
+        /// <param name='stderr'>
+        /// Redirect the standard error stream of the pod for this call. Defaults to
+        /// <see langword="true"/>.
+        /// </param>
+        /// <param name='stdin'>
+        /// Redirect the standard input stream of the pod for this call. Defaults to
+        /// <see langword="true"/>.
+        /// </param>
+        /// <param name='stdout'>
+        /// Redirect the standard output stream of the pod for this call. Defaults to
+        /// <see langword="true"/>.
+        /// </param>
+        /// <param name='tty'>
+        /// TTY if true indicates that a tty will be allocated for the exec call.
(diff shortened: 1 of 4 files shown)


src/Kubernetes.WebSocket.cs
```diff
@@ -67,16 +72,24 @@ public Task<WebSocket> WebSocketNamespacedPodExecAsync(string name, string @name
...
-            uriBuilder.Query = QueryHelpers.AddQueryString(string.Empty, new Dictionary<string, string>
+            foreach (var c in command)
             {
-                { "command", command},
-                { "container", container},
-                { "stderr", stderr ? "1": "0"},
-                { "stdin", stdin ? "1": "0"},
-                { "stdout", stdout ? "1": "0"},
-                { "tty", tty ? "1": "0"}
-            }).TrimStart('?');
```
tg123: that `trimstart ?` is because on net452, this api will NOT return with a `?` prefix (different from netstanard)

src/Kubernetes.WebSocket.cs
```diff
@@ -67,16 +72,24 @@ public Task<WebSocket> WebSocketNamespacedPodExecAsync(string name, string @name
 
             uriBuilder.Path += $"api/v1/namespaces/{@namespace}/pods/{name}/exec";
 
+            List<string> queryParameters = new List<string>();
```
tg123: I assume you want multi commands param, but I suggest using `queryhelper`

queryhelper cloud help you add multiple params
```
    var q = QueryHelpers.AddQueryString(string.Empty, new Dictionary<string, string>
            {
                {"a", "b"}
            });

            q = QueryHelpers.AddQueryString(q, "command", "1");
            q = QueryHelpers.AddQueryString(q, "command", "1");
            q = QueryHelpers.AddQueryString(q, "command", "2");
            Console.WriteLine(q); // ?a=b&command=1&command=1&command=2
```

src/Kubernetes.WebSocket.cs
```diff
@@ -37,6 +37,11 @@ public Task<WebSocket> WebSocketNamespacedPodExecAsync(string name, string @name
                 throw new ArgumentNullException(nameof(command));
             }
 
+            if (command.Count() == 0)
```
tg123: suggest: use `Any()` instead of `Count()`

qmfrederik:
@tg123 Thanks for the prompt review! I made the changes requested, let me know if it looks OK now.

src/Kubernetes.WebSocket.cs
brendandburns: Yeah, let's maintain this. (and is this covered by a test-case? if it's not it should be...)

src/IKubernetes.WebSocket.cs
```diff
@@ -52,7 +52,7 @@ public partial interface IKubernetes
         /// <return>
         /// A <see cref="ClientWebSocket"/> which can be used to communicate with the process running in the pod.
         /// </return>
-        Task<WebSocket> WebSocketNamespacedPodExecAsync(string name, string @namespace = "default", string command = "/bin/bash", string container = null, bool stderr = true, bool stdin = true, bool stdout = true, bool tty = true, Dictionary<string, List<string>> customHeaders = null, CancellationToken cancellationToken = default(CancellationToken));
```
brendandburns: Let's maintain this signature and add one that takes an `IEnumerable` I don't want to break the API contract for users of this library.

qmfrederik:
OK - I added overloads that take a `string` and `IEnumerable<string>`, so we keep the API stable. Also addressed the other feedback.

There's a unit test for `WebSocketNamespacedPodExecAsync`, not sure if you run unit tests on `net451`, though. (You can set up a CI build for `net451` using AppVeyor)

Let me know if you need anything else.

brendandburns:
Thanks!

PR #124, opened 2018-03-28
Eliminate all arbitrary delays, replace with synchronization.

brendandburns:
@tg123 @tintoy 

Deflake the WatchTests since they're failing a bunch in CI.

### DIFF
--- a/tests/WatchTests.cs
+++ b/tests/WatchTests.cs
@@ -91,19 +91,10 @@ namespace k8s.Tests
                 httpContext.Response.ContentLength = null;
 
                 await WriteStreamLine(httpContext, MockKubeApiServer.MockPodResponse);
-                await Task.Delay(TimeSpan.FromMilliseconds(100));
-
                 await WriteStreamLine(httpContext, MockBadStreamLine);
-                await Task.Delay(TimeSpan.FromMilliseconds(100));
-
                 await WriteStreamLine(httpContext, MockAddedEventStreamLine);
-                await Task.Delay(TimeSpan.FromMilliseconds(100));
-
                 await WriteStreamLine(httpContext, MockBadStreamLine);
-                await Task.Delay(TimeSpan.FromMilliseconds(100));
-
                 await WriteStreamLine(httpContext, MockModifiedStreamLine);
-                await Task.Delay(TimeSpan.FromMilliseconds(100));
 
                 // make server alive, cannot set to int.max as of it would block response
                 await Task.Delay(TimeSpan.FromDays(1));
@@ -116,7 +107,6 @@ namespace k8s.Tests
                 });
 
                 var listTask = client.ListNamespacedPodWithHttpMessagesAsync("default", watch: true).Result;
-
 
                 var events = new HashSet<WatchEventType>();
                 var errors = 0;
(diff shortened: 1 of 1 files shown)


tests/WatchTests.cs
```diff
@@ -91,19 +91,10 @@ public void SuriveBadLine()
                 httpContext.Response.ContentLength = null;
 
                 await WriteStreamLine(httpContext, MockKubeApiServer.MockPodResponse);
-                await Task.Delay(TimeSpan.FromMilliseconds(100));
```
tg123: those delays are to simulate that server is emitting messages

tests/WatchTests.cs
```diff
@@ -281,14 +266,11 @@ public void WatchServerDisconnect()
...
             Watcher<V1Pod> watcher;
             Exception exceptionCatched = null;
+            var exceptionReceived = new CountdownEvent(1);
 
             using (var server = new MockKubeApiServer(TestOutput, async httpContext =>
             {
                 await WriteStreamLine(httpContext, MockKubeApiServer.MockPodResponse);
-
-                // make sure watch success
-                await Task.Delay(TimeSpan.FromMilliseconds(200));
```
tg123: remove this delay would cause watch failed
exceptions in https://travis-ci.org/kubernetes-client/csharp/builds/359210626?utm_source=github_status&utm_medium=notification

tests/WatchTests.cs
```diff
@@ -355,11 +343,17 @@ public void TestWatchWithHandlers()
                 var events = new HashSet<WatchEventType>();
 
                 var watcher = listTask.Watch<V1Pod>(
-                    (type, item) => { events.Add(type); }
+                    (type, item) => {
+                        events.Add(type);
+                        eventsReceived.Signal();
```
tg123: here might be some bad practice
https://www.jetbrains.com/help/resharper/2018.1/AccessToDisposedClosure.html

tests/WatchTests.cs
```diff
@@ -326,13 +316,11 @@ private class DummyHandler : DelegatingHandler
         [Fact]
         public void TestWatchWithHandlers()
         {
+            using (CountdownEvent eventsReceived = new CountdownEvent(1))
```
tg123: suggest remove using due eventsReceived is referenced by closure.

tg123: shortcut for `countdown latch 1` is AutoresetEvent/MaunalresetEvent

tests/WatchTests.cs
tintoy: In this case I think that rule may not be correct - the watcher is in the same parent scope as the closure and should be disposed before the wait handle?

tests/WatchTests.cs
tintoy: Using means disposal is in reverse order of declaration so I think this should be safe.

tests/WatchTests.cs
tg123: Agree.
but, in best practice, the code, `watcher.cs`, might be changed and those conditions would be false someday.

I suggest follow some strict static checking even it is not a bug.

tintoy: Yeah fair enough - that's a good reason 😊

Should make sure that the wait handle is disposed deterministically at some point though (perhaps at the end of the test?) - otherwise you have to rely on GC to avoid leaking native handles.

tests/WatchTests.cs
brendandburns: But why do we need delays? At the end of the day it's just a byte stream, right?

tests/WatchTests.cs
brendandburns: We shouldn't need delays. I will fix the code to make sure that this can work...

tests/WatchTests.cs
brendandburns: Or I can move to Monitor instead of the CountdownEvent.

tests/WatchTests.cs
brendandburns: Agree, I believe this shoudl be safe as the watched will be deleted before the eventsReceived object.

tests/WatchTests.cs
tintoy: How will you count the required number of events?

brendandburns:
I believe synchronization is in the correct places now, and the tests are passing, I'm merging...

PR #125, opened 2018-03-28
revert changes to ci.sh in c0a42ad8847f1401ec0ffa2786e3f710a72699f0

tg123:
hi @tintoy 

 * `find . -name *.csproj -exec dotnet build {} \; ` should not be removed, it is to ensure all examples could be compiled before code check in.
 * no need to test $? because  `set -e` is already set.

any comments?

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -4,12 +4,9 @@
 set -e
 
 # Ensure no compile errors in all projects
-dotnet restore
-dotnet build --no-restore
+find . -name *.csproj -exec dotnet build {} \;
 
 # Execute Unit tests
 cd tests
-dotnet test --no-restore --no-build
-if [[ $? != 0 ]]; then
-    exit 1
-fi
+dotnet restore
+dotnet test


tintoy:
I'm pretty sure you don't need the `find` command because there's a `.sln` file that references both library and test project (so simply calling `dotnet build` from the root directory will already build both).

tintoy:
Ah, do you mean the sample projects are not part of the solution? It might be simpler to just ensure that all projects are part of the solution (makes working with both the `dotnet` CLI and Visual Studio much simpler).

tintoy:
But whatever you guys want to use, I probably should have just stuck with that; I just got sucked down the rabbit hole while trying to fix the build :wink:

brendandburns:
I'm happy to take a change to add those projects to the sln file. I use VS Code and it doesn't really know what to do with SLN files (at least as far as I can tell)

tintoy:
From memory (I've previously done some work that's tangentially-related), OmniSharp is happy to open any top-level project file that MSBuild can understand (which includes `.sln` files).

There's a command called `OmniSharp: Select Project` (you can also click the name next to the flame on the status bar to trigger it). That should show all project files as well as the solution file (which will work with all of them together).

tintoy:
I'm happy to do that in a separate PR (if you use `dotnet sln add` it even sets up solution folders to mimic your physical folder structure).

But either way, sorry for simply jumping in and changing the existing build script without asking; I just got a little stubborn after the build started failing :grin:

tintoy:
BTW, I could do that change as part of #127 to keep things simple.

brendandburns:
I'm going to merge this, and then we can look at the SLN based PR in the context of #127

PR #126, opened 2018-03-28
Fix YAML serialization of ResourceQuantity values

qmfrederik:
We recently upgraded from version 0.3 to 0.6 and noticed that reading YAML files which contain resource quantities (such as pods with resource constraints) would fail.

The difference appears to be that in 0.3, resource quantities were treated as `string` values and in 0.6 they are treated as first-class citizens.

However, no YAML serialization/deserialization was present for ResourceQuantity objects and hence (de)serialization would fail.

This PR fixes that problem by implementing the `IYamlConvertible` interface.

Unit tests are included; to be able to test serialization I also added the `Yaml.WriteToString` method.

### DIFF
--- a/src/ResourceQuantity.cs
+++ b/src/ResourceQuantity.cs
@@ -4,6 +4,9 @@ using System.Linq;
 using System.Numerics;
 using Fractions;
 using Newtonsoft.Json;
+using YamlDotNet.Core;
+using YamlDotNet.Core.Events;
+using YamlDotNet.Serialization;
 
 namespace k8s.Models
 {
@@ -11,7 +14,7 @@ namespace k8s.Models
     {
         public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
         {
-            var q = (ResourceQuantity) value;
+            var q = (ResourceQuantity)value;
 
             if (q != null)
             {
@@ -84,7 +87,7 @@ namespace k8s.Models
     ///     cause implementors to also use a fixed point implementation.
     /// </summary>
     [JsonConverter(typeof(QuantityConverter))]
-    public partial class ResourceQuantity
+    public partial class ResourceQuantity : IYamlConvertible
     {
         public enum SuffixFormat
         {
@@ -176,7 +179,15 @@ namespace k8s.Models
         // ctor
         partial void CustomInit()
         {
-            var value = (Value ?? "").Trim();
+            if (Value == null)
+            {
+                // No value has been defined, initialize to 0.
+                _unitlessValue = new Fraction(0);
+                Format = SuffixFormat.BinarySI;
+                return;
+            }
+
+            var value = Value.Trim();
 
             var si = value.IndexOfAny(SuffixChars);
             if (si == -1)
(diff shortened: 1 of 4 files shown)


src/ResourceQuantity.cs
```diff
@@ -175,8 +178,16 @@ public string CanonicalizeString(SuffixFormat suffixFormat)
 
         // ctor
         partial void CustomInit()
-        {
-            var value = (Value ?? "").Trim();
+        {
+            if (Value == null)
```
tg123: `new ResourceQuantity()` and `''` are not valid in previous design 
so, you want to make `new ResourceQuantity() === 0`?

src/Yaml.cs
```diff
@@ -30,5 +30,15 @@ public class Yaml {
...
             return obj;
         }
+
+        public static string SaveToString<T>(T value)
+        {
+            var serializer =
+                new SerializerBuilder()
+                .WithNamingConvention(new CamelCaseNamingConvention())
+                .Build();
+            var obj = serializer.Serialize(value);
```
tg123: I guess you may want `BuildValueSerializer`

tests/QuantityValueTests.cs
```diff
@@ -233,6 +233,20 @@ public void Serialize()
...
+        {
+            var value = Yaml.LoadFromString<ResourceQuantity>("\"1\"");
+            Assert.Equal(new ResourceQuantity(1, 0, DecimalSI), value);
+        }
+
+        [Fact]
+        public void SerializeYaml()
+        {
+            var value = Yaml.SaveToString(new ResourceQuantity(1, -1, DecimalSI));
+            Assert.Equal("100m\r\n...\r\n", value);
```
tg123: `BuildValueSerializer` will not emit document end.
or
`Environment.NewLine` ?

src/ResourceQuantity.cs
qmfrederik: `new ResourceQuantity()` would throw `System.FormatException: 'The string '' cannot be converted to fraction.'` in the previous version. I think that's a very confusing thing to happen, as the constructor doesn't take any parameter.

Either you allow for a parameter-less constructor, which doesn't throw, or you make the constructor private, so the parameter is not callable.

Calling `new ResourceQuantity('')` will throw the same exception both before and with the changes in this PR. I've added a unit test to make that behavior explicit.

That said, the way the YAML deserialize seems to work is that it creates a new instance of your object using the default constructor and then calls `Read` on that instance. That won't work if the default constructor throws, so that's why I made this change.

tests/QuantityValueTests.cs
qmfrederik: I changed the code to use `BuildValueSerializer()`

qmfrederik:
@tg123 Thanks for the review, I updated the PR with the changes you requested.

Regarding to the code editor - I use Visual Studio with pretty much the default code settings. You could add a [EditorConfig](https://docs.microsoft.com/en-us/visualstudio/ide/create-portable-custom-editor-options) file to this project if there's a specific code style you want to enforce.

tintoy:
I reckon editorconfig is a good idea, too - I'm pretty sure I made some changes that it would have helped catch 😊

src/ResourceQuantity.cs
tg123: agree.

src/Yaml.cs
```diff
@@ -17,18 +20,35 @@ public class Yaml {
...
+        {
+            var stringBuilder = new StringBuilder();
+            var writer = new StringWriter(stringBuilder);
+            var emitter = new Emitter(writer);
+
+            var serializer =
+                new SerializerBuilder()
+                .WithNamingConvention(new CamelCaseNamingConvention())
+                .BuildValueSerializer();
+            emitter.Emit(new StreamStart());
```
tg123: necessary?

qmfrederik: The StreamStart thing? Yes, otherwise the emitter will throw an exception, unfortunately.

PR #128, opened 2018-03-30
Don't throw in StreamDemuxer.Dispose

qmfrederik:
In `StreamMuxer.Dispose`,  we cancel the run loop and wait for the run loop to complete.
If the run loop exists with an exception, `this.runloop.Wait()` will cause that exception to be rethrown.
However, Dispose methods should never throw - see https://msdn.microsoft.com/en-us/library/bb386039.aspx.

This PR adds a try/catch statement.


brendandburns:
LGTM.

Issue #118, opened 2018-03-25
Test certificates are not longer valid for MacOS and .NET Core 2.X


sesispla:
I've found the following error executing the tests locally, with MacOS 10.12.6 (High Sierra) and latest .NET Core SDK (2.1.300-preview1-008174):

```
Failed   k8s.Tests.AuthTests.Cert
Error Message:
 Interop+AppleCrypto+AppleCommonCryptoCryptographicException : MAC verification failed during PKCS12 import (wrong password?)
Stack Trace:
   at Interop.AppleCrypto.X509ImportCertificate(Byte[] bytes, X509ContentType contentType, SafePasswordHandle importPassword, SafeKeychainHandle keychain, Boolean exportable, SafeSecIdentityHandle& identityHandle)
   at Internal.Cryptography.Pal.CertificatePal.FromBlob(Byte[] rawData, SafePasswordHandle password, X509KeyStorageFlags keyStorageFlags)
   at System.Security.Cryptography.X509Certificates.X509Certificate..ctor(Byte[] data)
   at System.Security.Cryptography.X509Certificates.X509Certificate2..ctor(Byte[] rawData)
   at k8s.Tests.AuthTests.Cert() in /Users/Sergio/Repositories/kubernetes/kubernetes-client-csharp/tests/AuthTests.cs:line 171
```

Accordingly to @qmfrederik in [corefx/24225](https://github.com/dotnet/corefx/issues/24225) _"If you have a PKCS#12 file which is not protected with a password, and which does have a MAC entry, opening the file will work on Windows and Linux but fails on Mac (which use CommonCrypto)."_

Maybe our test certificates can be regenerated with password protection.

tg123:
I do not have a mac, cloud you please test if code below works?

CertUtils.cs:92

```
                const string password = "k8s";
                store.Save(pkcs, password.ToCharArray(), new SecureRandom());
                return new X509Certificate2(pkcs.ToArray(), password);
```

sesispla:
Cool, we're almost there. Only one test failing now:

```
Failed   k8s.Tests.AuthTests.Cert
Error Message:
 Interop+AppleCrypto+AppleCommonCryptoCryptographicException : MAC verification failed during PKCS12 import (wrong password?)
Stack Trace:
[...]

PR #129, opened 2018-03-30
Run CI scripts on both Mac and Linux

qmfrederik:
This should help with #118 by running the CI scripts on both Mac and Linux.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ os:
 language: csharp
 mono: none
 env:
-  - DOTNET_SKIP_FIRST_TIME_EXPERIENCE=1 DOTNET_CLI_TELEMETRY_OPTOUT=1
+  - DOTNET_SKIP_FIRST_TIME_EXPERIENCE=1 DOTNET_CLI_TELEMETRY_OPTOUT=1 COMPlus_UseManagedHttpClientHandler=true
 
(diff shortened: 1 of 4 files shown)


qmfrederik:
So, after a bit of investigation:
- The build & tests now run on both Linux & macOS and all certificate-related tests pass.
- The CI build currently is failing because of the `WatchTests`, which I believe to be a flaky test. I think it's addressed in #124.
- There seems to be a problem with custom certificate handling on macOS. According to https://github.com/dotnet/corefx/issues/19718, https://github.com/dotnet/corefx/issues/27000 and https://github.com/dotnet/corefx/issues/26446, the only reliable fix is to use the `SocketHttpHandler`. On [.NET Core 2.1 Preview 1](https://blogs.msdn.microsoft.com/dotnet/2018/02/27/announcing-net-core-2-1-preview-1/), the only relieable way to use `SocketHttpHandler` is to to set `COMPlus_UseManagedHttpClientHandler` to `true`, I've updated the CI process accordingly. In a later preview, we can use `new HttpClient(new SocketsHttpHandler())`. I'll open a issue to track that.

brendandburns:
So I merged #131 so I think this is now obsoleted?

qmfrederik:
Yes, thx!

PR #131, opened 2018-03-30
Run integration tests as part of CI (after #129)

qmfrederik:
We spent some time paying off our technical debt this week. That included upgrading to the latest version of the C# Kubernetes client, Kubernetes and .NET Core 2.1 Preview.

Quite a few things broke, especially in stuff related to WebSockets and SSL. Things broke in different ways on Windows, Mac and Linux.

Hence, we thought it might be useful to add integration tests to the kubernetes-csharp client. This PR adds that for Travis on Linux, using minikube. I assume we can replicate the same behavior on macOS, too, but that's not included here.

It'll take a couple of commits to get things right, so please bear with me.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,3 +25,6 @@ before_install:
 
 script:
   - ./ci.sh
+
+after_script:
+  - if [ "$TRAVIS_OS_NAME" = "linux" ]; then ./integration-tests.sh; fi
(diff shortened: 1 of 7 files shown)


qmfrederik:
This seems to be OK now. I can't run the exec & attach demos (yet) as they run endless loops.
The tests uncovered an issue in the WebSocket client, I'll create a separate PR for that.

brendandburns:
Thanks!

Issue #120, opened 2018-03-26
Get access to entire kube config


billpratt:
I'm working on a small POC and would like access to entire kube config. Any reason why we couldn't have a public method that uses [this](https://github.com/kubernetes-client/csharp/blob/master/src/KubernetesClientConfiguration.ConfigFile.cs#L269)?

```
public static K8SConfiguration LoadKubeConfig(string kubeconfigPath = null)
{
    var fileInfo = new FileInfo(kubeconfigPath ?? KubeConfigDefaultLocation);
    
    return LoadKubeConfig(fileInfo);
}

private static K8SConfiguration LoadKubeConfig(FileInfo kubeconfig) { ... }
````

This would be similar to how the ```kubectl``` go package gets the config using [this](https://github.com/kubernetes/client-go/blob/master/tools/clientcmd/loader.go#L95) and [this](https://github.com/kubernetes/client-go/blob/master/tools/clientcmd/config.go#L84)

brendandburns:
This seems totally fine. Please send a PR to make those methods public?

billpratt:
Will do

brendandburns:
Closing since #133 is merged.

PR #133, opened 2018-04-02
Make LoadKubeConfig public and refactor to use YAML helper methods

billpratt:
Addressed #120 

@brendanburns With this PR I also refactored to use the YAML helper methods you added recently.


src/KubeConfigModels/ClusterEndpoint.cs
```diff
@@ -11,13 +11,13 @@ public class ClusterEndpoint
         /// <summary>
         /// Gets or sets the path to a cert file for the certificate authority.
         /// </summary>
-        [YamlMember(Alias = "certificate-authority")]
+        [YamlMember(Alias = "certificate-authority", ApplyNamingConventions = false)]
```
billpratt: ```ApplyNamingConventions = false``` is need when using naming conventions like CamelCaseNamingConvention

src/KubernetesClientConfiguration.ConfigFile.cs
```diff
@@ -57,7 +57,7 @@ public partial class KubernetesClientConfiguration
 
         /// <summary>
         /// </summary>
-        /// <param name="kubeconfig">Fileinfo of the kubeconfig, cannot be null, whitespaced or empty</param>
+        /// <param name="kubeconfig">String representation of kubeconfig contents, cannot be null, whitespaced or empty</param>
         /// <param name="currentContext">override the context in config file, set null if do not want to override</param>
         /// <param name="masterUrl">overrider kube api server endpoint, set null if do not want to override</param>
         public static KubernetesClientConfiguration BuildConfigFromConfigFile(string kubeconfig,
```
billpratt: I couldn't find any references to this method. Can we deprecate?

brendandburns: Sure, seems reasonable.

src/KubernetesClientConfiguration.ConfigFile.cs
```diff
@@ -259,53 +259,56 @@ private void SetUserDetails(K8SConfiguration k8SConfig, Context activeContext)
...
         }
 
         /// <summary>
-        ///     Loads Kube Config from stream.
+        ///     Loads Kube Config
         /// </summary>
         /// <param name="kubeconfig">Kube config file contents</param>
         /// <returns>Instance of the <see cref="K8SConfiguration"/> class</returns>
-        private static K8SConfiguration LoadKubeConfig(Stream kubeconfig)
+        public static K8SConfiguration LoadKubeConfig(FileInfo kubeconfig)
```
brendandburns: Let's keep this as a stream, since we want to support loading from network streams as well as files.

tests/KubernetesClientConfigurationTests.cs
```diff
@@ -322,6 +324,76 @@ public void AsUserExtra()
...
+            var expectedCfg = Yaml.LoadFromString<K8SConfiguration>(txt);
+
+            var cfg = KubernetesClientConfiguration.LoadKubeConfig("assets/kubeconfig.yml");
+
+            Assert.NotNull(cfg);
+
+            ConfigsEqual(expectedCfg, cfg);
+        }
+
+        private void ConfigsEqual(K8SConfiguration expected, K8SConfiguration actual)
```
brendandburns: This function is too long, I think. Factor it so there is an

`assertContextEqual(...)`

`assertUserEqual(...)`

`assertClusterEqual(...)`

set of methods and refactor the for loops to just call those methods?

brendandburns:
A few small comments, but basically LGTM.

src/KubernetesClientConfiguration.ConfigFile.cs
billpratt: Mark as [Obsolete] or remove?

src/KubernetesClientConfiguration.ConfigFile.cs
billpratt: Good catch. I'll add it back in and use Yaml.LoadFromStreamAsync. Should we keep the method with FileInfo kubeconfig?

Essentially, we would have these method signatures.
```
public static async Task<K8SConfiguration> LoadKubeConfigAsync(string kubeconfigPath = null)
public static K8SConfiguration LoadKubeConfig(string kubeconfigPath = null)
public static async Task<K8SConfiguration> LoadKubeConfigAsync(FileInfo kubeconfig)
public static K8SConfiguration LoadKubeConfig(FileInfo kubeconfig)
public static async Task<K8SConfiguration> LoadKubeConfigAsync(Stream kubeconfigStream)
public static K8SConfiguration LoadKubeConfig(Stream kubeconfigStream)
```

brendandburns: Your call about `FileInfo`. it's easy enough to turn a file name into a `Stream`

brendandburns:
We just merged a file move in #134 you'll need to rebase (but it should "just work")

billpratt:
@brendanburns I believe I've addressed everything.

brendandburns:
LGTM, thanks!

Issue #127, opened 2018-03-29
Move library project to subdirectory of /src


tintoy:
Does anybody have any objections to me moving the library project into a subdirectory of `/src`?

As mentioned to @brendanburns, I'd like to add a second library that adds support for a `Microsoft.Extensions.Configuration` provider targeting `Secret`s and `ConfigMap`s. This enables you to write .NET Core apps that natively retrieve their configuration (and can automatically reload that configuration if the underlying resource changes).

Here's an example using an implementation I did elsewhere if that helps:

https://github.com/tintoy/dotnet-kube-client/blob/dc323cb27938d417202b9217d3750fe7f05f39b8/samples/ConfigFromConfigMap/Program.cs#L42-L117

tintoy:
CC: @tg123 @qmfrederik 

This will use the official client, of course :wink:

brendandburns:
This seems fine to me. It doesn't really have any implications except for updating scripts etc. correct?

tintoy:
Yep! It may generate some merge conflicts for any other PRs we have in flight though :)

brendandburns:
We're in an ok-ish place now, I think, and actually I find git is pretty good about tracking moves like this..

Fire away...

PR #134, opened 2018-04-03
Move projects to subdirectories of /src and /tests

tintoy:
Resolves kubernetes-client/csharp#127.


brendandburns:
Did you use `git mv` to do this? This shows up as delete/add which makes it harder for git to understand what happened...

Thanks

tintoy:
I didn't realise there was a distinction! No worries will redo :)

brendandburns:
LGTM, thanks.

PR #141, opened 2018-04-24
Support relative paths in Kubernetes configuration files

qmfrederik:
I have a Kubernetes configuration file which looks like this:

```
apiVersion: v1
clusters:
- cluster:
    certificate-authority: ca.crt
    server: https://127.0.0.1:8443
  name: minikube
```

The current implementation will attempt to locate the `ca.crt` file in the current working directory. However, `ca.crt` is a path relative to the Kubernetes configuration file.

Thus, loading this file using the C# client will fail if the configuration file is located in a directory different from the working directory.

This PR changes the code so that:
- The `K8SConfiguration` object is aware of the path of the file from which it was loaded. A new `FileName` property has been added. If the `K8sConfiguration` object has been created in memory, from a stream,... the value of `Filename` will be `null`.
- The `KuberentesClientConfiguration` class will attempt to resolve relative paths using the `K8SConfiguration.FileName` property.

I've verified that the updated behavior matches that of `kubectl`.


brendandburns:
This is breaking a bunch of tests...

I think there's an interesting question of if this is a breaking change or not.  I think that this change's implementation is more "correct" in aligning with the way kubectl behaves, but it may also break some users.

At the very least we should note that this is a breaking change, and we should probably enable a boolean to opt out of this behavior...

qmfrederik:
@brendandburns Thanks, I added a mechanism to opt-out of the resolving of the relative paths by adding an optional boolean flag to the `BuildFromConfigFile` methods.

I fixed most of the unit tests by having them opt out of the new behavior.

I believe the behavior of kubectl is "more correct" and it would make sense for the C# client to mimic kubectl's behavior; but users can now override that choice.

brendandburns:
LGTM, thanks.

PR #142, opened 2018-04-24
Add ChannelIndex enumeration, and some documentation

qmfrederik:
The connection used for exec, attach and the like multiplexes multiple channels (such as stdin, stdout,...) over a single stream.

I found myself having to constantly lookup which index was used for which channel, so this PR tries to put an end to that by adding a `ChannelIndex` enumeration which allows you to write more descriptive code.

At the same time, I also added some docs for the `StreamDemuxer` and `MuxedStream` classes.


brendandburns:
LGTM, thanks!

PR #143, opened 2018-04-25
Don't throw an exception on MuxedStream.Flush

qmfrederik:
The `MuxedStream` class currently throws a `NotSupportedException` on `Flush`.

However, since all data which is written to this stream is immediately sent to the remote server (e.g. there is no local buffering/queueing), `Flush` can be implemented as a no-op.

This fix is useful when people use other classes, such as `StreamWriter`, on top of a `MuxedStream`. `StreamWriter` will call `.Flush` so throwing there would break things.


brendandburns:
LGTM.

PR #144, opened 2018-04-25
Fix flakey tests by making them async

qmfrederik:
Some of the tests (especially the watch-related tests) were flaky, failing at random times.

This PR tries to fix the flaky tests by:
1. Making them async
2. Using synchronization code (such as semaphores) instead of sleeps when one block of code needs to wait for another one.

As a side-effect, the `TestBase` class which was artificially setting a synchronization context can also be removed. There is no guarantee that the C# client for Kubernetes will execute at runtime in an environment with a synchronization context (it is very likely it will not), and the test should reflect that.

I've also updated the version of the test SDK to the latest released version, and added a reference to Nito.AsyncEx to the test project.


qmfrederik:
Thanks, I've standardized the timeouts on 15 seconds. The tests take ~200ms at most on my machine, so the timeout should be large enough to tolerate slow machines.

brendandburns:
Merging (though I've found that cloud build/test infra can be remarkably slow at times :)

PR #147, opened 2018-04-25
Add support for watching individual objects.

qmfrederik:
This adds support for the `/api/v1/watch/[namespaces/{namespace}]/{objectType}/{name}` watchers.
Currently includes for v1 objects only.

The heavy lifting is done in the `WatchObjectAsync<T>` method, so support for  other objects can be added fairly easily.

### DIFF
--- a/src/KubernetesClient/IKubernetes.Watch.cs
+++ b/src/KubernetesClient/IKubernetes.Watch.cs
@@ -1,0 +1,748 @@
+using k8s.Models;
+using System;
+using System.Collections.Generic;
+using System.Threading;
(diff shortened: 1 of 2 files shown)


qmfrederik:
Note to self - this doesn't handle the scenario where the server responds with a 410:

```
{"type":"ERROR","object":{"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"too old resource version: 25858 (41294)","reason":"Gone","code":410}}
```

The 410 response is not documented in the Kubernetes Swagger API.

PR #148, opened 2018-04-27
Auto-generate Watcher code (for individual objects)

qmfrederik:
I needed a watcher for another object, and instead of manually generating all the Watcher methods, I created a code generator for them.

This PR:
- Keeps only one method in the manually-generated watch code, `WatchObjectAsync<T>(string path, (...));`. You can pass any type and path you want, and it'll try and generate a Watcher for that type at that URL. This should help anyone who needs a watcher which is not (yet) implemented
- Adds a code generator for the watchers for individual methods (lists are excluded). This uses the v1.10 version of the Kubernetes API.
- Adds the generated code to source control.

The individual commits reflect these steps.


qmfrederik:
If there's a way to only get the C# part in the generator project (and do it in C#), sure, why not.

But I'm not fluent enough in the other languages to add a code generator for the watchers in those languages 😄 .

gen/KubernetesWatchGenerator/Program.cs
```diff
@@ -0,0 +1,250 @@
...
+using System.Threading.Tasks;
+
+namespace KubernetesWatchGenerator
+{
+    class Program
+    {
+        static async Task Main(string[] args)
+        {
+            // Download the latest spec
+            const string kubernetesBranch = "v1.10.0";
```
brendandburns: This is currently controlled by the `settings` file in the repo here:

https://github.com/kubernetes-client/csharp/blob/master/csharp.settings

We should read from that environment variable (or from that file) for consistency.

gen/KubernetesWatchGenerator/Program.cs
```diff
@@ -0,0 +1,250 @@
...
+                o => o.Path.Contains("/watch/")
+                && o.Operation.ActualParameters.Any(p => p.Name == "name")
+                && !blacklistedOperations.Contains(o.Operation.OperationId)).ToArray();
+
+            using (Stream interfaceFile = File.Open($"{outputDirectory}IKubernetes.Watch.cs", FileMode.Create, FileAccess.ReadWrite))
+            using (StreamWriter interfaceWriter = new StreamWriter(interfaceFile, Encoding.UTF8))
+            using (Stream classFile = File.Open($"{outputDirectory}Kubernetes.Watch.cs", FileMode.Create, FileAccess.ReadWrite))
+            using (StreamWriter classWriter = new StreamWriter(classFile, Encoding.UTF8))
+            {
+                interfaceWriter.WriteLine("using k8s.Models;");
```
brendandburns: Rather than this can we use a templating library, for example this one:

https://github.com/wildbit/mustachio

brendandburns:
This is great!  I think if you move to a templating library like the one I suggest, then it will be easier to iterate on this (and also easier to add additional language support and move it into the `gen` repo...)

qmfrederik:
@brendandburns Done, I've pushed updates so that the code now uses templating (via Nustache) and re-generated the classes.

brendandburns:
LGTM, thanks!

PR #149, opened 2018-04-27
Use automatic versioning

qmfrederik:
Currently, the version of the C# client is updated manually by editing `KubernetesClient.csproj`.

That's fine, but if you're working with interim versions of the package you'll need to manually update the version every time you push a new version to a feed.

This PR updates the process so that it uses [NerdBank.GitVersioning](https://github.com/AArnott/Nerdbank.GitVersioning) to automatically generate version numbers.

Roughly, the idea is:
1. In `version.json`, you specify the 'official' version (such as 0.7.0-beta)
2. If you build from a clean checkout of master, you'll get a version number like 0.7.0-beta.{gitHeight}, where gitHeight is the number of commits to master since you last changed the version number.
3. If you build from anything else, the Git SHA commit will be added

For bonus points, the NerdBank.GitVersioning will always inject certain information, such as the Git commit ID, as assembly attributes, so you can always trace back a version of the .dll to a Git commit.

The dependency is marked as `PrivateAssets="all"` so that it is a development dependency only; consumers of this library do not need to install NerdBank.GitVersioning.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,10 @@
 sudo: required
 dist: trusty
 
+# Don't do shallow clones
+git:
+  depth: false
+
 os:
   - osx
   - linux
--- a/src/KubernetesClient/KubernetesClient.csproj
+++ b/src/KubernetesClient/KubernetesClient.csproj
@@ -1,6 +1,5 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <PackageVersion>0.7.0-beta</PackageVersion>
     <Authors>The Kubernetes Project Authors</Authors>
     <Copyright>2017 The Kubernetes Project Authors</Copyright>
     <Description>Client library for the Kubernetes open source container orchestrator.</Description>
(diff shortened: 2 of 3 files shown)


tg123:
image

I had setup a daily build on VSTS with auto version, it breaks after runtime 2.1 introduced.

[THIS IS A MS PRODUCT AD]

and another concern is that the build system should read the git tag from the building code, not pulling it from remote.

qmfrederik:
In the end, it's a philosophical question, but I do think there's a lot of merit in determining the version number based on information stored in your source code repository (e.g. Git commit version) instead of using information available during the build (build number, date/time/...).

Two main advantages are:
- Everyone building from the same Git commit will get the exact same binaries. If the version numbers come from your build system, your final binaries will be different after every build - even if you rebuild the same commit.
- People doing local builds (e.g. not having access to the official build systems) automatically get unique version numbers.

brendandburns:
@tg123 I don't quite follow the "not from the remote" comment. Does this mean that the versioning code looks at the upstream git repo instead of the local file system?  That doesn't seem correct...

tg123:
@brendanburns @qmfrederik  sorry I checked the NerdBank.GitVersioning and found I am wrong please ignore my comments

brendandburns:
LGTM.

PR #150, opened 2018-04-27
Handle error messages in the watcher

qmfrederik:
The `Watcher` currently always expects `WatchEvent<T>` messages to be sent by the server (e.g. `WatchEvent<V1Pod>`).

That's not always the case, the server may also respond with a `WatchEvent<V1Status>` message in case an error occured. For example, if the resourceVersion set in the watch request is outdated, you may get a `HTTP 200` response with a single Error event - see https://github.com/kubernetes/kubernetes/pull/25369. 

This PR makes the watcher handle that scenario a bit better:
- It introduces a generic `KubernetesObject` class which you can use to parse _any_ JSON object returned by the Kubernetes API server. It exposes the ApiVersion and Kind properties which help you determine what object type is actually being returned.
- It introduces a `KubernetesException` object which wraps an `V1Status` message. That's useful because the Watcher currently uses `Exception` objects to relay errors.
- Updates the `Watcher` class to parse the JSON as a `KubernetesObject`,  and decide whether it's an error message or not, and handle things properly from there.
- Makes the constructor for the `Watcher` public. This helps with unit testing & will also help others when mocking, extending,...

A unit test is included.

### DIFF
--- a/src/KubernetesClient/KubernetesException.cs
+++ b/src/KubernetesClient/KubernetesException.cs
@@ -12,12 +12,19 @@ namespace k8s
     public class KubernetesException : Exception
     {
         /// <summary>
-        /// Initializes a new instance of the <see cref="EtcdException"/> class.
+        /// Initializes a new instance of the <see cref="KubernetesException"/> class.
         /// </summary>
         public KubernetesException()
         {
         }
 
+        /// <summary>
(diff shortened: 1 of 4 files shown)


brendandburns:
LGTM, awesome!

Issue #151, opened 2018-04-27
Add static Kind and ApiVersion properties on generated model classes


qmfrederik:
It would be super useful if generated classes could have a static `Kind` and `ApiVersion` property.

For example, right now I had to do:
```
if(genericObject.Kind == "Status")
{
    var status = SafeJsonConvert.DeserializeObject<V1Status>(json);
}
```

where `Status` is a magic value I had to introduce.

It would be great if I would be able to just do:
```
if(genericObject.Kind == V1Status.Kind)
{
    var status = SafeJsonConvert.DeserializeObject<V1Status>(json);
}
```

making the code a bit more explicit and allowing me to bypass the need for a magic constant.

Is this something the code generator could do?

brendandburns:
Maybe? Swagger doesn't really represent this kind of stuff well, see the discussion here:

https://github.com/OAI/OpenAPI-Specification/issues/1313

We'd have to post-process it on, similar to what you are doing with watch, or add an Enum to the swagger spec.

qmfrederik:
#152 fixed this

Issue #145, opened 2018-04-25
Deleting pods fails because the server returns a Pod object instead of a Status object


qmfrederik:
When you delete a Pod, the server responds with a Pod object instead of a Status object.
The C# client tries to parse the Pod object as a Status object and fails.

This seems to be a general issue with the Swagger/OpenAPI definition of the Kubernetes API, see kubernetes/kubernetes#59501, kubernetes-client/java#86, kubernetes/kubernetes#25716, kubernetes/kubernetes#30537

Opening an issue here mainly to make it discoverable for others.

qmfrederik:
OK - so it took me some digging to find out that was discussed before in #44 and an attempt to mitigate this has been made in #46 & #47.

I don't think the API is not very intuitive but you can do something like this:

```csharp
public static async Task<object> SafeDeleteNamespacedPodAsync(this IKubernetes client, V1DeleteOptions body, string name, string namespaceParameter, int? gracePeriodSeconds = null, bool? orphanDependents = null, string propagationPolicy = null, string pretty = null, CancellationToken cancellationToken = default)
{
   var result = await client.DeleteNamespacedPodAsync(body, name, namespaceParameter, gracePeriodSeconds, orphanDependents, propagationPolicy, pretty, cancellationToken).ConfigureAwait(false);

    if (result.ApiVersion != null || !result.HasObject)
    {
        return result;
    }
    else
    {
        return result.ObjectView<V1Pod>();
    }
}
```

tg123:
the design wants to avoid returning an untyped `object` type.
I don't have any better idea,

tintoy:
This behaviour was one of the issues that led to my suggestion in #101 :)

For what it's worth, here's how I handled it in dotnet-kube-client:

https://github.com/tintoy/dotnet-kube-client/blob/ff2d9bf577663e9c29994b383e67125ea60c13d0/src/KubeClient/ResourceClients/DeploymentClientV1Beta1.cs#L193-L212
https://github.com/tintoy/dotnet-kube-client/blob/ff2d9bf577663e9c29994b383e67125ea60c13d0/src/KubeClient/Models/StatusV1.cs#L11
https://github.com/tintoy/dotnet-kube-client/blob/ff2d9bf577663e9c29994b383e67125ea60c13d0/src/KubeClient/Models/KubeResourceListV1.cs#L13-L14
[...]

Issue #101, opened 2018-03-11
Common base class for Kubernetes resource models


tintoy:
Hi.

I was wondering if it might be useful to have a shared base class for models representing Kubernetes resources / resource-lists (the script I wrote to generate models from Swagger does this, and I imagine autorest can do something similar).

![modelhierarchy](https://user-images.githubusercontent.com/1646284/37257134-feee9456-25b8-11e8-82e4-b4f8498999e9.jpg)

 My existing client library does this and I've found that it makes it easier to deal with resources in a generic fashion:

https://github.com/tintoy/dotnet-kube-client/blob/develop/src/KubeClient/Models/PodV1.cs
https://github.com/tintoy/dotnet-kube-client/blob/develop/src/KubeClient/Models/PodListV1.cs
https://github.com/tintoy/dotnet-kube-client/blob/develop/src/KubeClient/ResourceClients/KubeResourceClient.cs#L90
https://github.com/tintoy/dotnet-kube-client/blob/develop/src/KubeClient/ResourceClients/KubeResourceClient.cs#L122

TBH given that the official client is mostly generated code, I don't know how useful it will be within the client itself, but it's almost certainly going to be useful to some of its consumers.

Again, I get that there are existing users and this might be a breaking change (at the binary-compatibility level at least) but I still think it's something worth considering.

tintoy:
Here's an example of where it's useful:

https://github.com/DimensionDataResearch/glider-gun-v2/blob/master/src/GliderGun.Actors/Events/JobEvents.cs
https://github.com/DimensionDataResearch/glider-gun-v2/blob/master/src/GliderGun.Actors/Events/ResourceEventBus.cs

brendandburns:
Yeah, I see the utility, but I don't want to manually generate the code.

Do you know if AutoRest supports an extension to add a parent class? If so we can look to add that extension...

tintoy:
Agree 😊

The classes are marked as partial, so even if it doesn't I can easily use Roslyn for "post-processing". Will have a think about it this weekend

tintoy:
Sorry, have been a bit flat-out at work, so I haven't had time to follow up on this; I'll try to have a proper look into it this weekend.

tintoy:
[...]

PR #152, opened 2018-04-28
Add a common IKubernetesObject interface, expose type metadata

qmfrederik:
This PR updates the generated model classes so that they:

- Implement a common `IKubernetesObject` interface which contains the `Kind` and `ApiVersion` properties
- Expose `KubeApiVersion`, `KubeKind` and `KubeGroup` constants with metadata about their kind

For example, for `V1Pod`,  the following code is added via a partial class:

```csharp
public partial class V1Pod : IKubernetesObject
{
    public const string KubeApiVersion = "v1";
    public const string KubeKind = "Pod";
    public const string KubeGroup = "";
}
```

This can be useful in a couple of scenarios. For example, APIs which can return a `V1Status` or `V1Pod` object could now expose a `IKubernetesObject` return type. You can then inspect the `KubeKind` property to determine whether it is a `V1Status` or `V1Pod` object and cast accordingly.

Similarly, with this, we could write code which takes any JSON-formatted object and properly serializes it. 

This will conflict with #148 so I'll rebase.

Contributes to #151, #145, #101


brendandburns:
generally LGTM, but needs rebase.

brendandburns:
Awesome! thanks!

PR #154, opened 2018-04-30
Allow callers to specify which WebSocket subprotocol to use

qmfrederik:
Kubernetes implements a couple of different variants of protocol used for communications over WebSockets (for exec, attach, port forward,...).

These are defined in [websocket.go#L36](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/server/remotecommand/websocket.go#L36).

This PR adds a `webSocketSubProtocol` parameter to each of these methods, allowing the caller to be explicit about which sub protocol they want to use.

This is especially useful for callers which want to use v4 of the protocol, which returns the error code as a JSON-formatted `V1Status` object on channel 4 (the error channel) and supports terminal resizing.


src/KubernetesClient/Kubernetes.WebSocket.cs
```diff
@@ -131,7 +131,7 @@ public Task<WebSocket> WebSocketNamespacedPodPortForwardAsync(string name, strin
                 tracingParameters.Add("name", name);
                 tracingParameters.Add("@namespace", @namespace);
                 tracingParameters.Add("ports", ports);
-                tracingParameters.Add("cancellationToken", cancellationToken);
```
brendandburns: Do we really want to remove this?

src/KubernetesClient/Kubernetes.WebSocket.cs
```diff
@@ -63,7 +63,7 @@ public Task<WebSocket> WebSocketNamespacedPodExecAsync(string name, string @name
                 tracingParameters.Add("stdin", stdin);
                 tracingParameters.Add("stdout", stdout);
                 tracingParameters.Add("tty", tty);
-                tracingParameters.Add("cancellationToken", cancellationToken);
```
brendandburns: Do we really want to remove this?

brendandburns:
I guess `cancelationToken` wasn't doing anything on the server, so it is safe to remove, but want to make sure that's intentional.

qmfrederik:
Yup, the `CancellationToken` is purely a .NET client-side thing. However, I noticed all generated methods in `Kubernetes` trace the `CancellationToken`, so let's have the WebSocket-related methods do the same thing, for consistency.

brendandburns:
LGTM.

PR #155, opened 2018-04-30
Handle WebSocket exceptions

qmfrederik:
This is a quick way of handling errors when invoking WebSocket calls. The server returns `400 Bad Request` and a body with details about what went wrong.
Unfortunately, the .NET Core API doesn't give us a way to access the body - see dotnet/corefx#29163.

The workaround, then, is to retry the request as a normal HTTP request and catch the exception as such.

The code will throw a `HttpOperationException` as that is what other methods throw as well.

### DIFF
--- a/src/KubernetesClient/Kubernetes.WebSocket.cs
+++ b/src/KubernetesClient/Kubernetes.WebSocket.cs
@@ -1,8 +1,11 @@
+using k8s.Models;
 using Microsoft.AspNetCore.WebUtilities;
 using Microsoft.Rest;
+using Microsoft.Rest.Serialization;
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Net;
 using System.Net.Http;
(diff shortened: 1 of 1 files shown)


tintoy:
Nice workaround, but just want to check - do all k8s Websocket requests support GET? If not, will the error message be the same (rather than indicating an unsupported verb, which might be confusing for callers).

qmfrederik:
Yes, because the [opening handshake of a WebSocket request](https://tools.ietf.org/html/rfc6455#page-6) is a `HTTP GET`.

We've tested it with e.g. `exec` to be sure 😄 .

tintoy:
I'll keep this in mind then when moving us over to the managed `HttpSocketHandler`, then :)

brendandburns:
(also the Java client uses GET exclusively, since that's all Java websockets supports...)

LGTM.

PR #157, opened 2018-05-07
Upgrade to .NET Core 2.1 RC1

qmfrederik:
.NET Core 2.1 RC1 is being released, so let's try to use that instead of preview2.


brendandburns:
LGTM, thanks.

PR #164, opened 2018-05-29
Fix issue where the watcher URLs would be prefixed with a '/'. 

qmfrederik:
This would lead to URLs such as `https://localhost:8443//api/...`, and the double `//` after the port number would case a `404 NOT FOUND` response.


brendandburns:
LGTM, thanks!

PR #166, opened 2018-05-31
Update to .NET Core 2.1 RTM

qmfrederik:
.NET Core 2.1 has RTM'ed, so upgrade to the final version.


brendandburns:
LGTM, thanks!

PR #167, opened 2018-05-31
Fix an issue where growing the ByteBuffer would fail

qmfrederik:
There was a bug in the `ByteBuffer` class where growing the buffer would fail if the read water mark was 0.

Having the read water mark at 0 means that there is no data available to read, but this was incorrectly treated as "all data in the buffer is available for reading".

I added a couple of extra unit tests for this behavior.

### DIFF
--- a/src/KubernetesClient/ByteBuffer.cs
+++ b/src/KubernetesClient/ByteBuffer.cs
@@ -287,22 +287,19 @@ namespace k8s
 
             var newBuffer = ArrayPool<byte>.Shared.Rent(size);
 
-            if (this.WriteWaterMark <= this.ReadWaterMark)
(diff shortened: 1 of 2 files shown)


qmfrederik:
The Travis buid is broken because apparently the installation of .NET Core 2.1 RC is broken:

```
The following packages have unmet dependencies:
 dotnet-sdk-2.1.300-rc1-008673 : Depends: aspnetcore-runtime-2.1.0-rc1-final but it is not going to be installed
                                 Depends: dotnet-runtime-2.1.0-rc1 but it is not going to be installed
```

brendandburns:
LGTM.

qmfrederik:
Rebased on top of master to fix the build failure.

PR #169, opened 2018-05-31
Bump version to 1.1

qmfrederik:
Current builds from master will be marked as 1.0.0-something, which NuGet considers to be prerelease versions of 1.0.0 (and hence older than the currently released 1.0.0 version).

This PR changes the version number so that new packages will be versioned `1.1-something`.

NuGet will then detect them as:
- Newer than 1.0.0
- Prerelease versions

### DIFF
--- a/version.json
+++ b/version.json
@@ -1,6 +1,6 @@
 {
   "$schema": "https://raw.githubusercontent.com/AArnott/Nerdbank.GitVersioning/master/src/NerdBank.GitVersioning/version.schema.json",
-  "version": "1.0.0",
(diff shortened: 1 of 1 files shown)


brendandburns:
LGTM, thanks. We should really script/automate the cutting of a release to do things like this.

qmfrederik:
Rebased on top of master to fix the build failure.

tintoy:
> We should really script/automate the cutting of a release to do things like this.

I've found GitVersion to be pretty good for doing semver-from-commit-history :)

Issue #130, opened 2018-03-30
Consider using SocketsHttpHandler


qmfrederik:
See #129 for details, but custom/client certificate handling on macOS seems to be broken unless using the `SocketsHttpHandler`.

In .NET Core 2.1 Preview 1, you can enable this using an environment variable. In Preview 2, you should be able to control this in code by running:

```
new HttpClient(new SocketsHttpHandler())
``` 

This issue is to track us making that change when .NET Core 2.1 Preview 2 ships.

tintoy:
Yep same goes for our WebSockets connection builder :)

tintoy:
Now that the preview2 SDK is available, I'll open a PR to switch the Travis build over to using it :)

tintoy:
If we want to start cleaning up the hacky code for the WebSocket client, we need the new `SocketsHttpHandler` from preview 2, _but_...

@brendanburns - the .NET Core packages for Debian / Ubuntu are [broken](https://travis-ci.org/tintoy/dotnet-kube-client/builds/367782907#L1793) for preview 2.

I got this going in my other project by installing from tarball instead (https://github.com/tintoy/dotnet-kube-client/commit/66c27cf86b875229e890f4d9ca275adb91f90e87), are you OK with me taking the same approach here? We could always wait for new packages but it looks like they've been broken for a while and may not be fixed until after preview 2.

brendandburns:
@tintoy I assume that the .NET core packages are fixed now that 2.1 final has released. Are we ready to start moving?

tintoy:
Yep! Will have a go at this tomorrow morning (it's Saturday afternoon here ATM)

tintoy:
@brendandburns - sorry have been flat out at work (new job) but will have time to do this first thing tomorrow :)

brendandburns:
No worries, want help? I'm happy to pitch in some code and split the work if you want...

--brendan

________________________________

tintoy:
Thanks! I think I'm OK for now - it's not a large piece of work (less than an hour), just haven't had time to do it yet. If I run into any difficulties tomorrow though I'll be putting my hand up :)

brendandburns:
SGTM thanks

--brendan

________________________________

tintoy:
Ok - working on this now.

PR #171, opened 2018-06-06
Remove hacky WebSocket connection factory

tintoy:
Replace it with the new (officially-supported) ClientWebSocket
SSL functionality in .NET Core 2.1.

Implements kubernetes-client/csharp#130.

CC: @brendandburns @qmfrederik


tintoy:
Thanks to last-minute changes in CoreFX for 2.1, we no longer need to manually set up the connection at all; we can simply use `ClientWebSocket` on `netcoreapp2.1` and it just does the right thing on each platform :)

brendandburns:
LGTM! (thanks)

Issue #173, opened 2018-06-11
X509VerificationFlags.AllowUnknownCertificateAuthority unexpected behavior.


davidorbelian:
As said [here](https://stackoverflow.com/questions/6097671/how-to-verify-x509-cert-without-importing-root-cert):

> It was also discovered through experimentation that setting
> `ChainPolicy.VerificationFlags = X509VerificationFlags.AllowUnknownCertificateAuthority;`
> will allow the build method to return true even if you don't add certificates to the ExtraStore which completely defeats the purpose of checking. I do not recommend using this flag for any reason.

I can confirm that I tried this and got same result.

Steps to reproduce: 
Open _kubeconfig_ and replace _certificate-authority-data_ with any other certificate and the result is still true when building the chain. 

The solution from StackOverflow offers to replace 
```csharp
chain.ChainPolicy.VerificationFlags = X509VerificationFlags.AllowUnknownCertificateAuthority;
var isValid = chain.Build((X509Certificate2) certificate);
return isValid;
```
with
```csharp
chain.Build((X509Certificate2) certificate);
var isValid = chain.ChainStatus.Length == 1 && chain.ChainStatus.First().Status == X509ChainStatusFlags.UntrustedRoot;
return isValid;
```

davidorbelian:
I opened an [issue](https://github.com/dotnet/corefx/issues/30284) on [dotnet/corefx](https://github.com/dotnet/corefx) to clarify this.

davidorbelian:
They replied as follows:

> If AllowUnknownCertificateAuthority is the only flag set then chain.Build() will return true if
> 
> - The chain correctly terminated in a self-signed certificate (via ExtraStore, or searched persisted stores)
> - None of the certificates are invalid per the requested revocation policy
> - All of the certificates are valid under the (optional) ApplicationPolicy or CertificatePolicy values
> - All of the certificates' NotBefore values are at-or-before VerificationTime and all of the certificates' NotAfter values are (at-or-)after VerificationTime.
>
> If that flag is not specified then an additional constraint is added:
> - The self-signed certificate must be registered as trusted on the system (e.g. in the LM\Root store).
>
[...]

PR #174, opened 2018-06-12
Fix issue with X509VerificationFlags.AllowUnknownCertificateAuthority behavior

davidorbelian:
Fix [this issue](https://github.com/kubernetes-client/csharp/issues/173) with [this solution](https://github.com/dotnet/corefx/issues/30284).

How to test:
Replace your _certificate-authority-data_ from _kubeconfig_ with data from another certificate and run. 
- BEFORE:
The program works despite the non-validity of the certificate because of validation does not fail.
- AFTER:
**System.Net.Http.HttpRequestException** is thrown because of validation fails.


src/KubernetesClient/Kubernetes.ConfigInit.cs
```diff
@@ -179,8 +180,11 @@ private void SetCredentials(KubernetesClientConfiguration config, HttpClientHand
 
                 // add all your extra certificate chain
                 chain.ChainPolicy.ExtraStore.Add(caCert);
-                chain.ChainPolicy.VerificationFlags = X509VerificationFlags.AllowUnknownCertificateAuthority;
-                var isValid = chain.Build((X509Certificate2) certificate);
+                chain.Build((X509Certificate2) certificate);
+
+                var rootCert = chain.ChainElements[chain.ChainElements.Count - 1].Certificate;
+                var isValid = rootCert.RawData.SequenceEqual(caCert.RawData);
```
bartonjs: If you want to test for a custom root you'd leave the check as it was, and then add  `isValid = isValid && rootCert.RawData.SequenceEqual(caCert.RawData);`

As it is, this change would start accepting expired and revoked certificates (and possibly certificates where the signatures didn't check out).

davidorbelian: Thank you!

davidorbelian:
@bartonjs can you please confirm now it is ok?

davidorbelian:
@bartonjs Sure, thank you!
@brendandburns What you think?

davidorbelian:
@brendandburns Done!

brendandburns:
LGTM, many thanks!

Issue #165, opened 2018-05-29
Watchers time out after 100 seconds


qmfrederik:
- The default timeout for the `HttpClient` is 100 seconds, so the watchers will timeout after this amount of time.
- There is no default API to override the timeout on a per-request basis
- https://www.thomaslevesque.com/2018/02/25/better-timeout-handling-with-httpclient/ describes an alternative but it implies setting the timeout to Infinite for all requests and then specifying the timeout on a per-request basis for all requests.

qmfrederik:
I'm not entirely sure, but this appears to be an issue when the server does not send any content at all in the first 100 seconds after the GET request is issued. When the server does send a first response, and subsequent updates come in intervals greater than 100 seconds, the code appears to work fine.

brendandburns:
@qmfrederik is there an action we can take here? Or is this to document the issue for people who encounter it?

PR #178, opened 2018-06-22
Extra tests for the Watcher classes

qmfrederik:
This PR adds two new tests
- `WatchEventsWithTimeout`, which mimics `WatchEvents` but adds a delay of 120 seconds between 2 watch events. The default timeout of the HttpClient is 100 seconds, and this test servers to make sure the HttpClient doesn't choke on the 2 minutes without any events
- `DirectWatchAllEvents` which is similar to `WatchAllEvents`, but directly uses the `Kubernetes.Watch*` API
- `DirectWatchAllEventsWithTimeout` which mimics a scenario in which you initiate a watch, but no events are sent in the first 100 seconds. In this scenario, the client times out. The test is disabled because this bug has not been fixed (see #165).


brendandburns:
LGTM, thank you!

PR #184, opened 2018-06-27
Watcher: notify the caller when the server closes the connection

### LINKED ISSUE #181: Watchers should send a notification when the connection is closed
The Kubernetes watch APIs have a timeout, after which the connection is closed.

Each watch requests gets assigned a random timeout between [timeout, 2 * timeout]. The default timeout is 30 minutes, see https://github.com/kubernetes/kubernetes/pull/15705.

The `Watcher` updates the `Watching` property but doesn't raise an event/invoke a callback when the server terminates the request.

Suggestion: make the Watcher APIs take an `onClosed` callback, and have the Watcher class expose an Closed event.

Background/rationale: I had naïve code like this which was using a watcher to block until a job runs to completion, but it would just idle forever when the connection is closed by the server:

```csharp
using (var watcher = await kubernetes.WatchObjectAsync<V1Job>(
    path: $"apis/batch/v1/watch/namespaces/{job.Metadata.NamespaceProperty}/jobs/{job.Metadata.Name}",
    resourceVersion: resourceVersion,
    onEvent: (eventType, j) =>
    {
        logger.LogInformation($"Got a {eventType} event for job {j.Metadata.Name}, resource version {j.Metadata.ResourceVersion}. The job completion time is set to {j.Status?.CompletionTime}");
        resourceVersion = j.Metadata.ResourceVersion;

        if (eventType == WatchEventType.Deleted)
        {
            logger.LogInformation($"The job was deleted.");
            jobCompleted = true;
            mre.Set();
        }
        else if (j.Status.HasCompleted())
        {
            logger.LogInformation($"The job has completed.");
            jobCompleted = true;
            mre.Set();
        }
        else if (j.Status.HasFailed())
        {
            logger.LogInformation($"The job has failed. {j.Status.GetFailureMessage()}.");
            jobCompleted = true;
            mre.Set();
        }
    },
    onError: (ex) =>
    {
        logger.LogInformation($"An error occurred within the watcher while polling. {ex.Message}");

        // Signal the calling thread so it will try again.
        mre.Set();
    },
    cancellationToken: cancellationToken).ConfigureAwait(false))
{
    await mre.WaitAsync(cancellationToken).ConfigureAwait(false);
}
```

qmfrederik:
The Kubernetes server can close the connection on any watch call. There's a default timeout of 30-60 minutes, though you can specify a custom timeout when you use the `timeoutSeconds` parameter.

Let the consumer of the `Watcher<T>` class know when the server closed the connection, by raising the `Closed` event or invoking the `onClosed` callback.

Also regenerates the API and updates the unit tests, so this may best be reviewed on a per-commit basis.

Fixes #181, #165

### DIFF
--- a/gen/KubernetesWatchGenerator/IKubernetes.Watch.cs.template
+++ b/gen/KubernetesWatchGenerator/IKubernetes.Watch.cs.template
@@ -35,6 +35,9 @@ namespace k8s
         /// <param name="onError">
         /// The action to invoke when an error occurs.
         /// </param>
+        /// <param name="onClosed">
+        /// The action to invoke when the server closes the connection.
+        /// </param>
         /// <param name="cancellationToken">
         /// A <see cref="CancellationToken"/> which can be used to cancel the asynchronous operation.
         /// </param>
@@ -55,6 +58,7 @@ namespace k8s
             Dictionary<string, List<string>> customHeaders = null,
             Action<WatchEventType, {{GetClassName operation}}> onEvent = null,
             Action<Exception> onError = null,
+            Action onClosed = null,
             CancellationToken cancellationToken = default(CancellationToken));
 
         {{/.}}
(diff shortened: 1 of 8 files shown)


qmfrederik:
Fixes #181, #165

brendandburns:
Looks reasonable but the unit tests are hanging on Travis for some reason...

qmfrederik:
Yeah, should be fixed now. The start of `WatcherLoop` was running synchronously, which caused a deadlock. I added `Task.Yield` to force the method to run async from the start.

brendandburns:
LGTM.

Issue #183, opened 2018-06-27
List namespaced services missing first element


ahaeber:
When using ListNamespacedServiceWithHttpMessagesAsync and watch I seem to not get the first element of the list.

I tried to follow the example at https://github.com/kubernetes-client/csharp/blob/master/examples/watch/Program.cs to watch list of services in a namespace. At first I always found one service missing in the list of services I got in the watch-response handler. Then I enabled Microsoft.Rest.ServiceClientTracing and found that the missing service is shown in the body of the response from Kubernetes. It does not show traces for the rest of the services, which is a bit weird but I guess that is how watches work.

I also tried simply running the watch example to list pods to see whether there was any difference, but it has the same behavior. In a namespace for Nginx ingress kubectl shows two pods:

>$ kubectl get pods -n ingress-nginx
>NAME                                        READY     STATUS    RESTARTS   AGE
>default-http-backend-170366572-2tc9b        1/1       Running   0          36d
>nginx-ingress-controller-3129142458-0x4k6   1/1       Running   0          36d

With the example Program it only shows one "Added" event:
>$ dotnet run -- "ingress-nginx"
>press ctrl + c to stop watching
>Added: name=nginx-ingress-controller-3129142458-0x4k6,labels=[app=ingress-nginx,pod-template-hash=3129142458]

(missing the pod `default-http-backend-170366572-2tc9b` here)
This is running in an AKS cluster with Kubernetes version 1.7.7.

I looked at the tests (https://github.com/kubernetes-client/csharp/blob/master/tests/KubernetesClient.Tests/WatchTests.cs) and found a few comments that says 
>first line of response is eaten by WatcherDelegatingHandler

[L87](https://github.com/kubernetes-client/csharp/blob/master/tests/KubernetesClient.Tests/WatchTests.cs#L87),  [L223](https://github.com/kubernetes-client/csharp/blob/master/tests/KubernetesClient.Tests/WatchTests.cs#L223), [L290]( https://github.com/kubernetes-client/csharp/blob/master/tests/KubernetesClient.Tests/WatchTests.cs#L290)
Could it be that the WatcherDelegatingHandler should not eat anything here and pass it to the watch-listener instead?

qmfrederik:
Do you get the complete list when you call `ListNamespacedServiceWithHttpMessagesAsync` with `watch` set to `false`?

Do you get all events when you call `WatchNamespacedServiceAsync`?

brendandburns:
[...]

PR #190, opened 2018-07-03
Update WatcherDelegatingHandler to stop eating first line (#183)

hossambarakat:


src/KubernetesClient/WatcherDelegatingHandler.cs
```diff
@@ -64,6 +66,37 @@ protected override bool TryComputeLength(out long length)
                 length = 0;
                 return false;
             }
+        }
+        internal class PeekableStreamReader : StreamReader
```
brendandburns: I'm worried that if someone calls any of the other `ReadBlock`, `Read` etc methods this is going to break...

At the very least let's override those messages and have them throw an exception so that we detect that it's happening...

hossambarakat: Makes sense to me. I have updated the pull request to override all the remaining `Read` methods to throw `NotImplementedException`

hossambarakat: @brendanburns The build is failing but I have no clue why and I don't have permission to re-trigger the build. Would you please point me to what should I check?

hossambarakat: @brendanburns, Was wondering if you have time to review the updated request

brendandburns:
LGTM, thanks!

PR #193, opened 2018-07-14
Add support for 'azure' AAD based authentication.

brendandburns:


tg123:
workaround at current version

```
   // Learn more at https://docs.microsoft.com/en-us/azure/api-management/api-management-howto-aad
            var aksuri = "https://.........................azmk8s.io";

            var clientId = "............................";
            var clientSecret = ".........................";
            var tenantId = ".............................";

            IKubernetes client = new Kubernetes(new AzureCredentials(new ServicePrincipalLoginInformation
            {
                ClientId = clientId,
                ClientSecret = clientSecret
            }, tenantId, AzureEnvironment.AzureGlobalCloud))
            {
                BaseUri = new Uri(aksuri)
            };


```

brendandburns:
@tg123 Can you clarify your comment? Do you think we should pursue that approach instead of this PR?

Thanks!

tg123:
1 auth provider should be something which can generate token automatically, like the code I post above.
2 my code need some change to Kubernetes.cs in order to be more elegant

brendandburns:
@tg123 thanks for the comment. I'm going to merge this code as is and we can iterate from there? This will work with existing `kubeconfig` files so I think that's an important first step...

Let me know if you have any concerns.

Thanks

tg123:
OK 👍

Issue #160, opened 2018-05-23
Fractions library isn't strongly named..


brendandburns:
Running examples I see:

```
CSC : warning CS8002: Referenced assembly 'Fractions, Version=3.0.1.0, Culture=neutral, PublicKeyToken=null' does not have a strong name. [/home/brendan/src/kubernetes-client/csharp/src/KubernetesClient/KubernetesClient.csproj]
```

@qmfrederik is there a way to fix this?

qmfrederik:
Is this on the .NET Framework or on .NET Core?

lamchester:
.NET Framework

lamchester:
(poke) any update on this?

tg123:
https://github.com/danm-de/Fractions

lamchester:
Yes, that's the fractions library. The problem is, it's not signed. With .NET Framework, assemblies that are strong name signed can't load assemblies that aren't.

Can you all either:
 - Don't strong name sign the KubernetesClient assembly in the published nuget package. Then we can get the .NET Framework to not enforce strong names
 - Depend on a version of the Fractions library that's strong name signed. I don't know if one exists.
 - Don't use the Fractions library

Obviously the latter two options would be preferable, in case KubernetesClient needs to be referenced from a strong name signed assembly.

brendandburns:
@lamchester I think the right thing to do here is to import the relevant Fractions.cs files into this repository and point the code at the local copy instead. The licenses are compatible, I'd be happy to review the PR.

Thanks

RohanNagar:
This can be closed now!

brendandburns:
closing! :)

ArchangelSDY:
Hi, do you have any timeline to publish a new NuGet package with this fix? Currently we are using v1.2.0 and are blocked by this issue.

brobichaud:
> Hi, do you have any timeline to publish a new NuGet package with this fix? Currently we are using v1.2.0 and are blocked by this issue.

Yeah I too am dead-ended on trying to use this k8s client.

qmfrederik:
Have you tried any of the more recent versions, like 1.4.1-g1600e44b00?
PS - Not sure why the git version is appended to builds coming from master, that's probably something we'll need to fix

brobichaud:
> Have you tried any of the more recent versions, like 1.4.1-g1600e44b00?
> PS - Not sure why the git version is appended to builds coming from master, that's probably something we'll need to fix

[...]

PR #194, opened 2018-07-16
Move Fractions library inside project

RohanNagar:
Addresses #160.

All of the `Fractions` code is taken as-is on their [master branch](https://github.com/danm-de/Fractions), with the only difference being instead of using generated error messages (i.e. [their `Resources` class](https://github.com/danm-de/Fractions/blob/master/src/Fractions/Properties/Resources.Designer.cs)), I just use the string error messages directly. These are located at:

```
Fractions/Formatter/DefaultFractionFormatter.cs : 16
```
[Link to Changed Line](https://github.com/kubernetes-client/csharp/pull/194/files#diff-9cd265efeb043f6105753d8fdeed1632R16)
[Original Code Before Change](https://github.com/danm-de/Fractions/blob/master/src/Fractions/Formatter/DefaultFractionFormatter.cs#L17)

```
Fractions/Fraction.CompareTo.cs : 23
```
[Link to Changed Line](https://github.com/kubernetes-client/csharp/pull/194/files#diff-3b47fab69319a9c27e26d81b5049f391R23)
[Original Code Before Change](https://github.com/danm-de/Fractions/blob/master/src/Fractions/Fraction.CompareTo.cs#L24)

```
Fractions/Fraction.Math.cs : 113
```
[Link to Changed Line](https://github.com/kubernetes-client/csharp/pull/194/files#diff-aedbc6ac4ec7f519867c21c814914530R113)
[Original Code Before Change](https://github.com/danm-de/Fractions/blob/master/src/Fractions/Fraction.Math.cs#L114)

```
Fractions/Fraction.ConvertFrom.cs : 43
```
[Link to Changed Line](https://github.com/kubernetes-client/csharp/pull/194/files#diff-d2a3d00750aceb0612471fde0f43065dR43)
[Original Code Before Change](https://github.com/danm-de/Fractions/blob/master/src/Fractions/Fraction.ConvertFrom.cs#L44)

Those 4 lines (and the removed `import` at the top of each file) are the only difference from the original code.

### DIFF
--- a/src/KubernetesClient/Fractions/Extensions/MathExt.cs
+++ b/src/KubernetesClient/Fractions/Extensions/MathExt.cs
@@ -1,0 +1,128 @@
+﻿using System;
+
+namespace Fractions.Extensions {
+    internal static class MathExt {
+        /// <summary>
+        /// Checks for an even number.
+        /// </summary>
+        /// <param name="number"></param>
+        /// <returns><c>true</c> if the number is even.</returns>
+        public static bool IsEven(this long number) {
+            return (number & 1) == 0;
+        }
+
+        /// <summary>
+        /// Checks for an odd number.
+        /// </summary>
+        /// <param name="number"></param>
+        /// <returns><c>true</c> if the number is odd.</returns>
+        public static bool IsOdd(this long number) {
+            return (number & 1) != 0;
+        }
+
+        /// <summary>
+        /// Get the greatest common divisor (GCD) of <paramref name="a"/> and <paramref name="b"/>.
+        /// </summary>
+        /// <param name="a">First number.</param>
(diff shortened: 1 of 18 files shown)


brendandburns:
This looks fine, but please add a README.md which includes a pointer to the original location for the source as well as the reason for the vendoring.

Also, please include the LICENSE.txt file from the original repo in that directory.

Thanks!

RohanNagar:
Thanks @brendanburns. Added the `README` and `LICENSE`.

brendandburns:
LGTM.

Thanks.

danm-de:
Hi, 
Fractions 4.0.1 should be strong named.

Issue #37, opened 2017-10-21
Code style checker before checking in


tg123:
Sadly, C# did not have a `go fmt` yet.
A check tool before checking in is wanted.

the tool should include, but not limit to, those features:
 * naming convention
 * line coding (crlf / lf)
 * utf8 bom 
 * run on all platform (linux, mac, windows, ...)

some candidates:
https://github.com/dotnet/codeformatter
https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md

sesispla:
Maybe [StyleCop.Analyzers](https://github.com/DotNetAnalyzers/StyleCopAnalyzers) is what we are looking for. Not sure if it can work with dotnet/cli, but worth to take a look.

StyleCop is the old guy helping us keep code uniform between developers, in the full Visual Studio/Windows world :)

Some [config examples](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/Configuration.md)

tg123:
I will have a look too.

I am porting some tools to .net core

https://github.com/tg123/codeformatter
https://github.com/tg123/editorconfig-core-net

and working on a .net core version of eclint to combine those tools together.

rahulrai-in:
Is this issue still open? I can see that .editorconfig file is present in the code. Is there anything else that needs to be done?

tg123:
please see #187 
Will/May apply resharper code style checker

rahulrai-in:
Why do we need to apply Resharper code style check? Would we like do away with .editorconfig?

tg123:
`.editorconfig` is a place to store policy
a tool (resharper or something else) is need to enforce policies wthin `.editorconfig` are applied to all files

those policies might be a battle between

```
func {
}
```

and 
```
func
{
}
```

unluckily, there is no `go fmt` for dotnet at the moment.

rahulrai-in:
Just wanted to make sure that I understand this correctly. Do we want to execute R# using the editorconfig on the solution then?

tg123:
@rahulrai-in sorry for the unclear
Resharper would be one choice for code style checking tools
but it does not support `.editorconfig` as I know

I just want to use `Resharper` to act as code style checking tools via VSTS and the polices might be defined somewhere else.

PR #203, opened 2018-09-21
align files to .editorconfig

tg123:
eclint fix **/*.cs

check will integrate with CI process
 

#37

@gitfool


gitfool:
@tg123 I didn't know about [eclint](https://github.com/jedmao/eclint). Nice!

I see all the line endings are still `CRLF` though, despite the `.editorconfig` specifying `LF`. When editing an existing file it seems the existing EOL is preserved. However, when creating a new file (in VS 2017 at least) it will use `LF`, so either `.editorconfig` should be changed to specify `CRLF` or the EOL should be changed to `LF` for all existing files.

tg123:
@gitfool 
only eclint with *.cs
and generated codes are ignored

gitfool:
I ran the same command as you on your branch, `eclint fix **/*.cs`, and all cs files were further updated, as expected, to use `LF` for EOL. So you must've passed `CRLF` on the command-line to eclint; but this has the problem I mention above.

tg123:
@gitfool I checked all .cs are LF except `/src/KubernetesClient/generated/`
for generated they should be handled by autorest.

my command
```
find . -name *.cs -exec file {} \; | grep CRLF | grep -v generated
```

gitfool:
Right. If I clone your repo in WSL then I get `LF`. If I clone in Windows then I get `CRLF`. The problem is that because there's no `.gitattributes` file, it depends on everyone's git client settings. If I edit an existing file on Windows it will be committed using `CRLF`.

The other approach is to remove the `end_of_line` entry from `.editorconfig` and add a `.gitattributes` file. See [dealing-with-line-endings](https://help.github.com/articles/dealing-with-line-endings/#platform-all) and the [gitattributes](https://github.com/alexkaratarakis/gitattributes) repo for some relevant templates. If specified correctly then text files will match the os platform when editing but always be stored in git using `LF`.

tg123:
@gitfool you may be a victim of `autocrlf`

gitfool:
Haha. I've been the victim of many EOL abuses. The point is _everyone_ has their own settings and you can't expect them to comply with your settings, so the best way to address this is with a `.gitattributes` file in the repo. This and `.gitignore` should be the first files added to a new repo. When added late you have to  [renormalize](https://stackoverflow.com/questions/7156694/git-how-to-renormalize-line-endings-in-all-files-in-all-revisions/47580886#47580886) line endings which can make a mess of the repo history.

gitfool:
That's all very well, and is one way to lock things down, but again you've missed the point. You're only considering what you use. A `.gitattributes` file is particularly important for cross-platform projects where developers will use different tools on different operating systems. I give up.

tg123:
@gitfool thanks I see, that is a good idea.
You can create a PR for that

gitfool:
Cool. Will do after this is PR merged, since they'll overlap.

brendandburns:
LGTM.

PR #205, opened 2018-09-28
run generate script for k8s 1.12

tg123:


brendandburns:
Let's rev the major version of this library to reflect the new version of k8s?

Otherwise, LGTM.

tg123:
1.2.12 or something?

itowlson:
@tg123 It looks previous versions of the library haven't been coupled to the k8s version, which makes sense for when the library needs to be revved independently.  Does this result in any breaking changes?  If not, maybe 1.3.0?

brendandburns:
Yeah, I agree with @itowlson `1.3.0` or `2.0.0`

For the java library we adopted a "new major version for each k8s release" approach, we haven't done that in csharp so far, but maybe we should start (in which case `2.0.0` would be the right version)

tg123:
bumped to 1.3

PR #209, opened 2018-10-14
Fix up YAML serialisation to understand AutoRest property renaming

### LINKED ISSUE #208: `Yaml.LoadFromString` (and others) are broken due to `namespace` keyword.
`V1ObjectMetadata` defines the `namespace` property as follows.

```
[JsonProperty(PropertyName = "namespace")]
public string NamespaceProperty { get; set; }
```

This is likely because `namespace` is a reserved keyword in C#.  However, YamlDotNet does not respect the `JsonProperty` attribute, so these deserializers fail when they come across `namespace` in the Yaml.

The best fix would be to either use an attribute that YamlDotNet understands, or to rename this property to `Namespace`, since C# doesn't care if the `N` is capitalized.  Lastly, to avoid confusion, this property to could be marked with the `@`.

itowlson:
AutoRest renames properties such as 'namespace' and 'readOnly' to avoid clashes with keywords in common .NET languages.  It uses `JsonPropertyAttribute` to map these back to the right names in JSON.  However, the YAML serialiser isn't aware of the JSON mappings, so loading YAML including such properties fails (similarly, writing them out produces the wrong YAML text).  This PR adds a custom type inspector to the YAML serialisation stack which performs the additional mapping needed.

Fixes #208.

### DIFF
--- a/src/KubernetesClient/Yaml.cs
+++ b/src/KubernetesClient/Yaml.cs
@@ -1,4 +1,8 @@
+using Newtonsoft.Json;
+using System;
+using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Text;
 using System.Threading.Tasks;
 using YamlDotNet.Core;
@@ -29,6 +33,7 @@ namespace k8s
             var deserializer =
                 new DeserializerBuilder()
                 .WithNamingConvention(new CamelCaseNamingConvention())
+                .WithTypeInspector(ti => new AutoRestTypeInspector(ti))
                 .Build();
             var obj = deserializer.Deserialize<T>(content);
             return obj;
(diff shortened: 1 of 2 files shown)


itowlson:
Don't merge this yet - we might be able to bypass this runtime fixup by changing our AutoRest config to emit YamlMemberAttribute.  I'll look into this.

tg123:
it is tough to modify autorest generator, I think.

itowlson:
Let's give up on the AutoRest config approach for now, and go with this.  Please review @brendandburns @tg123 - thanks!

CI failure looks like the known certificate test flake.

brendandburns:
/lgtm
/approve

@itowlson can we get a rebase? I fixed (well, commented out) that test :)

twitchax:
> @tg123 Yeah... I have been looking through the generation scripts and I can see where we do some `sed`-based post-processing but can't see any AutoRest config that would be easy to tweak. We _might_ be able to add another `sed` step that duplicates every
> 
> `[JsonProperty(PropertyName = "...")]`
> 
> into
> 
> `[JsonProperty(PropertyName = "..."), YamlMember(Alias = "...")]`
> 
> The relevant post-processing is around https://github.com/kubernetes-client/gen/blob/087e4b4d9b4e40973a24f848b47f0dacc08e9160/openapi/csharp.sh#L62. My sed-fu is rusty but I think it bears investigating...

@fearthecowboy may be interestedi n this feedback.

PR #211, opened 2018-10-18
Rev YamlDotNet library.

brendandburns:
@tg123


tg123:
LGTM

philips:
Build is failing. Can someone please look at this?

itowlson:
Is this #214 or is there another failure?  The proposal for 214 is to remove Core 2.1 from the test targets, though that test runs only on 2.1 so it might be better just to ignore it.

PR #212, opened 2018-10-19
fix a race condition.

xinyanmsft:
when multiple call to GetStream happens around the same time, on the
same inputIndex, a race condition will cause this.buffers.Add() to throw
exception.


brendandburns:
@xinyanmsft Thanks for the PR, for some reason the CLA signing doesn't seem to have worked, can you retry? Do you have your email address associated with your github account? That's a common problem...

There's a failing unit test, but please don't worry about that, there's something broken in the underlying CI/CD system, because that test is failing in a different PR too, I need to figure out what's going wrong...

xinyanmsft:
retry signing CLA

PR #170, opened 2018-06-06
Create KubernetesClientConfiguration from pre-loaded K8SConfiguration

markrendle:
Hi. I'm working on a GUI application that loads the `K8SConfiguration`, then creates `KubernetesClientConfiguration` instances based on user selection. Since I've already got the config data loaded, it makes sense to be able to create the client config directly from it, so I've added a new static method `BuildConfig` that takes the `K8SConfiguration` instance as a parameter. I guess I could have made the existing `GetKubernetesClientConfiguration` method public, but I don't like messing with existing API.

Oh, I also fixed up a doc comment that was wrong :)

PR #213, opened 2018-10-22
Create KubernetesClientConfiguration from pre-loaded K8SConfiguration

itowlson:
Recreates #170 by @markrendle (with Mark's consent, as he does not have bandwidth to update it).  From the original description:

"Hi. I'm working on a GUI application that loads the K8SConfiguration, then creates KubernetesClientConfiguration instances based on user selection. Since I've already got the config data loaded, it makes sense to be able to create the client config directly from it, so I've added a new static method BuildConfig that takes the K8SConfiguration instance as a parameter. I guess I could have made the existing GetKubernetesClientConfiguration method public, but I don't like messing with existing API.

Oh, I also fixed up a doc comment that was wrong :)"

And I tweaked a couple more doc comments too.

### DIFF
--- a/src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
+++ b/src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
@@ -39,6 +39,7 @@ namespace k8s
         }
 
         /// <summary>
+        /// Initializes a new instance of the <see cref="KubernetesClientConfiguration" /> from config file
         /// </summary>
         /// <param name="kubeconfig">Fileinfo of the kubeconfig, cannot be null</param>
         /// <param name="currentContext">override the context in config file, set null if do not want to override</param>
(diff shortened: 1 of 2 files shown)


src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
```diff
@@ -85,6 +87,15 @@ public partial class KubernetesClientConfiguration
             return k8SConfiguration;
         }
 
+        /// <summary>
+        /// Initializes a new instance of <see cref="KubernetesClientConfiguration"/> from pre-loaded config object.
+        /// </summary>
+        /// <param name="k8sConfig">A <see cref="K8SConfiguration"/>, for example loaded from <see cref="LoadKubeConfigAsync(string, bool)" /></param>
+        /// <param name="currentContext">Override the current context in config, set null if do not want to override</param>
+        /// <param name="masterUrl">Override the Kubernetes API server endpoint, set null if do not want to override</param>
+        public static KubernetesClientConfiguration BuildConfig(K8SConfiguration k8SConfig, string currentContext = null, string masterUrl = null)
```
tg123: just naming to make this easier to understand.

```suggestion
        public static KubernetesClientConfiguration BuildConfigFromCurrent(K8SConfiguration k8SConfig, string currentContext = null, string masterUrl = null)
```

itowlson: Good plan.  I'm going to propose `BuildConfigFromConfigObject` to align with `...FromConfigFile` and because 'current' doesn't quite describe it.  But agree it needs additional qualification - thanks!

PR #231, opened 2019-01-24
add WebSocket server certificate validation for net452

xinyanmsft:
For WebSocket, the server certificate validation does not work with net452. This is due to WebSocketOptions.RemoteCertificateValidationCallback not available to net452. 

This change hooks up the global System.Net.ServicePointManager.ServerCertificateValidationCallback when on net452 to handle server certificate validation.


tg123:
/LGTM

seems have to use `global ServerCertificateValidationCallback`
the nuget package is only targeting `netstandard1.4;netcoreapp2.1`

I am working azure pipeline to add `net452` target

PR #233, opened 2019-01-28
bug fix - remove port bytes during port forwarding

xinyanmsft:
In StreamDemuxer, if the buffer is created before connection is established, the port bytes are not removed when the bytes are delivering to the client. Repro code looks like (the key is to call remoteStreams.Start AFTER GetStream):

  var ws = await kubernetesClient.WebSocketNamespacedPodPortForwardAsync (... )
  var remoteStreams = new StreamDemuxer(ws);
  var stream = remoteStreams.GetStream(0, 0);
  remoteStreams.Start();

This change filters out the port bytes which are the 2nd and 3rd bytes sent in all cases.


src/KubernetesClient/StreamDemuxer.cs
```diff
@@ -199,6 +199,13 @@ protected async Task RunLoop(CancellationToken cancellationToken)
 
                     var streamIndex = buffer[0];
                     var extraByteCount = 1;
+                    byteCount += result.Count - extraByteCount;
```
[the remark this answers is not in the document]
xinyanmsft: I didn't realize StreamDemuxer is used in other scenarios as well. Added enum StreamType, default to RemoteCommand (pick the name RemoteCommand since exec and attach are implemented under kubelet\server\remotecommand directory). StreamType.RemoteCommand does not perform extra handling. If StreamType.PortForward is specified, StreamDemuxer will filter out the port bytes.

src/KubernetesClient/StreamDemuxer.cs
```diff
@@ -199,6 +204,13 @@ protected async Task RunLoop(CancellationToken cancellationToken)
...
+                    if (this.streamType == StreamType.PortForward && byteCount <= 2)
```
brendandburns: I'm not sure this will work correctly, what if byteCount becomes greater than 2 immediately, you won't strip the first two bytes.

I think the right thing to do is to have a `boolean skippedPort` outside the loop and then clip the two bytes at the front if `!skippedPort` and then set `skippedPort` to true...

Eitherway, I think this deserves a unit test to make sure it's working.

Thanks!

xinyanmsft: here's another pass. there was a mistake in my previous change where I didn't handle StreamDemuxer with multiple streams. Now there is a dictionary that holds the skipped bytes for each stream.
it turns out there is 0 StreamDemuxer tests. add StreamDemuxerTests that covers send / receive for both command and port forwarding. Added WebSocketMock as a helper.

brendandburns:
Awesome, thanks!

Issue #235, opened 2019-01-31
Loading of client certificates (PFX) crashes in Azure Functions sandbox


maganvill:
Happens when app is deployed as Azure Function, not dev time. 

Root cause is that CertUtil [creates a new X509Certificate2](https://github.com/kubernetes-client/csharp/blob/master/src/KubernetesClient/CertUtils.cs#L93) from client certificate byte array without setting the binding flags to "X509KeyStorageFlags.MachineKeySet | X509KeyStorageFlags.PersistKeySet | X509KeyStorageFlags.Exportable" that is needed, that I learnt from previous experience in loading client certificates in Azure functions.

My idea is to let developer set the required flags in [KubernetesClientConfiguration](https://github.com/kubernetes-client/csharp/blob/master/src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs) (default null of course). I will submit a PR. 

Let's hope that you professionals will comment and lead me on the right way if wrong =)

Exception:

> The SSL connection could not be established, see inner exception.
System.Net.Http.HttpRequestException : The SSL connection could not be established, see inner exception. ---> System.ComponentModel.Win32Exception : **The credentials supplied to the package were not recognized**
   at System.Net.SSPIWrapper.AcquireCredentialsHandle(SSPIInterface secModule,String package,CredentialUse intent,SCHANNEL_CRED scc)
   at System.Net.Security.SslStreamPal.AcquireCredentialsHandle(CredentialUse credUsage,SCHANNEL_CRED secureCredential)
   at System.Net.Security.SslStreamPal.AcquireCredentialsHandle(X509Certificate certificate,SslProtocols protocols,EncryptionPolicy policy,Boolean isServer)
   at System.Net.Security.SecureChannel.AcquireClientCredentials(Byte[]& thumbPrint)
   at System.Net.Security.SecureChannel.GenerateToken(Byte[] input,Int32 offset,Int32 count,Byte[]& output)
   at System.Net.Security.SecureChannel.NextMessage(Byte[] incoming,Int32 offset,Int32 count)
   at System.Net.Security.SslState.StartSendBlob(Byte[] incoming,Int32 count,AsyncProtocolRequest asyncRequest)
   at System.Net.Security.SslState.ProcessReceivedBlob(Byte[] buffer,Int32 count,AsyncProtocolRequest asyncRequest)
[...]

PR #236, opened 2019-01-31
Introducing KubernetesClientConfiguration.ClientCertificateKeyStoreFl…

maganvill:
…ags to specify how private key is imported.

See #235


maganvill:
Commit with wrong user

Issue #234, opened 2019-01-30
LoadPemFileCert() does not handle bundle certificates


etchang:
We currently working with a customer who is running the C# Kubernetes Client from within their cluster. They are providing their own CA certificate in a PEM file that contains the entire Certificate Trust Chain in it as well as comments
i.e.

```
-----BEGIN CERTIFICATE-----
<base 64 string>

-----END CERTIFICATE-----

subject=/CN=<subject>
issuer=/CN=<issuer>

-----BEGIN CERTIFICATE-----
<base 64 string>

-----END CERTIFICATE-----

subject=/CN=<subject>
issuer=/CN=<issuer>

-----BEGIN CERTIFICATE-----
<base 64 string>

-----END CERTIFICATE-----
```

 
Calling _KubernetesClientConfiguration.InClusterConfig()_ results in the following stack:
```

System.FormatException: The input is not a valid Base-64 string as it contains a non-base 64 character, more than two padding charcters, or an illegal character among the padding characters,
    at System.Convert.FromBase64_ComputeResultLength(Char* inputPtr, Int32 inputLength)
	at System.Convert.FromBase64CharPtr(Char* inputPtr, Int32 inputLength)
	at System.Convert.FromBase64String(String s)
	at k8s.CertUtils.LoadPemFileCert(String file)
	at k8s.KubernetesClientConfiguration.InClusterConfig()
	
```
	
It looks like the client's certificate parsing currently only handles comment-less cert files with a single certificate in them.
[https://github.com/kubernetes-client/csharp/blob/master/src/KubernetesClient/CertUtils.cs](url)
```
        /// <summary>
        /// Load pem encoded cert file
        /// </summary>
        /// <param name="file">Path to pem encoded cert file</param>
        /// <returns>x509 instance.</returns>
        public static X509Certificate2 LoadPemFileCert(string file)
        {
            var certdata = File.ReadAllText(file)
                .Replace("-----BEGIN CERTIFICATE-----", "")
                .Replace("-----END CERTIFICATE-----", "")
                .Replace("\r", "")
                .Replace("\n", "");

            return new X509Certificate2(Convert.FromBase64String(certdata));
        }
```

Since Kubernetes supports supplying custom certs to the cluster, the client should also be able to handle certificates formats allowed by OpenSSL.

tg123:
I will take a look

brendandburns:
@tg123 are you working on this? I have some cycles and thought I might take a crack at it if you haven't already started.

[...]

PR #245, opened 2019-02-15
pretend to support chain

tg123:
pretend to support #234

In current design, `X509Certificate2` is used to represent `CaCert` which is not a chain.
However, the bug #234 is talking about loading pem failed from pem which contains multiple certs.

This PR fixed the loading issue, but the chain should be supported in `K8SConfiguration` in future.


brendandburns:
Thanks @tg123 !!

PR #250, opened 2019-03-05
Improve README.md with better example and descriptions.

### LINKED ISSUE #249: Provide documentation on authentication mechanisms
It looks like there are a few ways to authenicate/connect to a cluster.

```csharp
var config = KubernetesClientConfiguration.BuildConfigFromConfigFile();
var client = new Kubernetes(config);
```

Is this documented somewhere? If not, plans to provide this to clarify this part?

brendandburns:
Fixes #249

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,33 +10,74 @@
 dotnet add package KubernetesClient
 ```
 
-# Generating the Client Code
-
-## Prerequisites
-
-You'll need a Linux machine with Docker.
-
-The generated code works on all platforms supported by .NET or .NET Core.
-
-Check out the generator project into some other directory
-(henceforth `$GEN_DIR`)
-
-```bash
-cd $GEN_DIR/..
-git clone https://github.com/kubernetes-client/gen
-```
-
-## Generating code
-
-```bash
-# Where REPO_DIR points to the root of the csharp repository
-cd ${REPO_DIR}/csharp/src/KubernetesClient
-${GEN_DIR}/openapi/csharp.sh generated ../csharp.settings
-```
-
 # Usage
 
-## Running the Examples
+## Authentication/Configuration
+You should be able to use an standard KubeConfig file with this library,
+see the `BuildConfigFromConfigFile` function below. Most authentication
+methods are currently supported, but a few are not, see the 
+[known-issues](https://github.com/kubernetes-client/csharp#known-issues)
+
+You should also be able to authenticate using the in-cluster service
+account using the `InClusterConfig` function shown below.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -10,33 +10,74 @@
...
+var config = KubernetesClientConfiguration.InClusterConfig()
+
+// Use the config object to create a client.
+var client = new Kubernetes(config);
 ```
 
-## Generating code
+### Listing Objects
+```c#
+var namespaces = client.ListNamespace();
```
itowlson: Nitpick: We should prefer showing the async methods to the blocking ones.  It is easy to translate in either direction of course, but we should set the expectation that async is the 'normal' way to do it.

README.md
```diff
@@ -10,33 +10,74 @@
...
+    Metadata = new V1ObjectMeta
+    {
+        Name = "test"
+    }
+};
+
+var result = client.CreateNamespace(ns);
+Console.WriteLine(result);
+
+var result = client.CreateNamespace(ns);
```
itowlson: Not clear why you're doing this twice.  Is it a copypasta, or are you intending to show the behaviour if a user creates an already-created resource?  If the latter then suggest commenting it to explain the expected output.

itowlson: Actually I think it must be a copypasta because redeclaring `result` won't compile will it?

README.md
```diff
@@ -10,33 +10,74 @@
...
+Console.WriteLine(result);
+
+var status = client.DeleteNamespace(ns.Metadata.Name, new V1DeleteOptions());
 ```
 
-# Usage
+### Deleting Objects
```
itowlson: Covered under `Creating and Deleting Objects` or did you mean to move the delete code from there to under here?

README.md
```diff
@@ -10,33 +10,74 @@
...
+```c#
```
itowlson: Possible unclosed Markdown tag?

README.md
brendandburns: yep, copy/paste

README.md
brendandburns: the examples are all sync right now. Let's do this in a separate PR.

README.md
brendandburns: oops. fixed.

README.md
brendandburns: fixed.

brendandburns:
@itowlson friendly ping on this one :)

brendandburns:
@itowlson pingity ping ping :)

Thanks!

PR #252, opened 2019-03-06
generator cmd param

tg123:
Change `watch generator` to `accepting csharp.setting` same as what `project gen` does.

This is a always a part to do the code generating on azure pipeline ❤️
<https://farmer1992.visualstudio.com/opensources/_release?view=mine&definitionId=7>


brendandburns:
/lgtm
/approve

Thanks!

PR #253, opened 2019-03-07
Add bundle certificate support

### LINKED ISSUE #234: LoadPemFileCert() does not handle bundle certificates

etchang:
Add support for cluster CA bundle certificates.

This changes the client to load potentially multiple certificates from the cluster's _ca.crt_ file. When the client goes to validate an SSL connection, it will add these certs to the certificate chain provided by the server's response and validate that it can rebuild the chain. It will then make sure that the last certificate in the chain matches one of the certificates loaded from _ca.crt_ to ensure the chain can be  #trusted.

Fixes #234


src/KubernetesClient/CertUtils.cs
```diff
@@ -19,22 +18,32 @@ public static class CertUtils
         /// Load pem encoded cert file
         /// </summary>
         /// <param name="file">Path to pem encoded cert file</param>
-        /// <returns>x509 instance.</returns>
-        public static X509Certificate2 LoadPemFileCert(string file)
+        /// <returns>List of x509 instances.</returns>
+        public static IList<X509Certificate2> LoadPemFileCert(string file)
```
tg123: should be `X509Certificate2Collection`

src/KubernetesClient/CertUtils.cs
```diff
@@ -19,22 +18,32 @@ public static class CertUtils
...
         {
-            var certs = new X509CertificateParser().ReadCertificates(File.OpenRead(file));
-            var store = new Pkcs12StoreBuilder().Build();
-            foreach (X509Certificate cert in certs)
```
tg123: here, already read all certs in the file.
the reason did not return a collection is due to the the return type is `X509Certificate2` before.

src/KubernetesClient/CertUtils.cs
```diff
@@ -19,22 +18,32 @@ public static class CertUtils
...
-            using (var pkcs = new MemoryStream())
+            var certdata = File.ReadAllText(file)
+                .Replace("\r", "")
+                .Replace("\n", "");
+
+            // Retrieve all certificates from the file
+            //
+            var r = new Regex("-----BEGIN CERTIFICATE-----(.*?)-----END CERTIFICATE-----");
+
+            var matches = r.Matches(certdata);
```
tg123: suggest to be `new X509CertificateParser().ReadCertificates` instead of building it ourselves.

src/KubernetesClient/Kubernetes.ConfigInit.cs
```diff
@@ -257,15 +260,32 @@ private void SetCredentials(KubernetesClientConfiguration config, HttpClientHand
             {
                 chain.ChainPolicy.RevocationMode = X509RevocationMode.NoCheck;
 
-                // add all your extra certificate chain
-                chain.ChainPolicy.ExtraStore.Add(caCert);
+                // Added our trusted certificates to the chain
+                //
+                foreach (var caCert in caCerts)
+                {
+                    chain.ChainPolicy.ExtraStore.Add(caCert);
```
tg123: AddRange

src/KubernetesClient/Kubernetes.ConfigInit.cs
```diff
@@ -257,15 +260,32 @@ private void SetCredentials(KubernetesClientConfiguration config, HttpClientHand
...
 
+                var isTrusted = false;
+
                 var rootCert = chain.ChainElements[chain.ChainElements.Count - 1].Certificate;
-                isValid = isValid && rootCert.RawData.SequenceEqual(caCert.RawData);
 
-                return isValid;
+                // Make sure that one of our trusted certs exists in the chain provided by the server.
+                //
+                foreach (var cert in caCerts)
```
tg123: IMHO, `chain.Build` is enough to validating the server certificate when ca bundles were added. 
I think we can remove the comparison with root cert during this change

tests/KubernetesClient.Tests/CertUtilsTests.cs
```diff
@@ -70,15 +70,5 @@ public void LoadFromInlineDataRelativePath()
             var cert = CertUtils.GeneratePfx(cfg);
             Assert.NotNull(cert.PrivateKey);
         }
-
-        /// <summary>
-        /// Checks 
-        /// </summary>
-        [Fact]
-        public void LoadPemWithMultiCert()
```
tg123: this test is to ensure no regression when loading bundle.

src/KubernetesClient/CertUtils.cs
etchang: Definitely better. Only caveat is that we need to do some additional steps to convert the bouncy x509 certificates to .Net crypt x509 certs to store them in an X509Certificate2Collection, but that is trivial.

tests/KubernetesClient.Tests/CertUtilsTests.cs
etchang: I've updated the test to make sure the certificates we've loaded match the original certificates we meant to load. Since the implementation of _LoadPemFileCert_ has changed this test is a bit different than what you originally wrote. Let me know if there are any other preferred way to test this as the CertificateValidationTests now also include tests that verify that we can properly build certificates using the bundle.

src/KubernetesClient/CertUtils.cs
etchang: Done

src/KubernetesClient/CertUtils.cs
etchang: Added this back

src/KubernetesClient/Kubernetes.ConfigInit.cs
etchang: Done

src/KubernetesClient/Kubernetes.ConfigInit.cs
etchang: I kept this around because of [this issue](https://github.com/dotnet/corefx/issues/30284) which was resolved in [this](https://github.com/kubernetes-client/csharp/pull/174) pr. 

From the post it seems like because we are using the flag to not validate the certificate authority that we need to do a manual check to make sure that we trust the certificate chain (based on our CA certs)

src/KubernetesClient/Kubernetes.ConfigInit.cs
```diff
@@ -152,7 +154,7 @@ public Kubernetes(KubernetesClientConfiguration config, params DelegatingHandler
         }
 #endif
 
-        private X509Certificate2 CaCert { get; }
+        private X509Certificate2Collection CaCert { get; }
```
tg123:
```suggestion
        private X509Certificate2Collection CaCerts { get; }
```

src/KubernetesClient/Kubernetes.ConfigInit.cs
```diff
@@ -58,7 +58,9 @@ public Kubernetes(KubernetesClientConfiguration config, params DelegatingHandler
                         throw new KubeConfigException("a CA must be set when SkipTlsVerify === false");
                     }
 
-                    using (System.IO.MemoryStream certStream = new System.IO.MemoryStream(config.SslCaCert.RawData))
+                    var caCert = CaCert[CaCert.Count - 1];
```
tg123: could it support collection here?

src/KubernetesClient/Kubernetes.ConfigInit.cs
etchang: Updated this across the project

src/KubernetesClient/Kubernetes.ConfigInit.cs
etchang: I'm not familiar with the Java libraries, but I changed it to at least add all the certs in the bundle to the trusted certificate list.

brendandburns:
@tg123 I defer to your review, LGTM it when you think it's ready.

Issue #140, opened 2018-04-24
Support for Mono/Xamarin


JonJam:
I am trying to use this client library in a Xamarin app, but I having issues due to HttpClientHandler.ServerCertificateCustomValidationCallback in Kubernetes.ConfigInit.

This throws a NotImplementedException in Xamarin as Mono doesn't implement it, see https://github.com/mono/mono/blob/master/mcs/class/System.Net.Http/System.Net.Http/HttpClientHandler.cs.

brendandburns:
I'm happy to take a PR to optionalize this on Mono, but that functionality is needed for most kubernetes configs.

JonJam:
PR Merged

tg123:
Hi @JonJam 
I am trying to clean targets

Could you please confirm `netstandard2.0` works fine with your app?
I tested locally but I am new to xamarin not quite sure if I am doing right.

Testcases would be great

JonJam:
Hi @tg123,

The source for my app is here: https://github.com/JonJam/kubemob, although I am no longer maintaining it.

I can confirm it does work with netstandard2.0 as I am referencing this from a netstandard2.0 shared class library.

PR #258, opened 2019-03-20
Make MSBuild.Sdk.Extras private, it is a build-time only dependency.

qmfrederik:
Since #140 the [KubernetesClient package took a public dependency on MSBuild.Sdk.Extras](https://www.nuget.org/packages/KubernetesClient/1.5.1-g60ffd20a7c).

MSBuild.Sdk.Extras is a (very specialized) build-time dependency, and we shouldn't force users of the KubernetesClient to take a dependency on this. This PR marks it as private instead.


qmfrederik:
@JonJam do you know why this dependency was added?

JonJam:
@qmfrederik This dependency was added to support targeting Xamarin

PR #261, opened 2019-03-20
Update CI dependencies

qmfrederik:
- Use .NET Core 2.2 on both Linux and macOS
- Use Kubernetes v1.13.4 for the integration tests


qmfrederik:
@brendanburns I updated the Travis configuration to use an Ubuntu Xenial image (instead of Trusty), so I'll need a new LGTM for this one.

PR #264, opened 2019-03-22
Add code coverage, Linux & macOS leg to Azure CI

qmfrederik:
This PR adds a macOS and Linux leg to the Azure CI pipeline script.
It reuses the CI scripts which are used on Travis.

Especially the Linux leg is of interest as it runs integration tests against a real Kubernetes cluster.

It also sets up code coverage reports which show line and branch coverage.

image

### DIFF
--- a/.azure-pipelines.yaml
+++ b/.azure-pipelines.yaml
@@ -1,6 +1,6 @@
 jobs:
 
-- job: build_xamarin
+- job: xamarin_build
   pool:
     vmImage: 'VS2017-Win2016'
   steps:
@@ -21,10 +21,9 @@ jobs:
     inputs:
       PathtoPublish: '$(Build.ArtifactStagingDirectory)/nupkg/'
 
-- job: Build
+- job: windows_build
   pool:
     vmImage: 'VS2017-Win2016'
-
 
   steps:
   - task: DotNetCoreCLI@2
@@ -37,22 +36,6 @@ jobs:
     inputs:
       projects: '**/*.sln'
 
-  # - task: MSBuild@1
-  #   displayName: 'Build solution **/*.sln'
-  #   inputs:
-  #     msbuildArchitecture: x64
-
-  #     configuration: Release
-
-  # - task: VSTest@2
-  #   displayName: 'VsTest - testAssemblies'
-  #   inputs:
-  #     testAssemblyVer2: |
-  #       tests\**\*Tests*.dll
-  #       !**\obj\**
-
-  #     codeCoverageEnabled: true
-
   - task: DotNetCoreCLI@2
     displayName: 'dotnet test'
     inputs:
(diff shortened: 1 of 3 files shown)


tg123:
awesome

/LGTM

tg123:
why resharper code check was removed?

.azure-pipelines.yaml
```diff
@@ -45,30 +28,66 @@ jobs:
...
-
   - task: DotNetCoreCLI@2
     displayName: 'dotnet pack'
     inputs:
       command: pack
       packagesToPack: src/KubernetesClient/KubernetesClient.csproj
       packDirectory: '$(Build.ArtifactStagingDirectory)/nupkg'
-      majorVersion: 1
-      minorVersion: 4      
-      versioningScheme: byPrereleaseNumber
```
tg123: why versioning was removed?
it is to replace git verison

qmfrederik:
The resharper check always failed when I ran this in my instance of VSTS; do I need any specific configuration to make that work?

.azure-pipelines.yaml
```diff
@@ -45,30 +28,66 @@ jobs:
       codeCoverageTool: 'cobertura'
       summaryFileLocation: '$(System.DefaultWorkingDirectory)/**/coverage.cobertura.xml'  
 
-  - task: alanwales.resharper-code-analysis.custom-build-task.ResharperCli@1
```
tg123: resharper?

tg123:
install and try again
<https://marketplace.visualstudio.com/items?itemName=alanwales.resharper-code-analysis>

.azure-pipelines.yaml
qmfrederik: The version numbers are embedded in the assemblies and NuGet packages you create. If you use build numbers which depend on external factors (such as the date and time the package was built or the CI job number), your build output (the binaries) will be different every time you build, _even if you build from the same source_.

This breaks [reproducible builds](https://reproducible-builds.org/). Having reproducible builds means I can clone this repository, build locally, get the SHA hash of the NuGet package it builds, and make sure it's the same which is uploaded on NuGet. It helps enhance trust.

The Kubernetes project is working towards reproducible builds - see https://github.com/kubernetes/kubernetes/issues/70131, and I think it would be good if the same holds true for the .NET client.

brendandburns:
@qmfrederik this is great, let's restore the resharper check and then we can check in.

Many thanks!

qmfrederik:
Resharper is back, so we should be good.

.azure-pipelines.yaml
tg123: IMHO, I prefer to `each build is an individual`.  
The build machine cannot be exactly the same.
For example, when your update dotnet 1 to dotnet 2, the output/nuget of compiler might be changed even the source codes are totally the same.

The nuget is a snapshot and the release team is responsible for the `official release`.

Maybe something like docker image could make the `build reproducible`.

@brendandburns what is your opinion?

brendandburns:
@qmfrederik this needs a rebase then it's good to go.

qmfrederik:
@brendanburns This has been rebased and should be good to go

brendandburns:
I'm going to force push this in since we're hitting some sort of flake.

PR #265, opened 2019-03-22
Mark all Fractions classes as internal

qmfrederik:
#194 embedded the Fractions within the KubernetesClient.

Because most of the Fractions types are marked as `public`, any library referencing both KubernetesClient and the original Fractions library would be confused at compile time as there are two types with the same name in two different projects.

Hence, this PR makes all Fractions-related types `internal`, so they are not visible to any other library at compile time.

Libraries can also use reflection at runtime to find types. To guard against libraries being confused by finding two types with the same name in two different libraries, I've also renamed the `Fractions` to `k8s.Internal.Fractions`.

### DIFF
--- a/src/KubernetesClient/Fractions/Extensions/MathExt.cs
+++ b/src/KubernetesClient/Fractions/Extensions/MathExt.cs
@@ -1,6 +1,6 @@
 using System;
 
-namespace Fractions.Extensions {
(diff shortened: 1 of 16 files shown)


tg123:
/LGTM

BTW
is strong name the only reason to embed the library?

qmfrederik:
> Is strong name the only reason to embed the library?

I think it was, yes.

PR #281, opened 2019-05-01
add websocket auth support for .net standard 2.0 support

xinyanmsft:
WebSocket certificate validation is not implemented for NETSTANDARD2_0. This means, when KubernetesClient.dll is consumed by a process running such as Visual Studio, WebSocket will not work. This change enables it.


brendandburns:
close/re-open to re-triger CI (which will hopefully pass this time :)

xinyanmsft:
the build succeeded. thanks for fixing it.

PR #283, opened 2019-05-04
Rev minikube version and disable ubuntu build.

brendandburns:
See if this fixes our CI...

@xinyanmsft @qmfrederik

PR #284, opened 2019-05-13
WIP - Re-enable Ubuntu build in CI

tomkerkhove:
Re-enable Ubuntu build in CI

Follow-up of #283

### DIFF
--- a/.azure-pipelines.yaml
+++ b/.azure-pipelines.yaml
@@ -98,9 +98,9 @@ jobs:
     vmImage: 'ubuntu-16.04'
   steps:
   - task: DockerInstaller@0
-    displayName: 'Install Docker'
+    displayName: 'Install Docker version 17.09.0-ce'
     inputs:
-      dockerVersion: 18.09.0-ce
+      dockerVersion: 17.09.0-ce
 
   - script: |
       ./install-linux.sh
--- a/.azure-pipelines.yaml
+++ b/.azure-pipelines.yaml
@@ -97,6 +97,11 @@ jobs:
   pool:
     vmImage: 'ubuntu-16.04'
   steps:
+  - task: DockerInstaller@0
+    displayName: 'Install Docker'
+    inputs:
+      dockerVersion: 18.09.0-ce
+
   - script: |
       ./install-linux.sh
     displayName: 'Install .NET & set up minikube'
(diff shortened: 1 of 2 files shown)


tomkerkhove:
Looks like the build is hanging because of minikube 🤔

tomkerkhove:
I've added logs for tracing what version for kubectl & minikube is being installed, but they don't show up somehow.

That said, I've changed minikube commands to use `--logtostderr` so that we see what is wrong.

Turns out we have a couple of issues:
- `unsupported docker version: 3.0.4`
- `Error reading config file at /home/vsts/.minikube/config/config.json: open /home/vsts/.minikube/config/config.json: no such file or directory` - Although this can be side-effect, but not sure

qmfrederik:
@tomkerkhove That's odd, perhaps you can add a `docker -v` to see what version of Docker is installed. Kubernetes is a bit picky with which version of Docker it supports - [the supported versions now are 1.13.1, 17.03, 17.06, 17.09, 18.06, 18.09](https://kubernetes.io/docs/setup/release/notes/#external-dependencies).

Azure DevOps regularly upgrades the software installed on the agents, which can cause sudden build breaks like this. (Travis is much more conservative in that respect)

tomkerkhove:
I'll add the docker version to the output and make sure we install a specific version so that we run the correct version if we can.

Looks like they've changed it recently: https://github.com/microsoft/azure-pipelines-image-generation/commit/8e6db92b5d9ebddbfd2d52309de92dd83a709cd3#diff-f87405c9dd43f633a04dd078666d61af

qmfrederik:
Yeah, so that means we'll probably need some packaging voodoo to uninstall the version of Docker that comes pre-installed, and replace it with a version which is supported by Kubernetes.

tomkerkhove:
Not really, they actually have it out of the box - https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/tool/docker-installer?view=azure-devops

Will update the PR in a few hours

tomkerkhove:
I'll abandon this PR for now and circle back once I get this working.

Don't want to burn too much build time/spam people

tomkerkhove:
> Not really, they actually have it out of the box - https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/tool/docker-installer?view=azure-devops

Nevermind, this is just the CLI

Issue #263, opened 2019-03-21
Requests hang if watch=true and there are no objects matching query


corruptmem:
To reproduce:

```csharp
var config = KubernetesClientConfiguration.BuildDefaultConfig();
var client = new Kubernetes(config);

Console.WriteLine("FOO");
var podStream = await client.ListNamespacedPodWithHttpMessagesAsync(
    "default",
    labelSelector: "app=1234", // <----- TRY COMMENTING ME OUT
    watch: true);

Console.WriteLine("BAR");

podStream.Watch<V1Pod>(
    (evt, pod) => Console.WriteLine($"{evt}: {pod?.Metadata?.Name}"),
    Console.WriteLine,
    () => Environment.Exit(0));

Console.WriteLine("WATCHING");
Console.ReadKey();
```

**Expected**: Output FOO, BAR, WATCHING then as any new pods get created, for the event to be printed

**Actual**: Outputs FOO then the program hangs until at least one pod gets created

If you comment out the labelSelector, the problem goes away - presumably unless you had a Kubernetes cluster with no pods in the default namespace.

---

Although it is possible to workaround this by treating the ListNamespacedPodWithHttpMessagesAsync (and similar) calls as taking an indefinite time to return under normal circumstances, it's not what I expected to happen when using the watcher.

The root cause of the issue appears to be the LineSeparatedHttpContent class used in the WatcherDelegatingHandler. It appears that inside SerializeToStreamAsync it "peeks" the first line which then calls ReadLineAsync and blocks until there's something to read.

tg123:
I think it is by design.
call with `watch=true` would return a stream, something like you are downloading a big file.
the call will hang until everything was read if there is no `WatcherDelegatingHandler` and it is impossible due to watched stream never end until timeout.

Maybe the better solution is to wrap List and Watch into one sugared watch

What do you think?

corruptmem:
@tg123 This is essentially what I am now doing, but it makes it all more difficult than it needs to be. Essentially having to translate the async-based List call into working the same way as the callback-based Watch call.

However I think it is possible to implement the List call so it returns immediately if watch is true. I implemented it as a test by having the LineSeparatedHttpContent.SerializeToStreamAsync method not bother peeking the stream, which makes the HttpContent stream (as returned by HttpContent.ReadAsStreamAsync) empty, rather than the first line of the watch result. 

It works for my purposes, but I'm not sure what doing that might break in the rest of the client.

brendandburns:
[...]

PR #300, opened 2019-08-27
improved async support for watches

blushingpenguin:
This improves cancellation support for watches.  Currently cancellation only works when data is returned from the server because the Watcher object ends up calling StreamReader.ReadLineAsync() to fetch data from the http response and that API does not actually support cancellation.

Although this can be worked around by closing the underlying stream from the http response, If the initial List* call does not receive a response from the server (because there are no events to report yet) then the response object is of course unavailable and there is no way to cancel the List* call.  To fix this I've just avoided using StreamReader and implemented a simple line stream class that is used during watching that does support cancellation.

I am also not really clear on why the List* calls fetch the first line of the response when watching -- it would seem reasonable to wait for the subsequent Watch call to return the data -- however with cancellation support added it's not a big deal either way.  (There's another open case about that: https://github.com/kubernetes-client/csharp/issues/263).

I've also added an asynchronous watch interface which can be used like this:

```csharp
var result = await _k8s.ListNamespacedPodWithHttpMessagesAsync(
    "default",
    watch: true,
    cancellationToken: cancellationToken
).ConfigureAwait(false);
// ... do stuff with result
await result.WatchAsync<V1Pod>(cancellationToken, (cancellationToken, type, pod) => {
    // ... do stuff asynchronously
    return Task.CompletedTask;
});
```

This propagates the async and cancellation support to the callbacks as well as being more 'natural' for async code by implementing watching simply as an asynchronous function.

Please let me know what you think!

Thanks,

Mark

### DIFF
--- a/src/KubernetesClient/PeekableStreamReader.cs
+++ b/src/KubernetesClient/PeekableStreamReader.cs
@@ -1,0 +1,110 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace k8s
+{
+    internal class PeekableStreamReader : IAsyncLineReader
+    {
+        private readonly Queue<string> _peek = new Queue<string>();
+        private Stream _stream;
+        private byte[] _buf = new byte[4096];
+        private int _bufPos = 0;
+        private int _bufLength = 0;
+        private bool _eof = false;
+
+        public PeekableStreamReader(Stream stream)
+        {
+            _stream = stream;
+        }
+
+        private void ShuffleConsumedBytes(byte[] dst)
+        {
+            if (_bufPos < _bufLength)
+            {
+                Buffer.BlockCopy(_buf, _bufPos, dst, 0, _bufLength - _bufPos);
+            }
+            _bufLength -= _bufPos;
+            _bufPos = 0;
+        }
+
+        private void CheckDisposed()
+        {
+            if (_stream == null)
+            {
+                throw new ObjectDisposedException("PeekableStreamReader");
+            }
+        }
+
+        public async Task<string> ReadLineAsync(CancellationToken cancellationToken)
+        {
+            CheckDisposed();
+
+            if (_peek.Count > 0)
+            {
+                return _peek.Dequeue();
+            }
+
+            for (;;)
+            {
+                // read buffered data
+                if (_bufPos < _bufLength)
+                {
+                    int nlPos = _eof ? _bufLength - 1 : Array.IndexOf(_buf, (byte)'\n', _bufPos, _bufLength -_bufPos);
+                    if (nlPos >= 0)
+                    {
+                        string result = Encoding.UTF8.GetString(_buf, _bufPos, nlPos - _bufPos + 1);
+                        _bufPos = nlPos + 1;
+                        return result;
+                    }
+                }
+                if (_eof)
+                {
+                    return null;
+                }
+                // make the buffer bigger if needed
+                if (_buf.Length - _bufLength < 512)
+                {
+                    if (_buf.Length > Int32.MaxValue / 2)
+                    {
+                        throw new InvalidOperationException(
+                            "trying to read a line > 2Gb from the server");
+                    }
+                    byte[] newbuf = new byte[_buf.Length * 2];
+                    ShuffleConsumedBytes(newbuf);
+                }
+                // consume any previously read data
+                else if (_bufPos > 0)
+                {
+                    ShuffleConsumedBytes(_buf);
+                }
+
+                int read = await _stream.ReadAsync(_buf, _bufLength,
+                    _buf.Length - _bufLength, cancellationToken).ConfigureAwait(false);
+                _bufLength += read;
+                _eof = read == 0;
+            }
+        }
+
+        public async Task<string> PeekLineAsync(CancellationToken cancellationToken)
+        {
+            CheckDisposed();
+
+            var line = await ReadLineAsync(cancellationToken).ConfigureAwait(false);
+            _peek.Enqueue(line);
+            return line;
+        }
+
+        public void Dispose()
+        {
+            if (_stream != null)
+            {
+                _stream.Dispose();
+                _stream = null;
+            }
+        }
+    }
+}
--- a/src/KubernetesClient/PeekableStreamReader.cs
+++ b/src/KubernetesClient/PeekableStreamReader.cs
@@ -7,8 +7,9 @@ using System.Threading.Tasks;
 
 namespace k8s
 {
-    internal class PeekableStreamReader : IAsyncLineReader
+    public class PeekableStreamReader : IAsyncLineReader
     {
+        private readonly int _maxLineLength;
         private readonly Queue<string> _peek = new Queue<string>();
         private Stream _stream;
         private byte[] _buf = new byte[4096];
@@ -16,8 +17,9 @@ namespace k8s
         private int _bufLength = 0;
         private bool _eof = false;
 
-        public PeekableStreamReader(Stream stream)
+        public PeekableStreamReader(Stream stream, int maxLineLength = Int32.MaxValue)
         {
+            _maxLineLength = maxLineLength;
             _stream = stream;
         }
 
@@ -39,21 +41,15 @@ namespace k8s
             }
         }
 
-        public async Task<string> ReadLineAsync(CancellationToken cancellationToken)
+        private async Task<string> ReadLineNoPeekAsync(CancellationToken cancellationToken)
         {
-            CheckDisposed();
 
-            if (_peek.Count > 0)
-            {
-                return _peek.Dequeue();
-            }
-
-            for (;;)
+            for (; ; )
             {
                 // read buffered data
                 if (_bufPos < _bufLength)
                 {
-                    int nlPos = _eof ? _bufLength - 1 : Array.IndexOf(_buf, (byte)'\n', _bufPos, _bufLength -_bufPos);
+                    int nlPos = _eof ? _bufLength - 1 : Array.IndexOf(_buf, (byte)'\n', _bufPos, _bufLength - _bufPos);
                     if (nlPos >= 0)
                     {
                         string result = Encoding.UTF8.GetString(_buf, _bufPos, nlPos - _bufPos + 1);
@@ -73,7 +69,7 @@ namespace k8s
                 // make the buffer bigger if needed
                 if (_buf.Length - _bufLength < 512)
                 {
-                    if (_buf.Length > Int32.MaxValue / 2)
+                    if (_buf.Length > _maxLineLength / 2)
                     {
                         throw new InvalidOperationException(
                             "trying to read a line > 2Gb from the server");
@@ -90,11 +86,22 @@ namespace k8s
             }
         }
 
+        public async Task<string> ReadLineAsync(CancellationToken cancellationToken)
+        {
+            CheckDisposed();
+
+            if (_peek.Count > 0)
+            {
+                return _peek.Dequeue();
+            }
+            return await ReadLineNoPeekAsync(cancellationToken);
+        }
+
         public async Task<string> PeekLineAsync(CancellationToken cancellationToken)
         {
             CheckDisposed();
 
-            var line = await ReadLineAsync(cancellationToken).ConfigureAwait(false);
+            var line = await ReadLineNoPeekAsync(cancellationToken).ConfigureAwait(false);
             _peek.Enqueue(line);
             return line;
         }
--- a/src/KubernetesClient/PeekableStreamReader.cs
+++ b/src/KubernetesClient/PeekableStreamReader.cs
@@ -19,8 +19,14 @@ namespace k8s
 
         public PeekableStreamReader(Stream stream, int maxLineLength = Int32.MaxValue)
         {
+            // This is arbitrary, but ~8k lines are not abnormal
+            if (maxLineLength < 32768)
+            {
+                throw new ArgumentOutOfRangeException(nameof(maxLineLength),
+                    "The maximum line length must be at least 32768");
+            }
             _maxLineLength = maxLineLength;
-            _stream = stream;
+            _stream = stream ?? throw new ArgumentNullException(nameof(stream));
         }
 
         private void ShuffleConsumedBytes(byte[] dst)
@@ -67,14 +73,17 @@ namespace k8s
                     ShuffleConsumedBytes(_buf);
                 }
                 // make the buffer bigger if needed
-                if (_buf.Length - _bufLength < 512)
+                int left = _buf.Length - _bufLength;
+                if (left < 512)
                 {
-                    if (_buf.Length > _maxLineLength / 2)
+                    int newlen = Math.Min(_maxLineLength / 2, _buf.Length) * 2;
+                    if (newlen <= _buf.Length)
                     {
                         throw new InvalidOperationException(
-                            "trying to read a line > 2Gb from the server");
+                            "Reading from the stream failed because a line from the server was unexpectedly " +
+                            $"greater than the configured maximum line length of {_maxLineLength}");
                     }
-                    byte[] newbuf = new byte[_buf.Length * 2];
+                    byte[] newbuf = new byte[newlen];
                     ShuffleConsumedBytes(newbuf);
                     _buf = newbuf;
                 }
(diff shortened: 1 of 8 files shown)


blushingpenguin:
The failing test is k8s.Tests.AuthTests.Cert which I don't think I've broken -- it runs on my linux box ok.

```
Starting test execution, please wait...
[xUnit.net 00:00:13.05]     k8s.Tests.WatchTests.DirectWatchEventsWithTimeout [SKIP]
[xUnit.net 00:00:13.16]     k8s.Tests.WatchTests.WatcherIntegrationTest [SKIP]
Skipped  k8s.Tests.WatchTests.DirectWatchEventsWithTimeout
Skipped  k8s.Tests.WatchTests.WatcherIntegrationTest
[xUnit.net 00:02:13.94]     k8s.Tests.AuthTests.Cert [FAIL]
Failed   k8s.Tests.AuthTests.Cert
Error Message:
 System.AggregateException : One or more errors occurred. (The SSL connection could not be established, see inner exception.)
---- System.Net.Http.HttpRequestException : The SSL connection could not be established, see inner exception.
-------- System.ComponentModel.Win32Exception : The credentials supplied to the package were not recognized
Stack Trace:
   at System.Threading.Tasks.Task`1.GetResultCore(Boolean waitCompletionNotification)
   at k8s.Tests.AuthTests.ExecuteListPods(IKubernetes client) in D:\a\1\s\tests\KubernetesClient.Tests\AuthTests.cs:line 36
   at k8s.Tests.AuthTests.Cert() in D:\a\1\s\tests\KubernetesClient.Tests\AuthTests.cs:line 229
----- Inner Stack Trace -----
   at System.Net.Http.ConnectHelper.EstablishSslConnectionAsyncCore(Stream stream, SslClientAuthenticationOptions sslOptions, CancellationToken cancellationToken)
   at System.Threading.Tasks.ValueTask`1.get_Result()
   at System.Net.Http.HttpConnectionPool.CreateConnectionAsync(HttpRequestMessage request, CancellationToken cancellationToken)
   at System.Threading.Tasks.ValueTask`1.get_Result()
   at System.Net.Http.HttpConnectionPool.WaitForCreatedConnectionAsync(ValueTask`1 creationTask)
   at System.Threading.Tasks.ValueTask`1.get_Result()
   at System.Net.Http.HttpConnectionPool.SendWithRetryAsync(HttpRequestMessage request, Boolean doRequestAuth, CancellationToken cancellationToken)
   at System.Net.Http.RedirectHandler.SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
   at k8s.WatcherDelegatingHandler.SendAsync(HttpRequestMessage request, CancellationToken cancellationToken) in D:\a\1\s\src\KubernetesClient\WatcherDelegatingHandler.cs:line 21
   at Microsoft.Rest.RetryDelegatingHandler.<>c__DisplayClass11_0.<<SendAsync>b__1>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at Microsoft.Rest.RetryDelegatingHandler.SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
   at System.Net.Http.HttpClient.FinishSendAsyncBuffered(Task`1 sendTask, HttpRequestMessage request, CancellationTokenSource cts, Boolean disposeCts)
   at k8s.Kubernetes.ListNamespacedPodWithHttpMessagesAsync(String namespaceParameter, Nullable`1 allowWatchBookmarks, String continueParameter, String fieldSelector, String labelSelector, Nullable`1 limit, String resourceVersion, Nullable`1 timeoutSeconds, Nullable`1 watch, String pretty, Dictionary`2 customHeaders, CancellationToken cancellationToken) in D:\a\1\s\src\KubernetesClient\generated\Kubernetes.cs:line 12257
----- Inner Stack Trace -----
   at System.Net.SSPIWrapper.AcquireCredentialsHandle(SSPIInterface secModule, String package, CredentialUse intent, SCHANNEL_CRED scc)
   at System.Net.Security.SslStreamPal.AcquireCredentialsHandle(CredentialUse credUsage, SCHANNEL_CRED secureCredential)
   at System.Net.Security.SslStreamPal.AcquireCredentialsHandle(X509Certificate certificate, SslProtocols protocols, EncryptionPolicy policy, Boolean isServer)
   at System.Net.Security.SecureChannel.AcquireClientCredentials(Byte[]& thumbPrint)
   at System.Net.Security.SecureChannel.GenerateToken(Byte[] input, Int32 offset, Int32 count, Byte[]& output)
   at System.Net.Security.SecureChannel.NextMessage(Byte[] incoming, Int32 offset, Int32 count)
   at System.Net.Security.SslState.StartSendBlob(Byte[] incoming, Int32 count, AsyncProtocolRequest asyncRequest)
   at System.Net.Security.SslState.ProcessReceivedBlob(Byte[] buffer, Int32 count, AsyncProtocolRequest asyncRequest)
   at System.Net.Security.SslState.StartReadFrame(Byte[] buffer, Int32 readBytes, AsyncProtocolRequest asyncRequest)
   at System.Net.Security.SslState.PartialFrameCallback(AsyncProtocolRequest asyncRequest)
--- End of stack trace from previous location where exception was thrown ---
   at System.Net.Security.SslState.ThrowIfExceptional()
   at System.Net.Security.SslState.InternalEndProcessAuthentication(LazyAsyncResult lazyResult)
   at System.Net.Security.SslState.EndProcessAuthentication(IAsyncResult result)
   at System.Net.Security.SslStream.EndAuthenticateAsClient(IAsyncResult asyncResult)
   at System.Net.Security.SslStream.<>c.<AuthenticateAsClientAsync>b__47_1(IAsyncResult iar)
   at System.Threading.Tasks.TaskFactory`1.FromAsyncCoreLogic(IAsyncResult iar, Func`2 endFunction, Action`1 endAction, Task`1 promise, Boolean requiresSynchronization)
--- End of stack trace from previous location where exception was thrown ---
   at System.Net.Http.ConnectHelper.EstablishSslConnectionAsyncCore(Stream stream, SslClientAuthenticationOptions sslOptions, CancellationToken cancellationToken)
Standard Output Messages:
 [Information] HttpsConnectionAdapter: Failed to authenticate HTTPS connection.
 System.IO.IOException: Authentication failed because the remote party has closed the transport stream.
    at System.Net.Security.SslState.StartReadFrame(Byte[] buffer, Int32 readBytes, AsyncProtocolRequest asyncRequest)
    at System.Net.Security.SslState.PartialFrameCallback(AsyncProtocolRequest asyncRequest)
 --- End of stack trace from previous location where exception was thrown ---
    at System.Net.Security.SslState.ThrowIfExceptional()
    at System.Net.Security.SslState.InternalEndProcessAuthentication(LazyAsyncResult lazyResult)
    at System.Net.Security.SslState.EndProcessAuthentication(IAsyncResult result)
    at System.Net.Security.SslStream.EndAuthenticateAsServer(IAsyncResult asyncResult)
    at System.Net.Security.SslStream.<>c.<AuthenticateAsServerAsync>b__50_2(IAsyncResult iar)
    at System.Threading.Tasks.TaskFactory`1.FromAsyncCoreLogic(IAsyncResult iar, Func`2 endFunction, Action`1 endAction, Task`1 promise, Boolean requiresSynchronization)
 --- End of stack trace from previous location where exception was thrown ---
    at Microsoft.AspNetCore.Server.Kestrel.Https.Internal.HttpsConnectionAdapter.InnerOnConnectionAsync(ConnectionAdapterContext context)
```

qmfrederik:
I had a glance at the PR, and it looks reasonable to me at first sight. You'll probably want to add a couple of unit tests for the `PeekableStreamReader`, though?

blushingpenguin:
Fair point, I've added tests that should cover all of the new code.

Thanks,

Mark

src/KubernetesClient/WatcherDelegatingHandler.cs
```diff
@@ -1,15 +1,23 @@
...
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.WebUtilities;
 
 namespace k8s
 {
+    /// <summary>
+    /// Represents line separated http content returned from ?watch=true calls
+    /// </summary>
+    public interface ILineSeparatedHttpContent
```
DavidParks8: This should probably be marked internal because it is an internal implementation detail that most people shouldn't have to worry about.

src/KubernetesClient/Watcher.cs
```diff
@@ -177,13 +183,119 @@ public static class WatcherExt
...
+        /// <param name="response">the api response</param>
+        /// <param name="onEvent">a callback when any event raised from api server</param>
+        /// <param name="onError">a callbak when any exception was caught during watching</param>
+        /// <param name="onClosed">
+        /// The action to invoke when the server closes the connection.
+        /// </param>
+        /// <param name="cancellationToken">A token that can be used to cancel the watch</param>
+        /// <returns>a watch object</returns>
+        public static async Task WatchAsync<T>(this HttpOperationResponse response,
+            CancellationToken cancellationToken,
```
DavidParks8: By convention, CancellationToken should be the last argument.

src/KubernetesClient/Watcher.cs
```diff
@@ -29,14 +29,14 @@ public class Watcher<T> : IDisposable
...
+        private readonly IAsyncLineReader _lineReader;
         private readonly Task _watcherLoop;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="Watcher{T}"/> class.
         /// </summary>
-        /// <param name="streamReader">
-        /// A <see cref="StreamReader"/> from which to read the events.
+        /// <param name="lineReader">
+        /// A <see cref="ILineReader"/> from which to read the events.
```
DavidParks8: "A" should "An"

src/KubernetesClient/PeekableStreamReader.cs
```diff
@@ -0,0 +1,118 @@
...
+
+        public async Task<string> ReadLineAsync(CancellationToken cancellationToken)
+        {
+            CheckDisposed();
+
+            if (_peek.Count > 0)
+            {
+                return _peek.Dequeue();
+            }
+            return await ReadLineNoPeekAsync(cancellationToken);
```
DavidParks8: Don't forget to ConfigureAwait(false);

src/KubernetesClient/PeekableStreamReader.cs
```diff
@@ -0,0 +1,118 @@
...
+        public async Task<string> PeekLineAsync(CancellationToken cancellationToken)
+        {
+            CheckDisposed();
+
+            var line = await ReadLineNoPeekAsync(cancellationToken).ConfigureAwait(false);
+            _peek.Enqueue(line);
+            return line;
+        }
+
+        public void Dispose()
```
DavidParks8: If this class must be public, it should implement the dispose pattern fully.

src/KubernetesClient/PeekableStreamReader.cs
```diff
@@ -0,0 +1,118 @@
...
+                {
+                    ShuffleConsumedBytes(_buf);
+                }
+                // make the buffer bigger if needed
+                if (_buf.Length - _bufLength < 512)
+                {
+                    if (_buf.Length > _maxLineLength / 2)
+                    {
+                        throw new InvalidOperationException(
+                            "trying to read a line > 2Gb from the server");
```
DavidParks8: Additional info could be helpful here. Perhaps adding a small sentence that clarifies that such sizes are not supported.

src/KubernetesClient/PeekableStreamReader.cs
```diff
@@ -0,0 +1,118 @@
...
+    {
+        private readonly int _maxLineLength;
+        private readonly Queue<string> _peek = new Queue<string>();
+        private Stream _stream;
+        private byte[] _buf = new byte[4096];
+        private int _bufPos = 0;
+        private int _bufLength = 0;
+        private bool _eof = false;
+
+        public PeekableStreamReader(Stream stream, int maxLineLength = Int32.MaxValue)
```
DavidParks8: Check inputs.

src/KubernetesClient/PeekableStreamReader.cs
```diff
@@ -0,0 +1,118 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace k8s
+{
+    public class PeekableStreamReader : IAsyncLineReader
```
DavidParks8: This implementation is not thread safe. Please ensure it works properly when the various methods are called at the same time on separate threads.

src/KubernetesClient/WatcherDelegatingHandler.cs
blushingpenguin: That makes it hard to unit test code that uses the Watch API.  With it exposed it's sufficient to do something like:

```csharp
mocks.k8s.ListNamespacedPodWithHttpMessagesAsync("test")
.Returns(Task.FromResult(new HttpOperationResponse<V1PodList>
{
    Body = new V1PodList(),
    Response = new HttpResponseMessage { Content = mockLineSeparatedHttpContent }
}));
```

Would putting it in an internal namespace (k8s.internal) suffice?  (asp.net core seems to use this pattern)

DavidParks8: You can use InternalsVisibleTo attribute to remove the issue entirely. I would check with the main code owners to see what their preference is.

blushingpenguin: Sorry, I didn't mean for unit testing the KubernetesClient library itself, I meant for unit testing code that uses the KubernetesClient library.

DavidParks8: Ah, still, is this something people should be testing? It looks like it is deep enough in the code that people might be better off mocking a higher level abstraction.

src/KubernetesClient/PeekableStreamReader.cs
blushingpenguin: I'm not sure why it needs to be thread safe?  The underlying stream that it ends up wrapped around from HttpClient (IIRC ChunkedEncodingReadStream) isn't thread safe and nor was the previous implementation of the same class.

src/KubernetesClient/WatcherDelegatingHandler.cs
blushingpenguin: If you call watch, there's nothing you can mock -- you could e.g. write a thin wrapper:

```csharp
interface IWatcher 
{
  Task Watch<T>(HttpOperationResponse response, Action<> onEvent, ...)
}

class Watcher : IWatcher
{
  Task Watch<T>(HttpOperationResponse response, Action<> onEvent, ...) 
  {
    return response.Watch(onEvent, ...)
  }
}
```

would it be worth including something like that as an alternative to calling the extension function?

src/KubernetesClient/PeekableStreamReader.cs
blushingpenguin: done

src/KubernetesClient/PeekableStreamReader.cs
blushingpenguin: I have updated this to read:

```csharp
throw new InvalidOperationException(
    "Reading from the stream failed because a line from the server was unexpectedly " +
    $"greater than the configured maximum line length of {_maxLineLength}");
```

src/KubernetesClient/PeekableStreamReader.cs
blushingpenguin: done

src/KubernetesClient/PeekableStreamReader.cs
blushingpenguin: done

src/KubernetesClient/Watcher.cs
blushingpenguin: I have moved cancellationToken to be the last argument and defaulted it to default(CancellationToken) as is typical for other .NET core apis.

src/KubernetesClient/PeekableStreamReader.cs
blushingpenguin: I have changed this to read:

```csharp
throw new InvalidOperationException(
    "Reading from the stream failed because a line from the server was unexpectedly " +
    $"greater than the configured maximum line length of {_maxLineLength}");
```

brendandburns:
This looks reasonable to me. I filed a much simpler PR to also add async support to watching. I think we should probably merge both. The HttpOperationResponse itself takes a while to return and thus, even with this async impl. it still turns basically into a sync call for the client...

tg123:
close via #313

PR #302, opened 2019-09-08
Add another couple of constructors to Fix #279 

### LINKED ISSUE #279: Allow to set HttpClient when creating the client
Currently, there is overload that let you set the HttpClient (for example, if you want to override the default timeout) and use `KubernetesClientConfiguration`. There is such a constructor on the generated `Kubernetes` object - but it doesn't accept configuration.

DavidParks8:
This change deduplicates some of the code of the Kubernetes class, and adds the ability to use KubernetesClientConfiguration with the http client factory pattern.

A sample project was added to show how this pattern might be used.

### DIFF
--- a/src/KubernetesClient/Kubernetes.ConfigInit.cs
+++ b/src/KubernetesClient/Kubernetes.ConfigInit.cs
@@ -13,84 +13,68 @@ namespace k8s
 {
     public partial class Kubernetes
     {
-#if MONOANDROID8_1
         /// <summary>
         ///     Initializes a new instance of the <see cref="Kubernetes" /> class.
         /// </summary>
         /// <param name='config'>
-        ///     Optional. The delegating handlers to add to the http client pipeline.
+        ///     The kube config to use.
         /// </param>
-        /// <param name="handlers">
-        ///     Optional. The delegating handlers to add to the http client pipeline.
+        /// <param name="httpClient">
+        ///     The <see cref="HttpClient" /> to use for all requests.
         /// </param>
-        public Kubernetes(KubernetesClientConfiguration config, params DelegatingHandler[] handlers) : this(new Xamarin.Android.Net.AndroidClientHandler(), handlers)
+        public Kubernetes(KubernetesClientConfiguration config, HttpClient httpClient) : this(config, httpClient, false)
         {
-            if (string.IsNullOrWhiteSpace(config.Host))
-            {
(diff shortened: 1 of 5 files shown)


DavidParks8:
Signed the CLA

brendandburns:
Thanks for the PR! The diffs here make me nervous that we're introducing a regression (and I'm not 100% how perfect our coverage is here...)

Any chance you can refactor the PR to minimize diffs? 

It looks like you included some cleanups along with new code. Any chance you can break them into two commits (or two PRs) so that I can review the cleanup and the new code separately?

Thanks!
(and sorry for the delay in reviewing...)

DavidParks8:
When implementing this, I faced the issue of either deduplicating, or copying the new constructors for each isolated platform. I chose to deduplicate (either way, the changeset would be large unless this PR didn't include parity across platforms). Since the major deduplication happened for the mono platform, I went ahead and dereferenced the new code into what the existing constructor would look like if the functions were inlined, then did a diff with the constructor from before this change. This resulted in the following changes: https://www.diffchecker.com/h24E7Ct7

Is that sufficient to show the relative risk?

src/KubernetesClient/Kubernetes.ConfigInit.cs
```diff
@@ -104,18 +88,18 @@ public Kubernetes(KubernetesClientConfiguration config, params DelegatingHandler
             {
                 throw new KubeConfigException("Bad host url", e);
             }
+        }
 
-            CaCerts = config.SslCaCerts;
```
brendandburns: Looks to me like this code was dropped? (I could be wrong...)

brendandburns: Oh, is it in the constructor above? I would just move all of that down here....

brendandburns:
Ok, I'm more comfortable with this now, one note about moving where CaCerts and IsTLSVerify is initialized and I think we're good to merge.

Thanks for the patience.

src/KubernetesClient/Kubernetes.ConfigInit.cs
DavidParks8: I would, but doing so won't compile because CaCerts only has an empty getter that needs to be assigned where it is declared, or directly within a constructor.

DavidParks8:
Thanks!

PR #308, opened 2019-09-27
Add a WatchAsync method.

### LINKED ISSUE #301: Watcher creation blocks until first change notification available
I would like to be able to instantiate a new Watcher for a resource before any instances of the resource are created.  For example, I would like to create a watcher that is watching for Deployments before any deployments have been created.  I expect construction of the Watcher to complete immediately regardless of whether any resources exist.  Then, I expect events to be dispatched on the Watcher's background thread.  If no resources of the type exist when the Watcher is created, I would expect the first event dispatched by the background thread to be an "Add" event when a new instance is created.  

Instead, it is not possible to create a new Watcher without blocking until the first instance of a resource is available.

Creation of a new Watcher requires first retrieving an HTTPOperationResponse for the given resource (ex. by calling `ListNamespacedPodWithHttpMessagesAsync`).  These methods block until they receive an initial response; however, when watching is enabled, there is no initial response (other than the HTTP Headers) until a resource is created.

Steps to Reproduce:
1) Run the "watch" example application with a Kubernetes cluster that does not have any deployments.
2) Observe that the console write that occurs after the watch has been created ("press ctrl + c ...") does not appear in the console
3) Only after creating a deployment or pod will the console output line appear

brendandburns:
Closes #301

### DIFF
--- a/examples/watch/Program.cs
+++ b/examples/watch/Program.cs
@@ -13,21 +13,21 @@ namespace watch
 
             IKubernetes client = new Kubernetes(config);
 
-            var podlistResp = client.ListNamespacedPodWithHttpMessagesAsync("default", watch: true).Result;
-            using (podlistResp.Watch<V1Pod>((type, item) =>
+            var podlistResp = client.ListNamespacedPodWithHttpMessagesAsync("tester", watch: true);
+            var watchTask = podlistResp.WatchAsync<V1Pod, V1PodList>((type, item) =>
             {
                 Console.WriteLine("==on watch event==");
                 Console.WriteLine(type);
                 Console.WriteLine(item.Metadata.Name);
                 Console.WriteLine("==on watch event==");
-            }))
-            {
-                Console.WriteLine("press ctrl + c to stop watching");
+            });
+            Console.WriteLine("press ctrl + c to stop watching");
 
-                var ctrlc = new ManualResetEventSlim(false);
-                Console.CancelKeyPress += (sender, eventArgs) => ctrlc.Set();
-                ctrlc.Wait();
-            }
+            var ctrlc = new ManualResetEventSlim(false);
+            Console.CancelKeyPress += (sender, eventArgs) => ctrlc.Set();
+            ctrlc.Wait();
+
+            watchTask.Result.Dispose();
         }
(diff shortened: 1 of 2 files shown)


examples/watch/Program.cs
```diff
@@ -13,8 +13,8 @@ private static void Main(string[] args)
 
             IKubernetes client = new Kubernetes(config);
 
-            var podlistResp = client.ListNamespacedPodWithHttpMessagesAsync("default", watch: true).Result;
-            using (podlistResp.Watch<V1Pod>((type, item) =>
+            var podlistResp = client.ListNamespacedPodWithHttpMessagesAsync("default", watch: true);
+            using (podlistResp.WatchAsync<V1Pod, V1PodList>((type, item) =>
```
tg123: do you want 
```
using (await ...)
```

here using is to dispose Task

tg123:
```suggestion
            using (await podlistResp.WatchAsync<V1Pod, V1PodList>((type, item) =>
```

brendandburns: I don't think so, because the whole goal was to print the 'ctl-c to cancel' before the wait completes, but perhaps I should drop the 'using' part..

tg123: `await WatchAsync` creates a new Watcher which will not block running to code inside `using`

brendandburns: fwiw, you can't use await here because then the Main would have to be async and dotnet doesn't like that. I tried using `.Result` but it hangs w/o printing the `ctrl-c` stuff.

I'm open to suggestions, but for now I think this is the best we can do (or we could just remove the `using` syntax...)

brendandburns: Actually, I just rewrote the code to avoid the `using` directive and to call `Dispose` directly.

brendandburns:
Comment addressed, please take another look.

examples/watch/Program.cs
tg123: sorry I finally understand the issue I did not look into #301 carefully before
please see #313 for my proposal

tg123:
close via #313

PR #309, opened 2019-09-27
Clean up and fix port forwarding example.

### LINKED ISSUE #297: Port Forward example not working
I am using Port forward example where I need to forward the port of running service. I have service running inside pod on port 3122. 
When I redirect it to 8080. The url is localhost:8080 and has UI but I don't see full UI. Only I can see the service name mentioned in Title bar of browser.

Debug:
```
if (bytesRec == 0 || Encoding.ASCII.GetString(bytes,0,bytesRec).IndexOf("<EOF>") > -1) {  
                            break;  
}
```
Coming always false `Encoding.ASCII.GetString(bytes,0,bytesRec).IndexOf("<EOF>") > -1`

But if I execute same using kubectl command then it works perfectly fine. 
`kubectl port-forward service/dbdashboard 8080:3122 -n my_namespace`

**Note:** k8s cluster is not installed on my system. I am using k8s cluster which is installed on Azure cloud.

brendandburns:
Fixes #297

### DIFF
--- a/examples/portforward/PortForward.cs
+++ b/examples/portforward/PortForward.cs
@@ -24,7 +24,7 @@ namespace portforward
         private async static Task Forward(IKubernetes client, V1Pod pod) {
             // Note this is single-threaded, it won't handle concurrent requests well...
             var webSocket = await client.WebSocketNamespacedPodPortForwardAsync(pod.Metadata.Name, "default", new int[] {80}, "v4.channel.k8s.io");
-            var demux = new StreamDemuxer(webSocket);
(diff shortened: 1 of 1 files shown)


examples/portforward/PortForward.cs
```diff
@@ -61,6 +62,10 @@ private static async Task Main(string[] args)
 
             await accept;
             await copy;
+            if (handler != null) {
```
tg123: handler?.Close()

tg123:
```suggestion
            handler?.Close();            
```

tg123:
/LGTM 
with some suggestion

PR #312, opened 2019-10-02
gen v1.16.0

tg123:


brendandburns:
Generally LGTM, but I think we should increment the version in `version.json`

Thanks!

PR #313, opened 2019-10-02
watcher to support empty response

tg123:
my proposal for #308 
move the async support into watcher to keep the `semantic meaning`  of watcher
tested on empty ns

### DIFF
--- a/examples/watch/Program.cs
+++ b/examples/watch/Program.cs
@@ -13,7 +13,7 @@ namespace watch
 
             IKubernetes client = new Kubernetes(config);
 
-            var podlistResp = client.ListNamespacedPodWithHttpMessagesAsync("tester", watch: true);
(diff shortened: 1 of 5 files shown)


examples/watch/Program.cs
```diff
@@ -13,8 +13,8 @@ private static void Main(string[] args)
 
             IKubernetes client = new Kubernetes(config);
 
-            var podlistResp = client.ListNamespacedPodWithHttpMessagesAsync("default", watch: true).Result;
-            using (podlistResp.Watch<V1Pod>((type, item) =>
+            var podlistResp = client.ListNamespacedPodWithHttpMessagesAsync("tester", watch: true);
```
brendandburns: Keep this as "default" I changed it for my cluster, but I should have reverted it...

brendandburns:
LGTM, but CI is failing and there's one small change.

PR #326, opened 2019-11-25
Changing the patch type for custom objects from "merge" to "json"

### LINKED ISSUE #305: Why do the Patch*CustomObject* methods use Merge Patch headers?
Hi folks,

We've been trying to use the `PatchNamespacedCustomObject` method to modify some fields in a CRD instance. This throws an error, as we're sending a `V1Patch` but these methods set the header to `merge-patch+json`, thus k8s can't unmarshal it at the other end. All the other Patch methods we can see use `json-patch+json`.

What is the rationale behind having these methods send a different header? The `V1Patch` class doesn't seem to support merge patches at this time, and while we can certainly work out how to form the JSON ourselves it seems like we're missing something here.

Cheers!

vinisoto:
To fix #305 and to be consistent with other Patch operations.


vinisoto:
Will do more investigation and re-submit if needed

PR #334, opened 2019-12-14
Gen v1.17.0

tg123:


tg123:
gitversion should work with dotnet 3, still looking into it

brendandburns:
I think I want to hold off on merging this because in theory 1.17 clients aren't required to be compatible with 1.14 clusters and I think at this point 1.14 clusters are more common than 1.17 clusters...

PR #336, opened 2019-12-19
Prevent NullReferenceException when using HttpClient to initialize Kubernetes client. Fixes #335

### LINKED ISSUE #335: NullReferenceException thrown by Kubernetes.InitializeFromConfig when initialized with HttpClient
When creating a new client instance by passing in an `HttpClient`, a `NullReferenceException` is thrown by the `InitializeFromConfig` method because the `HttpClientHandler` property is null. This can be demonstrated by attempting to run the httpClientFactory sample in the `examples` folder.

Full stack trace:
```
Unhandled Exception: System.NullReferenceException: Object reference not set to an instance of an object.
   at k8s.Kubernetes.InitializeFromConfig(KubernetesClientConfiguration config) in C:\dev\github\third_party\csharp\src\KubernetesClient\Kubernetes.ConfigInit.cs:line 108
   at k8s.Kubernetes..ctor(KubernetesClientConfiguration config, HttpClient httpClient, Boolean disposeHttpClient) in C:\dev\github\third_party\csharp\src\KubernetesClient\Kubernetes.ConfigInit.cs:line 46        at k8s.Kubernetes..ctor(KubernetesClientConfiguration config, HttpClient httpClient) in C:\dev\github\third_party\csharp\src\KubernetesClient\Kubernetes.ConfigInit.cs:line 25
   at httpClientFactory.Program.<>c.<Main>b__0_2(HttpClient httpClient, IServiceProvider serviceProvider) in C:\dev\github\third_party\csharp\examples\httpClientFactory\Program.cs:line 31
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitTransient(TransientCallSite transientCallSite, ServiceProviderEngineScope scope)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitConstructor(ConstructorCallSite constructorCallSite, ServiceProviderEngineScope scope)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitTransient(TransientCallSite transientCallSite, ServiceProviderEngineScope scope)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitIEnumerable(IEnumerableCallSite enumerableCallSite, ServiceProviderEngineScope scope)
   at Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.GetService[T](IServiceProvider provider)   at Microsoft.Extensions.Hosting.Internal.Host.StartAsync(CancellationToken cancellationToken)
   at httpClientFactory.Program.Main(String[] args) in C:\dev\github\third_party\csharp\examples\httpClientFactory\Program.cs:line 41
   at httpClientFactory.Program.<Main>(String[] args)
```

This is likely related to #307, where it was suggested that a fix to the AutoRest library might be necessary. After digging a bit deeper this afternoon, I think this may be an issue that does need to be worked through in the k8s client code. When using .NET Core's `HttpClientFactory` pattern, by the time you get an `HttpClient` instance, it's already been initialized with its `HttpClientHandler`, and there's no (good) way for AutoRest to get this back out. I believe that's the reason the code linked in #307 for `ServiceClient` doesn't already set the `HttpClientHandler` property when passed an already existing `HttpClient`.

The only two solutions I can think of are to either pass an already-configured `HttpClient` into the Kubernetes client constructor and not run the offending pieces of `InitializeFromConfig` in that scenario, or pass in both the `HttpClient` *and whatever `HttpClientHandler` was used to create it*. The latter feels a bit wonky, so I've worked up a solution that implements the former, since it plays nicer with the HttpClientFactory implementation that .NET Core provides. Once I've had time to vet a few additional scenarios (i.e., testing with frameworks other than .NET Core), I'll submit my changes as a PR for review.

BrandonSharp:
This should correct the NullReferenceException that occurs when passing in an HttpClient to the constructor of the Kubernetes client. This occurred because HttpClient objects are already initialized with an HttpClientHandler before being handed down to the Kubernetes client, and there's no way to retrieve that object from an HttpClient once initialized.

Because it's useful to set some values on an HttpClientHandler based on the Kubernetes configuration, I added a new file to the partial KubernetesClientConfiguration class that handles generation of a default HttpClientHandler based on the config. This replaced the need to call `InitializeFromConfig` in the Kubernetes.ConfigInit.cs class from the constructor that accepts an HttpClient. I've also updated the example named `httpClientFactory` to illustrate the use of this new function.

I also split the `SetCredentials` method in Kubernetes.ConfgInit.cs into two separate functions. Originally, this method was both setting credentials, and adding certificates to the HttpClientHandler. Because of the changes required to resolve this bug, it was necessary to pull the certificate related code out, and make it more generally usable, given an HttpClientHandler and a config object.

### DIFF
--- a/examples/httpClientFactory/Program.cs
+++ b/examples/httpClientFactory/Program.cs
@@ -31,7 +31,8 @@ namespace httpClientFactory
                             return new Kubernetes(
                                 serviceProvider.GetRequiredService<KubernetesClientConfiguration>(),
                                 httpClient);
-                        });
+                        })
+                        .ConfigurePrimaryHttpMessageHandler(config.CreateDefaultHttpClientHandler);
 
                     // Add the class that uses the client
                     services.AddHostedService<PodListHostedService>();
--- a/src/KubernetesClient/Kubernetes.ConfigInit.cs
+++ b/src/KubernetesClient/Kubernetes.ConfigInit.cs
@@ -43,7 +43,7 @@ namespace k8s
             ValidateConfig(config);
             CaCerts = config.SslCaCerts;
             SkipTlsVerify = config.SkipTlsVerify;
-            InitializeFromConfig(config);
+            SetCredentials(config); 
         }
 
         /// <summary>
@@ -151,8 +151,9 @@ namespace k8s
                 }
             }
 
-            // set credentails for the kubernernet client
-            SetCredentials(config, HttpClientHandler);
+            // set credentails for the kubernetes client
+            SetCredentials(config);
+            config.AddCertificates(HttpClientHandler);
         }
 
         private X509Certificate2Collection CaCerts { get; }
@@ -195,9 +196,7 @@ namespace k8s
         ///     Set credentials for the Client
         /// </summary>
         /// <param name="config">k8s client configuration</param>
-        /// <param name="handler">http client handler for the rest client</param>
-        /// <returns>Task</returns>
-        private void SetCredentials(KubernetesClientConfiguration config, HttpClientHandler handler)
+        private void SetCredentials(KubernetesClientConfiguration config)
         {
             // set the Credentails for token based auth
             if (!string.IsNullOrWhiteSpace(config.AccessToken))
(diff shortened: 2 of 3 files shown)


BrandonSharp:
Rebased my fork to pick up the recent changes in master that seems to have addressed some of the failing tests.

brendandburns:
Sorry for the delay here, holidays got in the way, I will try to review in the next day or so.

Thanks for the patience.

PR #337, opened 2019-12-20
fix gh action failed

tg123:
Nerdbank.GitVersioning cannot read the obj id of master due to depth is only 1


tg123:
other pending PR should rebase this to get a success build

PR #342, opened 2019-12-31
Use `Path.DirectorySeparatorChar` to have ServiceAccountPath start from root

darobs:
The path constructed in `KubernetesClientConfiguration::ServiceAccountPath` is a relative path, and this works fine as long as CWD is the root directory.

This PR uses `Path.DirectorySeparatorChar` to have `ServiceAccountPath` start from root, instead of a relative path.

This should work fine on Linux containers, but it still assumes the Windows container working drive is "C:"

### DIFF
--- a/src/KubernetesClient/KubernetesClientConfiguration.InCluster.cs
+++ b/src/KubernetesClient/KubernetesClientConfiguration.InCluster.cs
@@ -8,7 +8,7 @@ namespace k8s
     {
         private static string ServiceAccountPath =
             Path.Combine(new string[] {
-                "var", "run", "secrets", "kubernetes.io", "serviceaccount/"
+                $"{Path.DirectorySeparatorChar}var", "run", "secrets", "kubernetes.io", "serviceaccount"
             });
         private const string ServiceAccountTokenKeyFileName = "token";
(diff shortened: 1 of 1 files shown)


tg123:
do you mean `c:/var/xxxx` ?

darobs:
> do you mean `c:/var/xxxx` ?

In Windows, this pathname resolves to `\var\run\secrets\kubernetes.io\serviceaccount`, which will resolve to the root of the current working drive.

Example:
```
PS C:\temp> type a.txt
Hello world!
PS C:\temp> type \temp\a.txt
Hello world!
PS C:\temp> d:
PS D:\Windows> type \temp\a.txt
type : Cannot find path 'D:\temp\a.txt' because it does not exist.
```

So this PR works fine if the current working directory in the Windows container is on the same drive as `\var\run\secrets\kubernetes.io` is mounted.  It fixes the problem for Linux.

tg123:
@darobs 
is there any thing in `C:\var\run\secrets\kubernetes.io\serviceaccount`?
those are inject by system, I am not sure if it still there on Windows

IHMO, we need to find out a Windows way to do so.

brendandburns:
/lgtm
/approve

Regardless of windows or not this is an improvement on the current code.

PR #345, opened 2020-01-09
Rev dotnet to 3.1

brendandburns:
cc @tg123

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -3,7 +3,7 @@ jobs:
     runs-on: ubuntu-18.04
     strategy:
       matrix:
-        dotnet: [ '2.2.103', '3.0.100' ]
+        dotnet: [ '3.1.100' ]
     name: Dotnet ${{ matrix.dotnet }} sample
     steps:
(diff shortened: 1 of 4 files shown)


tg123:
IMHO, we should not upgrade to target `netcore31`

currently, the source code are targeting

- [X] netstandard1.4;
- [X] netstandard2.0;
- [X] net452;
- [X] net461;
- [X] netcoreapp2.1;
- [X] xamarinios10;
- [X] monoandroid81

but on nuget.org the published pkg is target below only

- [X] netstandard1.4;
- [X] netstandard2.0;
- [ ] net452;
- [ ] net461;
- [X] netcoreapp2.1;
- [ ] xamarinios10;
- [ ] monoandroid81


Most of them are overlap
e.g. code targeting `netstard1.4` could run on most of those platform including `xamarinios10` and `monoandroid81`.

multi targeting was introduced for websocket and certificates related api

As `netstand2.1` was released for a while, I think it is better to clean up and targeting and unify the platforms.

kkmsft:
Can you please update the ones under examples as well ?

PR #347, opened 2020-01-13
[WIP] Client cache +  informers - phase1

kkmsft:
This PR implements the local cache and generic informer infrastructure. The implementation mirrors the client-go implementation. 

More details on the design and implementation notes and TODO can be found in the README.md attached with the PR.  Each commit constitutes a modular piece of the implementation. 

Example usage is present under example/informers directory with the PR.

Note: WIP.


src/KubernetesClient/Cache/Store.cs
```diff
@@ -0,0 +1,88 @@
...
+        public string MetaNamespaceKeyFunc(IKubernetesObject o) 
+        {
+            try {
+                 var typedObject = (V1ObjectMeta)o.GetType().GetProperty("Metadata").GetValue(o, null);  
+                if (typedObject.NamespaceProperty == "") {
+                    return typedObject.Name;                
+                }
+                return typedObject.NamespaceProperty+"/"+typedObject.Name;
+            } catch (Exception ex) {
+                Console.WriteLine(ex);
```
brendandburns: Do we really want to swallow the exception here?

src/KubernetesClient/Cache/Store.cs
```diff
@@ -0,0 +1,88 @@
...
+                return "";
+            }            
+        }
+
+        public Store() {
+            _threadSafeStorage = new ThreadSafeStore();
+            _metaNamespaceFunc = MetaNamespaceKeyFunc;
+        }
+        public void Add(IKubernetesObject o)
```
brendandburns: Please clean up formatting, here and elsewhere.

src/KubernetesClient/Cache/Store.cs
```diff
@@ -0,0 +1,88 @@
...
+        {            
+            var key = _metaNamespaceFunc(o);            
+            _threadSafeStorage.Add(key, o);            
+        }
+
+        public Tuple<IKubernetesObject, bool> Get(IKubernetesObject o)
```
brendandburns: Is a `Tuple` really needed here? I know that's what Go does, but it seems like null and non-null are sufficient in this code.

src/KubernetesClient/Cache/Store.cs
```diff
@@ -0,0 +1,88 @@
...
+        {
+            var key = _metaNamespaceFunc(o);
+            var val = _threadSafeStorage.Get(key);
+            if (val == null) {
+                return new Tuple<IKubernetesObject, bool>(null, false);
+            }
+            return new Tuple<IKubernetesObject, bool>(val, true);
+        }
+
+        public Tuple<IKubernetesObject, bool> GetByKey(string key)
```
brendandburns: as above wrt `Tuple`

src/KubernetesClient/Cache/ThreadSafeStore.cs
```diff
@@ -0,0 +1,88 @@
...
+            _storageLock.WaitOne();           
+            try {                
+                if (_store.ContainsKey(key))
+                {
+                    _store[key] = value;
+                } else {
+                    _store.Add(key, value);            
+                }
+            } catch (Exception ex) {
+                Console.WriteLine(ex);
```
brendandburns: Don't swallow exceptions.

src/KubernetesClient/Cache/ThreadSafeStore.cs
```diff
@@ -0,0 +1,88 @@
+using System;
+using System.Threading;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace k8s.cache
+{
+    public class ThreadSafeStore : IThreadSafeStore
+    {
+        private static Mutex _storageLock;
```
brendandburns: Why not just use `ConcurrentDictionary`?

src/KubernetesClient/Informers/Controller.cs
```diff
@@ -0,0 +1,122 @@
...
+
+namespace k8s.informers
+{
+    public class Controller<T, L> : IController<T, L>
+    {
+        public IReflector<T,L> _reflector;       
+        public IStore _store;
+        private Queue<Tuple<WatchEventType, IKubernetesObject>> _fifo;
+        private AutoResetEvent _fifoEvent;
+        private Action<WatchEventType, IKubernetesObject> _onAdd;
```
brendandburns: Shouldn't these be `List<Action<...>>` ?

src/KubernetesClient/Informers/Controller.cs
```diff
@@ -0,0 +1,122 @@
...
+            _fifo = new Queue<Tuple<WatchEventType, IKubernetesObject>>();
+            _fifoEvent = new AutoResetEvent(false);
+            _reflector = new Reflector<T, L>(lw, TimeSpan.FromMinutes(1), _fifo, _fifoEvent);
+            
+        }
+
+        /// <summary>
+        /// Installs resource handler call backs for add, del and update of objects.
+        /// </summary>
+        public void AddResourceHandlers(Action<WatchEventType, IKubernetesObject> onAdd, Action<WatchEventType, IKubernetesObject> onDelete,
```
brendandburns: Per above "AddResourceHandler" implies that it would be added, when in fact it replaces an existing handler.

src/KubernetesClient/Informers/Controller.cs
```diff
@@ -0,0 +1,122 @@
...
+                        }
+                        // Ensure that we don't burn the CPU while trying to process the Q contents.
+                       Thread.Sleep(0);
+                    }                               
+                }
+                if (cancellationToken.IsCancellationRequested) {
+                    Console.WriteLine("Controller exiting due to cancel!!");
+                }
+            } catch (Exception ex) {
+                Console.WriteLine(ex);
```
brendandburns: Remove log statement here.

src/KubernetesClient/Informers/ListerWatcher.cs
```diff
@@ -0,0 +1,48 @@
...
+using System;
+
+namespace k8s.informers
+{    
+    public delegate Task<HttpOperationResponse<L>> ListerD<T, L>();
+    public delegate Watcher<T> WatcherD<T, L>(Action<WatchEventType, T> onEvent, Action<Exception> onError = null, Action onClosed = null);
+
+    public class ListerWatcher<T, L>
+    {           
+        public ListerD<T, L> Lister;
```
brendandburns: Why are these public?

src/KubernetesClient/Informers/Reflector.cs
```diff
@@ -0,0 +1,75 @@
...
+                            while(!c.IsCancellationRequested) {                                
+                                Thread.Sleep(0);
+                            }
+                            if (c.IsCancellationRequested) {
+                                Console.WriteLine("Cancellation requested. Going out of reflector work loop.");
+                            }
+                }                
+            
+            } catch (Exception ex) {
+                    Console.WriteLine(ex);
```
brendandburns: Remove logging

src/KubernetesClient/Informers/SharedInformer.cs
```diff
@@ -0,0 +1,59 @@
...
+            _controller = new Controller<T, L>(lw, _store);
+
+        }
+
+        /// <summary>
+        /// Installs resource handler call backs for add, del and update of objects.
+        /// </summary>        
+        public void AddResourceHandlers(Action<WatchEventType, IKubernetesObject> onAdd=null, Action<WatchEventType, IKubernetesObject> onDelete=null, Action<WatchEventType, IKubernetesObject, IKubernetesObject> onUpdate=null)
+        {
+            // TODO: Raise a user friendly exception.
```
brendandburns: Please do this TODO :)

src/KubernetesClient/Kubernetes.List.cs
```diff
@@ -0,0 +1,112 @@
...
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+
+
+namespace k8s
+{
+    public partial class Kubernetes
+    {        
+        public async Task<HttpOperationResponse<L>> List<T, L>(string namespaceParameter, CancellationToken cancellationToken=default(CancellationToken), bool? watch = default(bool?)) {
```
brendandburns: Why do you need this? The List methods are already implemented in the client.

kkmsft: Could not find any List method which are generic- takes Type and ListType and gets back a objects of ListType. All of them looked like they are specific - as in uses the Type in the function name and repeats the same code. Am I missing something here ?

brendandburns: What I have done elsewhere is pass the List function as a `Func<Type, ListType>` object rather than implementing a generic function like this.

macsux:
Is there a reason this implementation was based on the Go version instead of Java one? The languages are much more similar to Java then Go, and Java version provides a higher level abstraction.

gen/KubernetesWatchGenerator/ModelExtensions.cs.template
```diff
@@ -3,9 +3,12 @@ namespace k8s.Models
 {{#.}}
     public partial class {{GetClassName . }} : IKubernetesObject
     {
-        public const string KubeApiVersion = "{{GetApiVersion . }}";
-        public const string KubeKind = "{{GetKind . }}";
-        public const string KubeGroup = "{{GetGroup . }}";
+        private const string kubeApiVersion = "{{GetApiVersion . }}";	
+        private const string kubeKind = "{{GetKind . }}";
+        private const string kubeGroup = "{{GetGroup . }}";	
+        public string KubeApiVersion { get { return kubeApiVersion;} }
```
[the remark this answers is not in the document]
kkmsft: In-order to implement a generic lister, we need a way to find the ApiVersion and Kind from a given Kubernetes object. Currently, we don't have a way to do that (even when using reflection) because consts are treated as static members and hence not available from the object instances. Hence had to convert this to get to gather this from a instance of the Kubernetes object.

src/KubernetesClient/Cache/ThreadSafeStore.cs
kkmsft: Originally, wanted to do multiple operations under the lock hence choose the mutex approach. But looking at now, should be possible to use ConcurrentDictionary. Will do an iteration with ConcurrentDictionary.

src/KubernetesClient/Cache/ThreadSafeStore.cs
kkmsft: Yep, moving out of WIP will involve that and tests at every level.

src/KubernetesClient/Informers/Controller.cs
kkmsft: As in multiple Actions in case of add, delete and update ? Sure, can accommodate that requirement.

src/KubernetesClient/Informers/ListerWatcher.cs
kkmsft: Reflector code calls the ListWatcher's List and Watcher, hence they are kept as public.

src/KubernetesClient/Cache/IStore.cs
```diff
@@ -0,0 +1,18 @@
...
+using System.Collections.Generic;
+
+namespace k8s.cache
+{
+    public interface IStore {
+        void Add(IKubernetesObject o);
+        void Delete(IKubernetesObject o);
+        IKubernetesObject Update(IKubernetesObject o);
+        List<IKubernetesObject> List();
+        List<String> ListKeys();
```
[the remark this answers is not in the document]
kkmsft: Thank you, @tg123

brendandburns:
I sort of second the comment around the Java implementation. (though the Java implementation was definitely based on the Go one too)

I don't think we should be slavishly devoted to the Go approach, we should re-use concepts where appropriate, but feel free to modify things that make more sense in a oo language like C# or Java.

kkmsft:
Could @macsux , @brendandburns provide some specific examples in the WIP PR where it could be made more java style ?

src/KubernetesClient/Cache/IStore.cs
```diff
@@ -0,0 +1,18 @@
...
+        Tuple<IKubernetesObject, bool> Get(IKubernetesObject o);
```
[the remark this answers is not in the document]
kkmsft: Will be removing tuples shortly.

tests/KubernetesClient.Tests/ThreadSafeStoreTests.cs
```diff
@@ -0,0 +1,140 @@
...
+            }
+            Task.WaitAll(tasks);
+
+
+            int count = 0;
+            foreach (string key in store.ListKeys())
+            {
+                // Ensure we have not seen the key before.
+                Assert.False(dict[key] == true);
+                dict[key] = true;
```
tg123: missing assert every single key is true in case of store.listkeys did not return everything.

just a notice, i did see `assert(count, max)` not quite sure if checking count only is good enough

tests/KubernetesClient.Tests/ThreadSafeStoreTests.cs
```diff
@@ -0,0 +1,140 @@
...
+
+            val = store.Get(key);
+            Assert.Null(val);
+        }
+
+        [Fact]
+        public void MultiThread()
+        {
+            var store = new ThreadSafeStore();
+            Action<object> action = (object o) =>
```
tg123:
```suggestion
            var action = (int o) =>
```

the caller calls using int as  param

tests/KubernetesClient.Tests/ThreadSafeStoreTests.cs
```diff
@@ -0,0 +1,140 @@
...
+
+            raisedExcection = false;
+            // Delete again 
+            try
+            {
+                store.Delete(key);
+            }
+            catch (Exception)
+            {
+                raisedExcection = true;
```
tg123: did not check this var

want asset exception?

src/KubernetesClient/Cache/ThreadSafeStore.cs
```diff
@@ -0,0 +1,69 @@
...
+        public ThreadSafeStore()
+        {
+            _store = new ConcurrentDictionary<string, IKubernetesObject>();
+        }
+
+        public void Add(string key, IKubernetesObject value)
+        {
+            if (!_store.TryAdd(key, value))
+            {
+                throw new Exception(String.Format("Could not add value for key: {0}", key));
```
tg123: same as my comment on interface design

src/KubernetesClient/Cache/ThreadSafeStore.cs
```diff
@@ -0,0 +1,69 @@
...
+            }
+        }
+
+        public void Delete(string key)
+        {
+            IKubernetesObject oldVal;
+            if (!_store.TryRemove(key, out oldVal))
+            {
+                // Could not remove the value. Raise an exception.
+                throw new Exception(String.Format("Could not remove value with key: {0}", key));
```
tg123: would be `del` and `add` happens at the same time

examples/informers/Program.cs
```diff
@@ -0,0 +1,72 @@
...
+     
+            var config = KubernetesClientConfiguration.BuildConfigFromConfigFile(Environment.GetEnvironmentVariable("KUBECONFIG"));
+
+            IKubernetes client = new Kubernetes(config);
+            var lw = new ListerWatcher<V1Pod, V1PodList>(client);
+
+            ISharedInformer<V1Pod, V1PodList> informer = new SharedInformer<V1Pod, V1PodList>(lw);
+            var source = new CancellationTokenSource();
+            var cancellationToken = source.Token;
+            informer.Run(cancellationToken);
```
tg123: do we need to add a IDisposable to IInformer ?

macsux:
Cross-referencing another PR that is a Java port. It's mostly complete minus unit tests that I'm still working on porting.
https://github.com/kubernetes-client/csharp/pull/351

macsux:
I propose this PR is closed and continued on top of https://github.com/kubernetes-client/csharp/pull/351
The other PR is essentially full implementation of Informers and Cache based on the Java version. It is significantly further ahead then this PR (I've been working on it for a while, just haven't shared it until it was covered with tests).

Implementation is basically complete, it just needs to be covered with tests which I'm porting from Java. @kkmsft I would love help covering it with tests if you got cycles.

kkmsft:
Closing this to continue on https://github.com/kubernetes-client/csharp/pull/351.
@macsux - please let me know the tests you have not yet started. I will raise PRs against your repo 351 to complete the PR.

kkmsft:
Thank you @tg123 & @brendanburns for the comments on the WIP PRs.

PR #351, opened 2020-01-27
[WIP] Informers - Java port

macsux:
This is an implementation of the k8s informers and related objects. It is a port of Java implementation with notable differences:
- all blocking calls have been replaced by async methods
- work queues are based on .NET System.Threading.Channels to do efficient async queues
- tasks are scheduled on default task pool instead of dedicated pools 
- DeltaFifo provides explicit Read / Write type separation. Java/Go uses marker objects to create certain conditions, therefore returns were always `Object`. This implementation improves by wrapping those in explicit envelope object reducing the need for casting and type checks
- Watch has been extended with conversion to IObservable interface
- Many services expose start/stop semantics that have been modeled on IHostedService
- Test assertion syntax is greatly improved with the use of FluentAssertions. 
- Introduce the use of NSubstitute instead of Moq for test mocks, greatly improving readability (syntax does not require the use of setup lambdas). If the maintainers are OK with the use of the above libraries I can backport them on to existing tests in subsequent PRs.

Note that due to dependency on IHostedService and Channels, the code targets netstandard2.0 and above. In reality, informers should be done as a GenericHost on top of NetCoreApp. Therefore I propose that it is isolated into its own sub-library (ex. KubernetesClient.Controllers) as to not remove targets from Client SDK. Asking community/maintainers on the best way to solve this.
I've created separate csproj and SLN files as my IDE don't play well with conditional compiler flags. These will be removed in the final version

**Currently is missing unit tests so don't merge yet** but contains most of core implementation. Submitting PR early to get feedback while I add unit tests. There is a similar PR open https://github.com/kubernetes-client/csharp/pull/347 but this one is much further ahead


macsux:
CLA should now be signed

brendandburns:
Can you work with @kkmsft in conjunction with #347 to get to a single implementation of this?

Perhaps you can share the implementation work.

Thanks!

macsux:
I commented on the other PR, but given that this one is much further ahead I propose the other one is closed and we continue iterating on this codebase. I'm happy to get help covering it with tests and creating examples.

@brendandburns can you please comment on the things I mentioned in the header especially around idea of creating a subpackage due to dependencies that were taken on?

src/KubernetesClient/Informer/Cache/Lister.cs
```diff
@@ -0,0 +1,58 @@
...
+            }
+
+            return _indexer.ByIndex(_indexName, _namespace);
+        }
+
+        public TApiType Get(string name)
+        {
+            string key = name;
+            if (_namespace?.Length != 0)
+            {
```
kkmsft: Can we please move this to a generic util method ?

kkmsft:
@macsux - Thank you for your PR. 

- Can you please add  a simple real example like https://github.com/kubernetes-client/csharp/blob/e267e68b65863fc69fc7b5d1e9eddfb5bb34a572/examples/informers/Program.cs in interacting with the API Server ?

- Can you  please provide some insights on how you were able to do the listing without using a generic list from the API Server (like: https://github.com/kubernetes-client/csharp/blob/e267e68b65863fc69fc7b5d1e9eddfb5bb34a572/src/KubernetesClient/Kubernetes.List.cs) ?

kkmsft:
@macsux - any thoughts on the above ^^ ?

macsux:
> @macsux - any thoughts on the above ^^ ?
I got swamped at work, been a bit busy, but should get a breathing room in the coming days to finish this off.
See intended usage pattern here
https://github.com/macsux/kubernetes-client-csharp/blob/91d052abc2660f5a04de329b259fd523ab642874/tests/KubernetesClient.Tests/Informer/Cache/ReflectorTest.cs#L37

Basically you pass in a delegate to the list method that's part of core K8s SDK. This is the way it works in java API. Having said that there's room for improvement. Realistically I want to add a generic method to hit the API for any type of object. The autogenerated methods have a lot of duplicate code that bogs down IDE and makes assembly way bigger then it should be, and makes calling CRDs a pita. 

I'll add full samples when I finish covering with unit tests

macsux:
Just an update, as I started porting in unit tests I became deeply unhappy in their structure as the ones in Java rely on delays to propagate messages to end as many components are control loops. I'm refactoring interaction between many components to be based on tpl dataflow blocks to allow 1.observability of completion
2.in memory actor model allowing for many threading locks to be removed

macsux:
This work has been superseded by a complete rewrite that is more flexible in architecture and leverages native paradigms. It's now open as a different PR here https://github.com/kubernetes-client/csharp/pull/394
I'm closing this PR as I don't plan on working on this stream of work, as the other PR offers a much better implementation.

PR #353, opened 2020-02-06
Improve async config loading.

### LINKED ISSUE #352: BuildConfigFromConfigFile() hangs
The static methods on `KubernetesClientConfiguration` like `BuildConfigFromFile` are hanging when using from certain environments where SynchronisationContext is not default. For instance WPF, Xamarin etc. Specifically on any line that uses `.GetAwaiter().GetResult();`.

Could you provide true async methods where you are waiting for operation result?

brendandburns:
Fixes #352


aloneguid:
Lgtm

Issue #357, opened 2020-02-12
Build fails on VS Enterprise 2019 Version 16.4.5


pragyamehta:
**The TargetFramework value 'xamarinios10' was not recognized. It may be misspelled. If not, then the TargetFrameworkIdentifier and/or TargetFrameworkVersion properties must be specified explicitly.	

The TargetFramework value 'monoandroid81' was not recognized. It may be misspelled. If not, then the TargetFrameworkIdentifier and/or TargetFrameworkVersion properties must be specified explicitly.**

Mobile development with .NET is already installed for this VS
![Annotation 2020-02-11 162805](https://user-images.githubusercontent.com/22248649/74291961-81fba680-4ceb-11ea-9971-62b61ab311c8.png)

brendandburns:
I admit I don't really use VS Enterprise so I'm not sure I can help with this. If you figure it out, I'm happy to take a PR to fix the .sln/.csproj files.

tg123:
you may want use dotnet build to avoid this
@brendanburns I suggest to remove those targets as in #345 they are never in our nuget package and covered by netstandard

the reason they were introduce is to support ssl and websocket on those platfrom

brendandburns:
@tg123 does that mean that a Xamarin user can use the `netstandard` nugets from mobile?

tg123:
@brendanburns
Yes 
https://docs.microsoft.com/en-us/xamarin/cross-platform/app-fundamentals/net-standard
https://docs.microsoft.com/en-us/dotnet/standard/net-standard

I am happy to reforge the targets if you approve

brendandburns:
@tg123 sure, that sounds great, thanks.

Issue #317, opened 2019-10-28
Nuget package is not getting built for all target frameworks


alexanderboswell:
In the [project file](https://github.com/kubernetes-client/csharp/blob/master/src/KubernetesClient/KubernetesClient.csproj), net452 is listed as a target framework but when viewing the published package it is not listed as being supported:
![image](https://user-images.githubusercontent.com/21343215/67705723-c8afd580-f974-11e9-959c-3b0f9c474e09.png)

It looks like in the [project file](https://github.com/kubernetes-client/csharp/blob/master/src/KubernetesClient/KubernetesClient.csproj) it is only building the packages for `"'$(OS)' != 'Windows_NT'"`.

I have been able to build the package locally for net452, so it seems to be supported and working but the build pipeline is not generating a package.

tg123:
This is because the project is built on linux.
Github Action is a better place to have the build run where Windows is supported

brendandburns:
Yeah, we should migrate to github actions...

tg123:
start at KubernetesClient 1.6.20
net452 was included

Issue #303, opened 2019-09-08
Why so many targeted frameworks?


DavidParks8:
Targeting a bunch of specific frameworks kind of defeats the point of using .net standard, does it not? I looked through the code and it seems like there are good initial reasons for targeting a bunch of platforms, but now that .net core has matured a bit, it might be time to look into unifying to just .net standard 2.0 and adding convenience methods/support for Microsoft.Extensions.DependencyInjection and Microsoft.Extensions.Configuration.

Thoughts?

brendandburns:
I want to make sure that we continue to support Mono/Xamarin.

And if I remember correctly there were challenges with .net standard and SSL certificate handling.

But in general, I the goal would be a minimal set that spans all relevant use cases (Framework, Core and Mobile)

tg123:
@brendandburns 

<https://docs.microsoft.com/en-us/dotnet/standard/net-standard>

IMHO, we should target `.netstandard` only for all platform

Issue #239, opened 2019-02-07
Please add net452 to KubernetesClient NuGet package


xinyanmsft:
Please include net452 binary in the NuGet package.

Today when I invoke 'dotnet pack' on my machine, the resulting package contains net452. However the downloaded official package does not include it.

This means, when running in a full framework process, things like WebSocket authentication will not work (the issue was fixed in pull request #231, but NuGet package needs to include 452 to make it available).

Thanks!

tg123:
I am working with integrating with azure pipeline which will enable windows base targets.

brendandburns:
Yeah, right now the build step happens in Linux on Travis, that's probably not ideal...

brendandburns:
Now that the pipeline is checked in, I will create a pipeline to build that artifact...

PR #361, opened 2020-03-06
[breaking changes] cleanup targets

tg123:
This is to cleanup targets to ensure the quality of code

<https://docs.microsoft.com/en-us/dotnet/standard/net-standard>
remove xamarin ios and android as they can reference netstandard2 directly (confirmed here #140)

removing those targets would also help building issues (#357)
and this is the first step to build nuget on gh action (cannot target xaramin). #239 #317 

here is the matrix after this patch

- [ ] netstandard1.4;
- [X] netstandard2.0;
- [X] net452;
- [ ] net461;
- [X] netcoreapp2.1;
- [ ] xamarinios10;
- [ ] monoandroid81

some comments for why other platforms are kept/removed:
`net452` it is still [supported by MS until 2023](https://stackoverflow.com/questions/58732338/when-is-end-of-life-for-dotnet-framework-4-5-2)
`netstandard1.4` [since `netcore1.0` is no longer support](https://devblogs.microsoft.com/dotnet/net-core-1-0-and-1-1-will-reach-end-of-life-on-june-27-2019/), netcore app should migrate to netcore2, we drop support for `netstandard1` there
`netcoreapp2.1`, I am working on to migrated those certificates code to pure `netstandard2.0`  will send another PR after this is merged
`net461`, included in `netstandard2.0`

see aslo
#303 #345

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -31,7 +31,7 @@ echo Generating Code Coverage reports
 export PATH="$PATH:$HOME/.dotnet/tools"
 export DOTNET_ROOT=$(dirname $(realpath $(which dotnet))) # https://github.com/dotnet/cli/issues/9114#issuecomment-401670622
 dotnet tool install --global dotnet-reportgenerator-globaltool --version 4.0.15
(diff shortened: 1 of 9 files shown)


brendandburns:
/lgtm
/approve

Thanks!

PR #365, opened 2020-03-09
fix DisposeWatch test fickle

tg123:
seems a regression in upgrading t `Nito.AsyncEx 5.0`


tg123:
There are many other `await Task.WhenAny(eventsReceived.WaitAsync(), Task.Delay(TestTimeout));`

still may cause test fail but this case fail frequently,
I am file a bug to AsyncEx

PR #366, opened 2020-03-09
revert move fraction internal

tg123:
revert #194 

as fraction 4.0.1 supports strong-name


brendandburns:
close/reopen to retrigger tests.

PR #369, opened 2020-03-13
fix watch generator uses different name pattern from gen project

### LINKED ISSUE #362: Some kubernetes objects do not implement IKubernetesObject
I have found some Kubernetes objects which do not implement IKubernetesObject:
- Extensionsv1beta1Ingress
- Networkingv1beta1Ingress
I use the Kubernetes client  version 1.6.16-gb07e78afa4 
Is this on purpose or a bug?

tg123:
fix #362 

also fix some watch operations are missing

### DIFF
--- a/gen/KubernetesWatchGenerator/Program.cs
+++ b/gen/KubernetesWatchGenerator/Program.cs
@@ -13,34 +13,38 @@ namespace KubernetesWatchGenerator
 {
     class Program
     {
+        static readonly Dictionary<string, string> ClassNameMap = new Dictionary<string, string>();
+
         static async Task Main(string[] args)
         {
             if (args.Length < 2)
             {
-                Console.Error.WriteLine($"usage {args[0]} path/to/csharp.settings");
+                Console.Error.WriteLine($"usage {args[0]} path/to/generated");
                 Environment.Exit(1);
             }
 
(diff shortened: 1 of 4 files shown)


gen/KubernetesWatchGenerator/Program.cs
```diff
@@ -150,7 +101,7 @@ static void ToXmlDoc(RenderContext context, IList<object> arguments, IDictionary
                     {
                         if (!first)
                         {
-                            context.Write(Environment.NewLine);
+                            context.Write("\n");
```
brendandburns: Shouldn't we keep the `Environment.NewLine` ?

brendandburns:
Looks good, one really minor comment.

gen/KubernetesWatchGenerator/Program.cs
tg123: it should be `eol` of the generated files, `IKubernetes.Watch.cs`, `Kubernetes.Watch.cs` and `ModelExtensions.cs`.

those files are generated using `lf` already.
and should be generated using `lf` only as `.editorconfig` defined the rule `eol = lf`

brendandburns: ok, thanks.

PR #370, opened 2020-03-14
Ensure that awaits do not continue on the captured context.

andykernahan:
Hi,

This PR is a continuation of #353 which addressed  #352 . Unfortunately, the synchronous methods still result in deadlock when invoked from a non-default synchronization context; this is caused by an await in the call graph continuing on the captured context.

For example, the following call graph results in deadlock:

```csharp
KubernetesClientConfiguration.BuildConfigFromConfigFile(...)
KubernetesClientConfiguration.BuildConfigFromConfigFileAsync(...)
KubernetesClientConfiguration.LoadKubeConfigAsync(...)
Yaml.LoadFromStreamAsync(...)
```

I have also taken the opportunity to update all other awaits in non-generated code.

Cheers.

### DIFF
--- a/src/KubernetesClient/StreamDemuxer.cs
+++ b/src/KubernetesClient/StreamDemuxer.cs
@@ -57,7 +57,7 @@ namespace k8s
         /// </summary>
         public void Start()
         {
-            this.runLoop = this.RunLoop(this.cts.Token);
+            this.runLoop = Task.Run(() => this.RunLoop(this.cts.Token));
         }
 
         /// <inheritdoc/>
@@ -193,9 +193,6 @@ namespace k8s
 
         protected async Task RunLoop(CancellationToken cancellationToken)
         {
-            // This is a background task. Immediately yield to the caller.
-            await Task.Yield();
-
             // Get a 1KB buffer
             byte[] buffer = ArrayPool<byte>.Shared.Rent(1024 * 1024);
             // This maps remembers bytes skipped for each stream.
--- a/src/KubernetesClient/StreamDemuxer.cs
+++ b/src/KubernetesClient/StreamDemuxer.cs
@@ -57,7 +57,7 @@ namespace k8s
         /// </summary>
         public void Start()
         {
-            this.runLoop = Task.Run(() => this.RunLoop(this.cts.Token));
+            this.runLoop = Task.Run(async () => await this.RunLoop(this.cts.Token));
         }
 
         /// <inheritdoc/>
--- a/src/KubernetesClient/Watcher.cs
+++ b/src/KubernetesClient/Watcher.cs
@@ -57,7 +57,7 @@ namespace k8s
             OnClosed += onClosed;
 
             _cts = new CancellationTokenSource();
-            _watcherLoop = this.WatcherLoop(_cts.Token);
+            _watcherLoop = Task.Run(() => this.WatcherLoop(_cts.Token));
         }
 
         /// <inheritdoc/>
@@ -91,14 +91,11 @@ namespace k8s
 
         private async Task WatcherLoop(CancellationToken cancellationToken)
         {
-            // Make sure we run async
-            await Task.Yield();
-
             try
             {
                 Watching = true;
                 string line;
-                _streamReader = await _streamReaderCreator();
+                _streamReader = await _streamReaderCreator().ConfigureAwait(false);
 
                 // ReadLineAsync will return null when we've reached the end of the stream.
                 while ((line = await _streamReader.ReadLineAsync().ConfigureAwait(false)) != null)
@@ -164,7 +161,7 @@ namespace k8s
             Action onClosed = null)
         {
             return new Watcher<T>(async () => {
-                var response = await responseTask;
+                var response = await responseTask.ConfigureAwait(false);
 
                 if (!(response.Response.Content is WatcherDelegatingHandler.LineSeparatedHttpContent content))
                 {
--- a/src/KubernetesClient/Watcher.cs
+++ b/src/KubernetesClient/Watcher.cs
@@ -57,7 +57,7 @@ namespace k8s
             OnClosed += onClosed;
 
             _cts = new CancellationTokenSource();
-            _watcherLoop = Task.Run(() => this.WatcherLoop(_cts.Token));
+            _watcherLoop = Task.Run(async () => await this.WatcherLoop(_cts.Token));
         }
 
         /// <inheritdoc/>
--- a/tests/KubernetesClient.Tests/KubernetesClientConfigurationTests.cs
+++ b/tests/KubernetesClient.Tests/KubernetesClientConfigurationTests.cs
@@ -406,6 +406,13 @@ namespace k8s.Tests
             AssertConfigEqual(expectedCfg, cfg);
         }
 
+        [UIFact]
+        public void BuildConfigFromConfigFileInfoOnNonDefaultSynchronizationContext()
+        {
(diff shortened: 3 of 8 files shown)


src/KubernetesClient/StreamDemuxer.cs
```diff
@@ -57,7 +57,7 @@ public StreamDemuxer(WebSocket webSocket, StreamType streamType = StreamType.Rem
         /// </summary>
         public void Start()
         {
-            this.runLoop = this.RunLoop(this.cts.Token);
+            this.runLoop = Task.Run(() => this.RunLoop(this.cts.Token));
```
tg123: here the handle is to keep the task ref-ed by this obj and could be `waited` for in the future
in this case, `dispose` will wait for runloop to get things done and do some clean up.

wrapping with a task will lose the control to runloop task totally.
Task.Run will return immediately with an null

you may want `Task.Run(async() => await RunLoop())` if you want to keep sync context clean.
but I do not think it is necessary because no one wait on its parent obj

src/KubernetesClient/Watcher.cs
```diff
@@ -57,7 +57,7 @@ public Watcher(Func<Task<StreamReader>> streamReaderCreator, Action<WatchEventTy
             OnClosed += onClosed;
 
             _cts = new CancellationTokenSource();
-            _watcherLoop = this.WatcherLoop(_cts.Token);
+            _watcherLoop = Task.Run(() => this.WatcherLoop(_cts.Token));
```
tg123: same as above

src/KubernetesClient/StreamDemuxer.cs
andykernahan: > wrapping with a task will lose the control to runloop task totally. Task.Run will return immediately with an null

This change does not produce that behaviour.

As specified the [documentation](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.run?view=netcore-2.0#System_Threading_Tasks_Task_Run_System_Func_System_Threading_Tasks_Task__), the `Task.Run(Func<Task>)` overload returns a proxy for the task returned by the provided `function`; therefore, the status of `runLoop` will reflect exactly that of the task returned by `RunLoop`.

We can observe this behaviour with [this](https://gist.github.com/andykernahan/a39fba0445e3936728f18447a6b7d841) `netcoreapp2.0` program which prints a value to stdout after simulating an asynchronous workload. As expected, the program prints the value and then exits.

> you may want `Task.Run(async() => await RunLoop())` if you want to keep sync context clean.
but I do not think it is necessary because no one wait on its parent obj

I'm afraid that I do not understand: making the suggested change does not affect the `SynchronizationContext` from any perspective.

The `function` is scheduled on `TaskScheduler.Default` which in turn executes it on the `ThreadPool` where there is no `SynchronizationContext`; therefore, there is no `SynchronizationContext` when `function` is invoked.

We can observe this behaviour with [this](https://gist.github.com/andykernahan/3f1a04a130098e3e358128b5220f676d) `netcoreapp2.0` program which prints `SynchronizationContext.Current`. As expected, the programs prints the expected `No SynchronizationContext`.

src/KubernetesClient/Watcher.cs
andykernahan: Please see above.

src/KubernetesClient/StreamDemuxer.cs
tg123: Yes you are right 
at first I think your code is like below:
```
            var task = Task.Run(() =>
            {
                Task.Delay(TimeSpan.FromDays(1));
            });

            task.Wait(); // did not wait for the inner task
            Console.WriteLine("wait 1 d done"); 
```

`Task.Run(() => some task)` return and wait `some task` implicitly.

but if your code is like this 
```
            var task = Task.Run(async () =>
            {
                await Task.Delay(TimeSpan.FromDays(1));
            });
```
or
```
 Task.Run(async () =>await Task.Delay(TimeSpan.FromDays(1)));
```

you can keep bug away when editing the code in the future. 
That would be better for people to read the code.

tg123:
/LGTM 

IHMO `async () => await` might be a better way

src/KubernetesClient/StreamDemuxer.cs
brendandburns: Let's use @tg123 's suggestion to make the code more readable and less compact.

tests/KubernetesClient.Tests/KubernetesClientConfigurationTests.cs
```diff
@@ -383,6 +383,13 @@ public void LoadKubeConfigStream()
             AssertConfigEqual(expectedCfg, cfg);
         }
 
+        [UIFact]
```
brendandburns: Is there a reason why this is a `UIFact`? If so, please add a comment for future readers (and me! :)

src/KubernetesClient/StreamDemuxer.cs
andykernahan: The maintainability aspect is interesting. However, converting the lambda to be statement bodied and omitting the `await` (as in your example) generates a CS4014 compiler warning due to the call not being awaited.

Would you like me to apply @tg123 's statement bodied suggestion or the expression bodied suggestion?

tests/KubernetesClient.Tests/KubernetesClientConfigurationTests.cs
andykernahan: The use of [`UIFactAttribute`](https://github.com/AArnott/Xunit.StaFact/blob/v0.3.18/src/Xunit.StaFact/UIFactAttribute.cs) is intrinsic to the test: it results in the test method being invoked from within a UI-like `SynchronizationContext`, which reproduces the issue this PR fixes.

I will add a comment as suggested.

src/KubernetesClient/StreamDemuxer.cs
tg123: hi
my suggestion is ` Task.Run(async () => await this.RunLoop(this.cts.Token));`

this one is show how bug happens
```
this.runLoop = Task.Run(() =>
{
     this.RunLoop(this.cts.Token); // bug forget await, same as your compiler warning
     // people wants to add something new
});
```

people might forget to await that `RunLoop`

while ` Task.Run(async () => await this.RunLoop(this.cts.Token));` will become

` Task.Run(async () => {
await this.RunLoop(this.cts.Token)); // never forget to await
// people wants to add something new
}`

my first look into the code is like `Task.Run( void () => ... )`. So I think it might be a bug.

tg123:
rebase master to pass the ci

src/KubernetesClient/StreamDemuxer.cs
andykernahan: > this one is to show how bug happens, not suggestion

Yes, I understood that; I referred to that as your example, not suggestion.

I'm happy to make the suggested change of:

```csharp
Task.Run(async () => await this.RunLoop(this.cts.Token));
```

tg123: will compiler give warning for this one?

andykernahan: No. The compiler will not warn for any of the following:
```csharp
this.runLoop = Task.Run(() => this.RunLoop(this.cts.Token));
this.runLoop = Task.Run(async () => await this.RunLoop(this.cts.Token));
this.runLoop = Task.Run(async () =>
{
    await this.RunLoop(this.cts.Token);
});
```

Issue #372, opened 2020-03-16
KubernetesClientConfiguration.BuildDefaultConfig() does not populate namespace of the current context


rynowak:
I'm using `KubernetesClientConfiguration.BuildDefaultConfig()` to try and emulate things operations the user could do at the command-line (using their current context).

I noticed that I don't get the namespace populated in the config. In this case I would expect it to be set to `blazing-pizza` but I get back `null`.

```C#
var config = KubernetesClientConfiguration.BuildDefaultConfig();
Console.WriteLine(config.Namespace); // always null
```

My contexts:

```txt
$ kubectl config get-contexts
CURRENT   NAME           CLUSTER        AUTHINFO                                NAMESPACE
*         rynowak-kubo   rynowak-kubo   <user>                                 blazing-pizza
```

```txt
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.4", GitCommit:"8d8aa39598534325ad77120c120a22b3a990b5ea", GitTreeState:"clean", BuildDate:"2020-03-12T23:41:24Z", GoVersion:"go1.14", Compiler:"gc", Platform:"darwin/amd64"}
```

What I've noticed about this is that I can work around this because it's populated on `ContextDetails` but not on `Context`. I haven't been able to find any examples on the net of a case where `Context.Namespace` would get populated based on the yaml. Might be historical?

<img width="595" alt="image" src="https://user-images.githubusercontent.com/1430011/76782588-bc82b400-676d-11ea-9f50-77bdfdded5a1.png">

My (heavily redacted) `~/.kube/config`

```yaml
apiVersion: v1
clusters:
- cluster:
    certificate-authority-data: <cert>
    server: <server>
  name: rynowak-kubo
- cluster:
    certificate-authority-data: <cert>
    server: <server>
  name: rynowak-ndc-demo
contexts:
- context:
    cluster: rynowak-kubo
    namespace: blazing-pizza
    user: <user>
  name: rynowak-kubo
current-context: rynowak-kubo
kind: Config
preferences: {}
users:
- name: <user>
  user:
    client-certificate-data:  <cert>
    client-key-data: <key>
    token: <token>
- name: <user>
  user:
    client-certificate-data: <cert>
    client-key-data: <key>
    token: <token>
```

rynowak:
[...]

PR #373, opened 2020-03-16
Use value from ContextDetails to populate Namespace

rynowak:
**Use value from ContextDetails to populate Namespace**

Fixes: #372

This change uses the value from ContextDetails.Namespace to populate
KubernetesClientConfiguration.Namespace.

The issue is there's a Namespace property on both Context and ContextDetails
 - The property on Context is used today
 - The property on ContextDetails is not
 - The property on ContextDetails maps to the actual yaml config

---

**Obsolete Context.Namespace**

*I did a separate commit for this so we can skip it if you don't want it. Reasonable people disagree on whether its good semver to obsolete something in a patch/minor release. I'd like some confirmation as well about whether this property is supposed to be there for some reason*

This property doesn't map to anything in the YAML and thus will never be
set. Other clients I checked (java, golang) don't look for a property
at this level.

I think this was likely a mistake, and it should be obsoleted because
it will never be populated.

Example:

```yaml
contexts:
- context:
    cluster: ...
    namespace: ... # this is ContextDetails.Namespace
    user: ...
  name: foo
```

```yaml
contexts:
- context:
    cluster: ...
    namespace: ...
    user: ...
  name: foo
  namespace: ... # this is Context.Namespace
```

---

### DIFF
--- a/src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
+++ b/src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
@@ -209,7 +209,7 @@ namespace k8s
             SetUserDetails(k8SConfig, activeContext);
 
             // namespace
-            Namespace = activeContext.Namespace ?? activeContext.ContextDetails?.Namespace;
+            Namespace = activeContext.ContextDetails?.Namespace;
         }
 
         private void SetClusterDetails(K8SConfiguration k8SConfig, Context activeContext)
--- a/src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
+++ b/src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
@@ -193,6 +193,13 @@ namespace k8s
                 throw new KubeConfigException($"CurrentContext: {currentContext} not found in contexts in kubeconfig");
             }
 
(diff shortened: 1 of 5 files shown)


rynowak:
I signed it

src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
```diff
@@ -193,6 +193,13 @@ private void InitializeContext(K8SConfiguration k8SConfig, string currentContext
                 throw new KubeConfigException($"CurrentContext: {currentContext} not found in contexts in kubeconfig");
             }
 
+            if (string.IsNullOrEmpty(activeContext.ContextDetails?.Cluster))
+            {
+                // This serves as validation for any of the properties of ContextDetails being set.
+                // Other locations in code assume that ContextDetails is non-null.
+                throw new KubeConfigException($"Cluster not set for context `{currentContext}` in kubeconfig");
+            }
```
rynowak: Added this validation because I see a bunch of code running after this at assumes `activeContext.ContextDetails` is null, and it will crash with a null-ref based on invalid config data.

brendandburns:
/lgtm
/approve

Thanks for the PR!

PR #376, opened 2020-03-20
fix watch=true doesn't respect cancellation token

### LINKED ISSUE #375: ListNamespacedServiceWithHttpMessagesAsync with watch=true doesn't respect cancellation token with empty list
When trying to watch for service based on fieldSelector and the service doesn't exist, cancellation token is not respected and the call hangs.

``` C#
var cts = new CancellationTokenSource();
cts.CancelAfter(TimeSpan.FromSeconds(30));

var service = await KubernetesClient.ListNamespacedServiceWithHttpMessagesAsync(
                        namespaceParameter: DefaultNamespace,
                        fieldSelector: "metadata.name=non-existing-service",
                        watch: true,
                        cancellationToken: cts.Token);
```

KuberneteClient version: `1.6.12`

_Edit: Added timeout for cancellation token cancellation._

tg123:
fix #375

### DIFF
--- a/src/KubernetesClient/WatcherDelegatingHandler.cs
+++ b/src/KubernetesClient/WatcherDelegatingHandler.cs
@@ -48,16 +48,14 @@ namespace k8s
             public override void Flush() => _innerStream.Flush();
 
             public override int Read(byte[] buffer, int offset, int count) =>
-                _innerStream.ReadAsync(buffer, offset, count, _cancellationToken).ConfigureAwait(false)
-                    .GetAwaiter().GetResult();
+                _innerStream.ReadAsync(buffer, offset, count, _cancellationToken).GetAwaiter().GetResult();
 
             public override long Seek(long offset, SeekOrigin origin) => _innerStream.Seek(offset, origin);
 
             public override void SetLength(long value) => _innerStream.SetLength(value);
 
             public override void Write(byte[] buffer, int offset, int count) =>
-                _innerStream.WriteAsync(buffer, offset, count, _cancellationToken).ConfigureAwait(false).GetAwaiter()
-                    .GetResult();
+                _innerStream.WriteAsync(buffer, offset, count, _cancellationToken).GetAwaiter().GetResult();
 
             public override bool CanRead => _innerStream.CanRead;
 
--- a/src/KubernetesClient/WatcherDelegatingHandler.cs
+++ b/src/KubernetesClient/WatcherDelegatingHandler.cs
@@ -45,32 +45,19 @@ namespace k8s
                 _cancellationToken = cancellationToken;
             }
 
-            public override void Flush()
-            {
-                _innerStream.Flush();
-            }
+            public override void Flush() => _innerStream.Flush();
 
-            public override int Read(byte[] buffer, int offset, int count)
-            {
-                return _innerStream.ReadAsync(buffer, offset, count, _cancellationToken).ConfigureAwait(false)
+            public override int Read(byte[] buffer, int offset, int count) =>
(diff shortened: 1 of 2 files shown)


src/KubernetesClient/WatcherDelegatingHandler.cs
```diff
@@ -27,20 +27,76 @@ internal class WatcherDelegatingHandler : DelegatingHandler
...
+            private readonly Stream _innerStream;
+            private readonly CancellationToken _cancellationToken;
+
+            public CancelableStream(Stream innerStream, CancellationToken cancellationToken)
+            {
+                _innerStream = innerStream;
+                _cancellationToken = cancellationToken;
+            }
+
+            public override void Flush()
```
brendandburns: `public override void Flush() => _innertStream.Flush;` ?

src/KubernetesClient/WatcherDelegatingHandler.cs
```diff
@@ -27,20 +27,76 @@ internal class WatcherDelegatingHandler : DelegatingHandler
...
+                _innerStream.Flush();
+            }
+
+            public override int Read(byte[] buffer, int offset, int count)
+            {
+                return _innerStream.ReadAsync(buffer, offset, count, _cancellationToken).ConfigureAwait(false)
+                    .GetAwaiter().GetResult();
+            }
+
+            public override long Seek(long offset, SeekOrigin origin)
```
brendandburns: Can you use delegate syntax here?

src/KubernetesClient/WatcherDelegatingHandler.cs
```diff
@@ -49,13 +105,13 @@ protected override async Task SerializeToStreamAsync(Stream stream, TransportCon
...
 
-                StreamReader = new PeekableStreamReader(_originStream);
+                StreamReader = new PeekableStreamReader(new CancelableStream(_originStream, _cancellationToken));
 
                 var firstLine = await StreamReader.PeekLineAsync();
 
                 var writer = new StreamWriter(stream);
 
-//                using (writer) // leave open
+                //                using (writer) // leave open
```
brendandburns: let's just delete this line.

brendandburns:
Couple of small style comments.

src/KubernetesClient/WatcherDelegatingHandler.cs
```diff
@@ -27,20 +27,63 @@ internal class WatcherDelegatingHandler : DelegatingHandler
...
+            public CancelableStream(Stream innerStream, CancellationToken cancellationToken)
+            {
+                _innerStream = innerStream;
+                _cancellationToken = cancellationToken;
+            }
+
+            public override void Flush() => _innerStream.Flush();
+
+            public override int Read(byte[] buffer, int offset, int count) =>
+                _innerStream.ReadAsync(buffer, offset, count, _cancellationToken).ConfigureAwait(false)
```
andykernahan: Consider omitting `ConfigureAwait(false)` as it does not affect the behaviour of the awaiter's `GetResult`. It is also potentially misleading to the reader as they may falsely believe these methods (`CancelableStream.{Read, Write}`) to be insusceptible to the classic async/await deadlock.

andykernahan: Consider omitting `ConfigureAwait(false)` as it does not affect the behaviour of the awaiter's `GetResult`; therefore, it is potentially misleading to the reader as they may falsely believe these methods (`CancelableStream.{Read, Write}`) to be insusceptible to the classic async/await deadlock.

tg123: fixed thanks

brendandburns:
/lgtm
/approve

Thanks!

PR #378, opened 2020-03-20
Add interfaces for generated Kubernetes objects

macsux:
Add interfaces for generated Kubernetes objects that can allow working with them without using concrete types. This work is needed for future shared informers / controllers components being developed


macsux:
Seems like travis build just timed out for whatever reason

PR #379, opened 2020-03-23
Create watch handle for http client factory

### LINKED ISSUE #374: Using List...Async(watch: true) with HttpClientFactory the .Watch callback is never invoked
I have setup a `Kubernetes` using a `HttpClient` from `HttpClientFactory`, and called one of the `List...Async` methods, passing `watch: true`. i have then setup the the watch callback using the extension methods
`resp.Watch((eventType, resource) => _logger.LogInformation("{type}, {resource}", eventType, resource))`
However my callback is never invoked - even when I add/remove a matching resource in the cluster.

If don't provide a `HttpClient` when constructing the `Kubernetes` object everything works as expected - The callback is initially invoked for all of the current resources, and then any add/remove/updates to the cluster also cause the callbacks to be called.

narthollis:
Resolves #374 

Introduce a new static Method on the KubernetesClientConfiguration class for use with HttpClientFactory and AddTypedClient.

The addition of using `.AddHttpMessageHandler(KubernetesClientConfiguration.CreateWatchHandler);` allows `List...Async().Watch` to work.

### DIFF
--- a/examples/httpClientFactory/Program.cs
+++ b/examples/httpClientFactory/Program.cs
@@ -32,7 +32,8 @@ namespace httpClientFactory
                                 serviceProvider.GetRequiredService<KubernetesClientConfiguration>(),
                                 httpClient);
                         })
-                        .ConfigurePrimaryHttpMessageHandler(config.CreateDefaultHttpClientHandler);
+                        .ConfigurePrimaryHttpMessageHandler(config.CreateDefaultHttpClientHandler)
+                        .AddHttpMessageHandler(KubernetesClientConfiguration.CreateWatchHandler);
 
                     // Add the class that uses the client
                     services.AddHostedService<PodListHostedService>();
--- a/src/KubernetesClient/KubernetesClientConfiguration.HttpClientHandler.cs
+++ b/src/KubernetesClient/KubernetesClientConfiguration.HttpClientHandler.cs
@@ -14,7 +14,7 @@ namespace k8s
             {
                 if(this.SkipTlsVerify)
                 {
-                    httpClientHandler.ServerCertificateCustomValidationCallback = 
+                    httpClientHandler.ServerCertificateCustomValidationCallback =
                         (sender, certificate, chain, sslPolicyErrors) => true;
                 }
                 else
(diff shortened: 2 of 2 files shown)


tg123:
I am thinking if there is a way to ensure the handler is installed for every single http client passing to Kubernetes ctor

narthollis:
I think the only way to do that will be an set of extension methods for
setting up the typed client.

To avoid additional dependencies this is something we could potentially do
as an additional NuGet package

But as was mentioned there is not a way to modify the client after creation.

If wanted I can look at creating a set of extension methods in place if
this single function?

narthollis:
To expand on my previous comment,

The best way to get a "single call gets everything setup" with `HttpClientFactory` would be an extension method to `IHttpClientBuilder` such that the usage would look something like:

```csharp

services.AddHttpClient("k8s")
    .ConfigureForKubernetesClient(s => s.GetRequireService<KubernetesClientConfiguration>());

```

The problem with this approach is that it would introduce a dependency on `Microsoft.Extensions.Http` which in turn has dependencies on `Microsoft.Extensions.DependencyInjection.Abstractions`, `Microsoft.Extensions.Logging`, `Microsoft.Extensions.Options` - which in turn have there own set of dependencies.

The cleanest way to provide such an extension method without unnecessarily increasing the dependency bloat would be to introduce a new NuGet package (say `KubernetesClient.HttpExtensions`) that contains these extension methods.

tg123:
/LGTM

I think, at least, it is better to have this workaround than no way to do so.
My only concern is that we did not prevent ppl from forgetting to add handler in the future.

narthollis:
Is it normal for these tests to fail on some environments and not others?

ie. are these build failures something I need to be looking into?

Unless the examples are used as part of the tests I would not have expected my changes to effect the tests.

tg123:
@narthollis  please rebase with master to fix the failure

narthollis:
Ok, builds are now all passing.

If it is of interest I will take a look at creating another pull request with a new project containing extension methods for configuring a KubernetesClient `HttpClientFactory`

PR #385, opened 2020-03-25
Gen v1.18.0

tg123:


brendandburns:
We should really figure out how we want to version things. In the Java client we introduce a new major version for each version of the Kubernetes API. That way people are fore-warned there might be breaking changes.

tg123:
agreed
i think we should include k8s version in nuget version.

brendandburns:
@tg123 can you take care of refining the version, then we can merge this.

tg123:
@brendandburns for sure, I will send out a proposal about how versioning.

tg123:
@brendandburns what do you think if we bump this to `v2.0`
and bump major version on every code generating

this is the solution 1 in #400  (same as java)

brendandburns:
That sounds great to me.

--brendan

tg123:
bumped to 2.0 and ready to merge

PR #386, opened 2020-03-26
Improve Watcher thread utilisation.

### LINKED ISSUE #381: Watcher thread utilisation.
Currently each `Watcher` instance causes a thread to be blocked until the next `WatchEvent` is read from the HTTP connection. In a system where there are many `Watcher` instances, this may cause the number of threads to grow large.

I believe this behaviour is caused by `WatcherDelegatingHandler` (and its related types) not being asynchronous all the way. The following (abbreviated) call graph illustrates this:
```
Watcher.WatcherLoop
 PeekableStreamReader.ReadLineAsync
  StreamReader.ReadLineAsync
   TextReader.ReadLineAsync               // (1) starts task
    PeekableStreamReader.ReadLine         // (2)
     CancelableStream.Read                // waits
      ChunkedEncodingReadStream.ReadAsync // (3)
```
The task (and its thread) started by 1 to asynchronously invoke 2 is ultimately blocked waiting for 3.

We can observe this behaviour by viewing the parallel stacks after creating ten `Watcher` instances:
```csharp
public async Task WatcherThreadBehaviour()
{
    using (var server = new MockKubeApiServer(testOutput, async httpContext =>
    {
        httpContext.Response.StatusCode = 200;
        await httpContext.Response.Body.FlushAsync();
        await Task.Delay(Timeout.Infinite);
        return true;
    }))
    {
        var client = new Kubernetes(new KubernetesClientConfiguration
        {
            Host = server.Uri.ToString()
        });
        for (int i = 0; i < 10; i++)
        {
            client.ListNamespacedPodWithHttpMessagesAsync("default", watch: true)
                  .Watch<V1Pod, V1PodList>(delegate { });
        }
        await Task.Delay(Timeout.Infinite);
    }
}
```
We observe the ten blocked threads:

parallel stacks

Note this behaviour predates #376, prior to which we observed the following:
```
Watcher.WatcherLoop
 PeekableStreamReader.ReadLineAsync
  StreamReader.ReadLineAsync
   TextReader.ReadLineAsync         // (1) starts task
    PeekableStreamReader.ReadLine   // (2)
     ChunkedEncodingReadStream.Read // (3)
```
parallel stacks

I believe it possible to eliminate this behaviour by making `WatcherDelegatingHandler` asynchronous all the way. I'm happy to undertake this change, or liaise with @tg123 to effect it?

andykernahan:
This PR fixes #381. We no longer observe any blocked threads:

parallel stacks

Note that as `CancelableStream` is used exclusively by `PeekableStreamReader`, it could be made `private`. This would enable:
* It to be made read only (i.e. `CancelableReadStream`) and `{Flush, FlushAsync, Write, WriteAsync, SetLength}` made to throw.
* The simplification of the user supplied cancellation token support as `PeekableStreamReader` provides `None` to `ReadAsync`.

Please let me know your thoughts?

### DIFF
--- a/src/KubernetesClient/WatcherDelegatingHandler.cs
+++ b/src/KubernetesClient/WatcherDelegatingHandler.cs
@@ -105,16 +105,16 @@ namespace k8s
                 base.Dispose(disposing);
             }
 
-            private CancellationTokenSourceSlim CreateCancellationTokenSource(CancellationToken userCancellationToken)
+            private LinkedCancellationTokenSource CreateCancellationTokenSource(CancellationToken userCancellationToken)
             {
-                return new CancellationTokenSourceSlim(_cancellationToken, userCancellationToken);
+                return new LinkedCancellationTokenSource(_cancellationToken, userCancellationToken);
             }
 
-            private readonly struct CancellationTokenSourceSlim : IDisposable
+            private readonly struct LinkedCancellationTokenSource : IDisposable
             {
                 private readonly CancellationTokenSource _cancellationTokenSource;
 
-                public CancellationTokenSourceSlim(CancellationToken token1, CancellationToken token2)
+                public LinkedCancellationTokenSource(CancellationToken token1, CancellationToken token2)
                 {
                     if (token1.CanBeCanceled && token2.CanBeCanceled)
                     {
--- a/src/KubernetesClient/WatcherDelegatingHandler.cs
+++ b/src/KubernetesClient/WatcherDelegatingHandler.cs
@@ -47,8 +47,24 @@ namespace k8s
 
             public override void Flush() => _innerStream.Flush();
 
+            public override async Task FlushAsync(CancellationToken cancellationToken)
+            {
+                using (var cancellationTokenSource = CreateCancellationTokenSource(cancellationToken))
+                {
+                    await _innerStream.FlushAsync(cancellationTokenSource.Token).ConfigureAwait(false);
+                }
+            }
+
             public override int Read(byte[] buffer, int offset, int count) =>
                 _innerStream.ReadAsync(buffer, offset, count, _cancellationToken).GetAwaiter().GetResult();
+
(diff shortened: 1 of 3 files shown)


andykernahan:
Some builds are failing as the SDK is not selecting the correct default of C# 7.3 (as [_should_](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version#defaults) be the case for .NET Core 2.x):

|Build|OS|SDK|Default|
|-|-|-|-|
|GitHub|*|2.2.108|7.0|
|Travis|OSX|2.2.105|7.0|
|Travis|Linux|2.2.402|7.3|

Note that 2.1 is LTS for 2.x and 2.2 is [end of life](https://dotnet.microsoft.com/download/dotnet-core). I can create an issue for this?

I will specify the language version in the project to resolve the immediate issue.

tg123:
Thanks

I will take a look carefully this weekend.

yes we should target runtime netcoreapp2.1 for LTS 
or move to 3.1?

brendandburns:
I think we should move to 3.1

src/KubernetesClient/WatcherDelegatingHandler.cs
```diff
@@ -107,24 +172,18 @@ protected override bool TryComputeLength(out long length)
...
-            public override string ReadLine()
-            {
-                if (_buffer.Count > 0)
-                {
-                    return _buffer.Dequeue();
-                }
-
-                return base.ReadLine();
-            }
+            public override string ReadLine() => throw new NotImplementedException();
```
brendandburns: Shouldn't we leave this implementation?

andykernahan: The method is effectively unused now. It was made to throw for consistency with the other unused members.

andykernahan:
> I think we should move to 3.1

Would you prefer I create separate PR for that? In addition to the SDK bump, the latest 2.0 & 2.1 runtimes must be installed so the tests run.

brendandburns:
separate PR would be great.

PR #402, opened 2020-04-06
Allow token refresh for GCP

macsux:


tg123:
IMHO, it is better to have token refresh via rest api than via cli

macsux:
While I agree, you would need to setup a whole oauth2 authentication
mechanism from scratch. The current method is to read token from kubectl
config. This works well for local runs when you're using current context of
kubectl, and not appropriate for service deployments. To setup standalone
auth, you will need to provide clientid, secret and manage token lifecycle
manually. I think both should be supported - kubectl for local and rest
when you're doing actual development.

As a side note, current implementation of local kubectl config is deeply
flawed for all providers. It only reads current access token when
kubernetes object is constructed. Once that token is expired, the app will
stop worrying.

macsux:
As we're supporting reading tokens out of kubectl config, there's limited information available there. There's no way to gain access to clientid/secret/refreshtoken using current approach. Given that it is still desirable to be able to allow apps to authenticate using creds found in current kubectl context, I propose the current authenticator be renamed to something like KubectlGcp, and another one built to support token management via REST for when you want to configure authentication explicitly.

src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
```diff
@@ -345,23 +346,9 @@ private void SetUserDetails(K8SConfiguration k8SConfig, Context activeContext)
...
-                                                                 , DateTimeOffset.Now)
-                                         <= 0)
-                                        {
-                                            throw new KubeConfigException("Refresh not supported.");
-                                        }
-                                    }
-                                }
-
-                                AccessToken = config["access-token"];
+                                TokenProvider = new GcpTokenProvider(config["cmd-path"]);
```
brendandburns, line 351: I don't know the expected details of a GCP kubeconfig, but is this field present in all configs?

Are the GCP kubeconfig contents documented anywhere?

src/KubernetesClient/Authentication/GcpTokenProvider.cs
```diff
@@ -0,0 +1,57 @@
...
+        public Task<AuthenticationHeaderValue> GetAuthenticationHeaderAsync(CancellationToken cancellationToken)
+        {
+            if (DateTime.UtcNow.AddSeconds(30) > _expiry)
+            {
+                RefreshToken();
+            }
+            return Task.FromResult(new AuthenticationHeaderValue("Bearer", _token));
+        }
+
+        private void RefreshToken()
```
brendandburns, line 32: Shouldn't this be async and use an async command line runner?

src/KubernetesClient/Authentication/GcpTokenProvider.cs
```diff
@@ -0,0 +1,57 @@
...
+using System.Net.Http.Headers;
+using System.Threading;
+using System.Threading.Tasks;
+using k8s.Exceptions;
+using Microsoft.Rest;
+using Newtonsoft.Json.Linq;
+
+namespace k8s.Authentication
+{
+    public class GcpTokenProvider : ITokenProvider
```
brendandburns, line 12: This file needs a unit test.

src/KubernetesClient/KubernetesClientConfiguration.cs
```diff
@@ -76,5 +77,7 @@ public partial class KubernetesClientConfiguration
         /// </summary>
         /// <value>The access token.</value>
         public string AccessToken { get; set; }
```
brendandburns, line 79: I feel like instead of both having an AccessToken field and a TokenProvider field, we should just switch to TokenProvider everywhere and use a `StaticTokenProvider` or somesuch for the static token case.

src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
macsux: I don't believe google "documents" it as it automatically configures context for you, but it is the way they do it. To be safe I can make this behavior apply only when the specified field is present

src/KubernetesClient/KubernetesClientConfiguration.cs
macsux: The issue atm is tokens actually expire. This is very noticeable with GCP provider which issues short lived tokens - KubernetesClient stops working after about 10 minutes. After that token is done, you need to interact with kubectl to force it to get a new token, and re-read it by recreating KubernetesClient.

src/KubernetesClient/Authentication/GcpTokenProvider.cs
macsux: Switched to non blocking command line runner

src/KubernetesClient/Authentication/GcpTokenProvider.cs
macsux: How would you expect the unit test to look? This class is 90% call to an external process which a) doesn't exist in unit test environment
b) requires to be authenticated / configured in order to respond correctly

I can certainly try to create a shell script to mock of what gcloud cli returns, though I question the value of such test. Beyond invoking external command, the rest of this class is just two lines of json querying.

src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
brendandburns: Yeah, I think I prefer that.

src/KubernetesClient/KubernetesClientConfiguration.cs
brendandburns: Yeah, I get the problem, what I'm suggesting is that we remove the `AccessToken` field in favor of a `TokenProvider` everywhere.

src/KubernetesClient/Authentication/GcpTokenProvider.cs
brendandburns: There's sufficient logic (e.g. return code != 0 -> failure, etc) that I think it deserves a test.

There's an equivalent test for the Javascript client here:

https://github.com/kubernetes-client/javascript/blob/master/src/exec_auth_test.ts

src/KubernetesClient/KubernetesClientConfiguration.cs
macsux: I'm trying to do this for AKS, but the cluster that I just spun up seems to use certs for authentication, as I got client-key-data, client-certificate-data, and token in my config file. The current code seems to be written where there's access_token. Under what scenario would this be so?

I feel like if we're supporting kubectl current context authentication, then we need to support all different types of authentication that kubectl uses, which includes
- OIDC - discover endpoints automatically
- external CLI to get token (like gcp)
- hardcoded / static (aks seems to use static token + cert)
- Oauth2 - hardcoded endpoints

Would be good to see a sample of what AWS EKS config looks like, I don't have one.

macsux: Seems like this is where all the different built in authenticators are defined
https://github.com/kubernetes/client-go/tree/master/plugin/pkg/client/auth

src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
macsux: I found where this is coming from and in case of "gcp" it will always be present as it's hardcoded behavior of official go client. https://github.com/kubernetes/client-go/blob/master/plugin/pkg/client/auth/gcp/gcp.go

Imo no reason to do the extra check

src/KubernetesClient/KubernetesClientConfiguration.cs
macsux: Based on my research, it seems there are 3 out of the box authentication providers that kubectl uses: gcp, azure & oidc, and each of these would translate to their own unique token provider. They are as listed in the above URL. These act very much like HttpMessageHandlers. The selection of the provider is dependent on 
```
users:
  user:
    auth-provider:
      name: XXX
```

My implementation for gcp is actually correct. Ideally, we want to have separate implementation for the other two types of providers. I propose that is done in a separate PR.

I won't be able to code for azure plugin because I'm lacking an AKS cluster that is integrated with Azure AD. I can't set this up as you need to be Azure tenant admin to configure this, and I don't have those rights in my org. If someone wants to provide me with AKS cluster that is enabled for Azure AD integration, I would be happy to create that token authenticator. The instructions on how to set this up are here: https://docs.microsoft.com/en-us/azure/aks/azure-ad-integration

src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
brendandburns: Sounds good, can you add a pointer to this "documentation" in a comment

e.g. 

``// docs: https://github.com/kubernetes/client-go/blob/master/plugin/pkg/client/auth/gcp/gcp.go#L58`

brendandburns:
I'm ok with the `TokenProvider` refactor in a different PR. Please add a unit test for the GCP Token Provider and I think this is ready to merge.

tests/KubernetesClient.Tests/Authentication/GcpTokenProvider.cs
```diff
@@ -0,0 +1,21 @@
...
+using System.IO;
+using k8s.Authentication;
+using Xunit;
+
+namespace k8s.tests.Authentication
+{
+    public class GcpTokenProviderTest
+    {
+        [Fact]
+        public void AcquireRefreshToken()
```
brendandburns, line 11: What is this testing? I don't see any asserts.

tests/KubernetesClient.Tests/OperatingSystem.cs
```diff
@@ -0,0 +1,12 @@
+using System;
+
+namespace k8s.Tests
+{
+    [Flags]
+    public enum OperatingSystem
```
brendandburns, line 6: why not inline this in OperatingSystemFact.cs? I don't think it's used anywhere else.

brendandburns: oh, I take it back, I see it in the test too. nevermind.

brendandburns:
Looks good, but I think there is a stray unneeded test cases in there?

macsux:
> I think there is a stray unneeded test cases in there
What do you mean?

brendandburns:
`tests/KubernetesClient.Tests/Authentication/GcpTokenProvider.cs` doesn't appear to actually be testing anything? There are no asserts in that file, it just creates the token provider.

macsux:
Accidental checked in file - removed

brendandburns:
@macsux I resolved some merge/rebase conflicts, can you double check the code looks right and then this is ready to merge.

Thanks for the patience.

brendandburns:
Looks like there are some formatting fixes needed. Please run `dotnet format --exclude src/KubernetesClient/generated`

Issue #397, opened 2020-04-02
MuxedStream for STDIN does not close itself properly


admilazz:
When a MuxedStream representing STDIN is closed it should send a message indicating this to the remote side. As a simple example, if you execute /bin/cat on a pod, send some data over STDIN, and dispose the STDIN stream, the command will never complete because the closure of the stream is not communicated to the server, so cat waits forever for STDIN to be closed.

admilazz:
It should suffice to execute this:

    socket.CloseOutputAsync(WebSocketCloseStatus.NormalClosure, null, cancelToken)

when the STDIN stream is closed.

admilazz:
Turns out that's not quite sufficient, since it intermittently truncates the data. I think the RunLoop method in StreamDemuxer.cs must also be fixed to wait for a full closure...

admilazz:
In the SPDY protocol (which kubectl exec uses), the closing of STDIN is signaled by a zero-byte data frame with the FIN bit set, and this I've verified from a hex dump of the network traffic.

Presumably you do the same thing for web sockets. I've tried sending an empty frame with the finish flag...
````await socket.SendAsync(new byte[] { (byte)ChannelIndex.StdIn }, WebSocketMessageType.Binary, true, cancelToken)````
but it doesn't seem to close the stream. (Or if it does close the stream, the client remains blocked in webSocket.ReceiveAsync.)

I've tried doing that and then following it up with a second empty message with the finish flag, thinking that maybe it's looking for an empty message. No dice.

I've also tried sending a zero-byte message, without the channel index. 
````await socket.SendAsync(new byte[0], WebSocketMessageType.Binary, true, cancelToken)````
Same thing. StreamDemuxer remains blocked on ReceiveAsync.

The only thing that consistently ends the command is to call CloseOutputAsync:
```await socket.CloseOutputAsync(WebSocketCloseStatus.NormalClosure, null, cancelToken)```
i.e. to do a "half close" of the connection. But the server doesn't seem to interpret that as "finish sending everything and then close". It seems to interpret it as "close ASAP", which may truncate the data being sent back. Or perhaps it's the client in StreamDemuxer that's not doing things right.

It's all made so difficult by the lack of any Kubernetes websocket protocol documentation that I could find.

admilazz:
[...]

PR #406, opened 2020-04-08
Add a fluent Kubernetes API

admilazz:
This change adds a fluent API that is concise, flexible, and works well with custom resources and custom actions. The basic flow is that from the client, you can create a request, and given a request you can modify or execute it. If `c` is a Kubernetes client, then these expressions create different types of requests:

    c.New<V1Pod>() // list all pods in all namespaces
    c.New<V1Pod>("ns") // list all pods in a namespace
    c.New<V1Pod>("ns", "name") // get a single pod
    c.New<V1Pod>("ns", "name").Delete() // delete a pod by name
    c.New(pod).Post() // create a pod
    c.New(pod, setBody: false).Delete() // delete a pod if you have the object already
    c.New<V1Pod>("ns").LabelSelector("X").FieldSelector("Y") // list pods with label and field selectors
    c.New(pod).Status().Put() // update a pod's status
    c.New(pod).Subresource("log") // read a pod's logs
    c.New(pod).Put().DryRun(true) // simulate replacing a resource
    c.New().RawUri("/apis/…") // specify a raw URL
    etc...

It works with both built-in and custom resource type. For advanced operations, you can:

* Use custom resource types and HTTP verbs
* Modify request and content headers
* Modify the query string
* Access subresources
* Stream the request or upgrade it for web socket or SPDY connections -- I can send a PR for exec over SPDY later (addressing issue #397 )
* Stream the response and access the raw network connection
* Specify group, version, and kind manually, or specify the entire raw URL
* Configure watches -- I can send a PR for an improved watch type later

To execute any request so constructed, it's simply: `req.ExecuteAsync()`. This returns a response object that gives you full access to the status code, response headers, and response stream. (It does not throw an exception if an HTTP error response is returned.) It also has convenience methods for reading and deserializing the response in various ways.

A request can be executed multiple times, including in parallel. (I.e. a request is not modified or consumed by executing it.) You can clone a request if you want to write a method that takes a request and executes a modified version of it without mutating the original.

If you want to execute a request, check the status (including throwing exceptions for errors), and deserialize the body all in one go, it's simply `req.ExecuteAsync<T>()`.

To assist with modifying resources, it supports an atomic get-modify-update operation via `req.ReplaceAsync<T>(...)`.

As mentioned above, I also have improved watching and exec implemented (over both SPDY and web sockets), and I can send PRs for those later after handling legal and other issues.

**NOTE: This PR builds on and includes changes from two other PRs (https://github.com/kubernetes-client/csharp/pull/404 and https://github.com/kubernetes-client/csharp/pull/405).** I'm not sure how best to represent that in a PR, but the files actually added or changed in this specific PR are:

* Kubernetes.ConfigInit.cs
* Kubernetes.Fluent.cs
* Kubernetes.WebSocket.cs
* KubernetesRequest.cs
* Scheme.cs
* Kubernetes.Fluent.Tests.cs

### DIFF
--- a/src/KubernetesClient/Kubernetes.ConfigInit.cs
+++ b/src/KubernetesClient/Kubernetes.ConfigInit.cs
@@ -203,19 +203,7 @@ namespace k8s
         /// </summary>
         private void SetCredentials()
         {
-            // set the Credentails for token based auth
-            if (!string.IsNullOrWhiteSpace(config.AccessToken))
-            {
-                Credentials = new TokenCredentials(config.AccessToken);
-            }
-            else if (!string.IsNullOrWhiteSpace(config.Username) && !string.IsNullOrWhiteSpace(config.Password))
-            {
-                Credentials = new BasicAuthenticationCredentials
-                {
-                    UserName = config.Username,
-                    Password = config.Password
-                };
-            }
+            Credentials = CreateCredentials(config);
         }
 
         internal readonly KubernetesClientConfiguration config;
@@ -283,6 +271,23 @@ namespace k8s
             var settings = new JsonSerializerSettings() { NullValueHandling = NullValueHandling.Ignore };
             settings.Converters.Add(new Newtonsoft.Json.Converters.StringEnumConverter());
             return settings;
+        }
+
+        /// <summary>Creates <see cref="ServiceClientCredentials"/> from a Kubernetes configuration, or returns null if the configuration
+        /// contains no credentials of that type.
+        /// </summary>
+        internal static ServiceClientCredentials CreateCredentials(KubernetesClientConfiguration config)
+        {
+            if(config == null) throw new ArgumentNullException(nameof(config));
+            if(!string.IsNullOrEmpty(config.AccessToken))
+            {
+                return new TokenCredentials(config.AccessToken);
+            }
+            else if(!string.IsNullOrEmpty(config.Username))
+            {
+                return new BasicAuthenticationCredentials() { UserName = config.Username, Password = config.Password };
+            }
+            return null;
         }
 
         /// <summary>Gets the <see cref="JsonSerializerSettings"/> used to serialize and deserialize Kubernetes objects.</summary>
--- a/src/KubernetesClient/Kubernetes.ConfigInit.cs
+++ b/src/KubernetesClient/Kubernetes.ConfigInit.cs
@@ -72,7 +72,7 @@ namespace k8s
             get => _scheme;
             set
             {
-                if(value == null) throw new ArgumentNullException(nameof(Scheme));
+                if (value == null) throw new ArgumentNullException(nameof(Scheme));
                 _scheme = value;
             }
         }
@@ -278,12 +278,12 @@ namespace k8s
         /// </summary>
         internal static ServiceClientCredentials CreateCredentials(KubernetesClientConfiguration config)
         {
-            if(config == null) throw new ArgumentNullException(nameof(config));
-            if(!string.IsNullOrEmpty(config.AccessToken))
+            if (config == null) throw new ArgumentNullException(nameof(config));
+            if (!string.IsNullOrEmpty(config.AccessToken))
             {
                 return new TokenCredentials(config.AccessToken);
             }
-            else if(!string.IsNullOrEmpty(config.Username))
+            else if (!string.IsNullOrEmpty(config.Username))
             {
                 return new BasicAuthenticationCredentials() { UserName = config.Username, Password = config.Password };
             }
--- a/src/KubernetesClient/Kubernetes.ConfigInit.cs
+++ b/src/KubernetesClient/Kubernetes.ConfigInit.cs
@@ -8,6 +8,7 @@ using System.Security.Cryptography.X509Certificates;
 using k8s.Exceptions;
 using k8s.Models;
 using Microsoft.Rest;
+using Newtonsoft.Json;
 
 namespace k8s
 {
@@ -41,9 +42,8 @@ namespace k8s
         public Kubernetes(KubernetesClientConfiguration config, HttpClient httpClient, bool disposeHttpClient) : this(httpClient, disposeHttpClient)
         {
             ValidateConfig(config);
-            CaCerts = config.SslCaCerts;
-            SkipTlsVerify = config.SkipTlsVerify;
-            SetCredentials(config); 
+            this.config = config;
+            SetCredentials();
         }
 
         /// <summary>
@@ -59,10 +59,24 @@ namespace k8s
             : this(handlers)
         {
             ValidateConfig(config);
-            CaCerts = config.SslCaCerts;
-            SkipTlsVerify = config.SkipTlsVerify;
-            InitializeFromConfig(config);
+            this.config = config;
+            InitializeFromConfig();
         }
+
+        /// <summary>Gets or sets the <see cref="KubernetesScheme"/> used to map types to their Kubernetes groups, versions, and kinds.
+        /// The default is <see cref="KubernetesScheme.Default"/>.
+        /// </summary>
+        /// <summary>Gets or sets the <see cref="KubernetesScheme"/> used to map types to their Kubernetes groups, version, and kinds.</summary>
+        public KubernetesScheme Scheme
+        {
+            get => _scheme;
+            set
+            {
+                if(value == null) throw new ArgumentNullException(nameof(Scheme));
+                _scheme = value;
+            }
+        }
+
 
         private void ValidateConfig(KubernetesClientConfiguration config)
         {
@@ -86,7 +100,7 @@ namespace k8s
             }
         }
 
-        private void InitializeFromConfig(KubernetesClientConfiguration config)
+        private void InitializeFromConfig()
         {
             if (BaseUri.Scheme == "https")
             {
@@ -107,25 +121,25 @@ namespace k8s
                 }
                 else
                 {
-                    if (CaCerts == null)
+                    if (config.SslCaCerts == null)
                     {
                         throw new KubeConfigException("A CA must be set when SkipTlsVerify === false");
                     }
 #if NET452
                     ((WebRequestHandler) HttpClientHandler).ServerCertificateValidationCallback = (sender, certificate, chain, sslPolicyErrors) =>
                     {
-                        return Kubernetes.CertificateValidationCallBack(sender, CaCerts, certificate, chain, sslPolicyErrors);
+                        return Kubernetes.CertificateValidationCallBack(sender, config.SslCaCerts, certificate, chain, sslPolicyErrors);
                     };
 #elif XAMARINIOS1_0
                     System.Net.ServicePointManager.ServerCertificateValidationCallback += (sender, certificate, chain, sslPolicyErrors) =>
                     {
                         var cert = new X509Certificate2(certificate);
-                        return Kubernetes.CertificateValidationCallBack(sender, CaCerts, cert, chain, sslPolicyErrors);
+                        return Kubernetes.CertificateValidationCallBack(sender, config.SslCaCerts, cert, chain, sslPolicyErrors);
                     };
 #elif MONOANDROID8_1
                     var certList = new System.Collections.Generic.List<Java.Security.Cert.Certificate>();
 
-                    foreach (X509Certificate2 caCert in CaCerts)
+                    foreach (X509Certificate2 caCert in config.SslCaCerts)
                     {
                         using (var certStream = new System.IO.MemoryStream(caCert.RawData))
                         {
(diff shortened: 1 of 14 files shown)


admilazz:
I'll add some unit tests as well.

macsux:
This is similar to https://github.com/tintoy/dotnet-kube-client. I personally prefer that approach vs the autogenerated one we have right now as part of official SDK, because autogenerated stuff is too verbose and doesn't properly cover CRDs. I think we need buy-in from maintainers to move into that direction. If there is a buy-in, then I think we need to have a longer discussion about what the API syntax should look like (potentially compare to what the KubeClient does), as they tackle some extra scenarios. For example, as quoted off their page

> Some operations in the Kubernetes API can return a different response depending on the arguments passed in. For example, a request to delete a v1/Pod returns the existing v1/Pod (as a PodV1 model) if the caller specifies DeletePropagationPolicy.Foreground but returns a v1/Status (as a StatusV1 model) if any other type of DeletePropagationPolicy is specified.

> To handle this type of polymorphic response KubeClient uses the KubeResultV1 model (and its derived implementations, KubeResourceResultV1<TResource> and KubeResourceListResultV1<TResource>).

> KubeResourceResultV1<TResource> can be implicitly cast to a TResource or a StatusV1, so consuming code can continue to use the client as if it expects an operation to return only a resource or expects it to return only a StatusV1:

As far as the proposed syntax in this PR, I would prefer to see List/Get type operations to be explicit rather than part of `.New<T>` which doesn't really feel intuitive.

admilazz:
Hi, macsux. I appreciate the comment.

Taking a look now at dotnet-kube-client, I think it seems completely different. Their client appears to be a strongly typed client that works with specific Kubernetes types. For example, in their example you do `client.PodsV1().Delete(…)`. That's not to say it's better or worse, but that model requires the generation (manually or otherwise) of lots of code to add all the operations for all the different Kubernetes resource types, and then there's the question of whether their client can work for custom resources, custom operations, etc.

By contrast, the fluent client I'm proposing is completely generic. It has zero knowledge about pods, for example. It just works with any type and any operation. That means it can be very small, requiring no code generation, and very flexible. But the user has to know how to use it correctly.

> Some operations in the Kubernetes API can return a different response depending on the arguments passed in. For example, a request to delete a v1/Pod returns the existing v1/Pod (as a PodV1 model) if the caller specifies DeletePropagationPolicy.Foreground but returns a v1/Status (as a StatusV1 model) if any other type of DeletePropagationPolicy is specified.

This fluent library can handle that just fine. E.g. `r.ExecuteAsync<V1Pod>()` versus `r.ExecuteAsync<V1Status>()`. Or, in the unlikely case that you don't know which kind of delete request you're sending, you can `r.ExecuteAsync()` and figure out it out dynamically.

> As far as the proposed syntax in this PR, I would prefer to see List/Get type operations to be explicit rather than part of .New<T> which doesn't really feel intuitive.

That starts restraining the possibilities and making assumptions, though. List and Get are both GET requests, and they differ only in the type of object returned. Imagine a resource with a subresource. If that subresource is called "backups", then you would expect GETting it to return a list of backups. But if the subresource is called "backupStatus" then you'd expect it to return a single object. The URL has exactly the same form (with only "backups" replaced by "backupStatus"). That is to say, whether a request is a list request cannot be determined in general from the request itself. It requires external knowledge. Now, it can be done for built-in types because you can make assumptions about how built-in Kubernetes resources work, but once you get into custom resources you can no longer make those assumptions. I think that having a single, general-purpose interface that works the same way for all types is better than adding methods like List() which may or may not work depending on the type of resource being accessed. Or, if List() and Get() are just synonyms and do exactly the same thing, then List() would add nothing but eye candy...

macsux:
Just heads up, you may want to integrate the label selector builder. I have it coded specifically for informers pieces that are part of a different active PR, but it seems like your code can leverage same pattern:

https://github.com/macsux/kubernetes-client-csharp/blob/feature/rx-informers/src/KubernetesClient/Informers/KubernetesInformerOptions.cs

https://github.com/macsux/kubernetes-client-csharp/blob/feature/rx-informers/src/KubernetesClient/Informers/KubernetesInformerOptionsBuilder.cs

https://github.com/macsux/kubernetes-client-csharp/blob/feature/rx-informers/tests/KubernetesClient.Tests/KubernetesInformerOptionsBuilderTests.cs

src/KubernetesClient/KubernetesRequest.cs
```diff
@@ -0,0 +1,769 @@
...
+            // add the headers
+            if(!string.IsNullOrEmpty(config?.AccessToken))
+            {
+                req.Headers.Authorization = new AuthenticationHeaderValue("Bearer", config.AccessToken);
+            }
+            else if(!string.IsNullOrEmpty(config?.Username))
+            {
+                req.Headers.Authorization = new AuthenticationHeaderValue("Basic",
+                    Convert.ToBase64String(Encoding.UTF8.GetBytes(config.Username + ":" + config.Password)));
+            }
```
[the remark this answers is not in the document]
admilazz: I think currently it's not handled by message handlers attached to the HttpClient even in the normal Kubernetes client. But it does construct a Credentials object rather than setting the header manually (although the effect is ultimately the same). I think I'll change this to use the same Credentials object that the client has already constructed.

I saw your PR for refreshing the access token, by the way. Seems good and important to have.

admilazz:
Regarding the label selector builder, that seems like the kind of thing that can easily be added (possibly via extension methods):

    r.LabelSelectorBuilder().HasLabel("x").LabelEquals("a", "b").Apply()

or, perhaps simpler and cleaner...

    r.LabelSelectorBuilder(b => b.HasLabel("x").LabelEquals("a", "b"))

It seems like the kind of thing I'd want to support if it already existed in the library, but if it doesn't already exist I'd like to try to keep this PR as slim as possible, at least until I see whether it gains any traction. (Also, if I added a label selector builder I'd also feel compelled to add a field selector builder and probably something around field managers, for consistency. :-P)

src/KubernetesClient/KubernetesRequest.cs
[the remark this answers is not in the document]
admilazz: I changed this to

    if(credentials != null) await credentials.ProcessHttpRequestAsync(req, cancelToken);

where credentials is taken from the Kubernetes object or supplied by the user. It'll be in the next iteration.

src/KubernetesClient/Kubernetes.ConfigInit.cs
```diff
@@ -41,9 +42,8 @@ public Kubernetes(KubernetesClientConfiguration config, HttpClient httpClient) :
         public Kubernetes(KubernetesClientConfiguration config, HttpClient httpClient, bool disposeHttpClient) : this(httpClient, disposeHttpClient)
         {
             ValidateConfig(config);
-            CaCerts = config.SslCaCerts;
-            SkipTlsVerify = config.SkipTlsVerify;
```
brendandburns, line 45: It looks like this is getting dropped?

admilazz: The config object is saved below in `this.config = config`, and they are accessible from there.

src/KubernetesClient/KubernetesRequest.cs
```diff
@@ -0,0 +1,769 @@
+using System;
```
brendandburns, line 1: This file needs formatting. in particular, space between `if` and `(...)`

admilazz: I assume that applies to `for` and `while` as well. I'll go reformat the whole thing if people think the change is worth having. :-)  Just let me know what changes are needed.

brendandburns:
fwiw, I'm not opposed to this client, but the group/version/kind info isn't something that developers typically think about when they create objects programattically.

e.g. when you do `new V1Pod(...)` the GVK isn't populated so you're going to have a bunch of developers who need to do the boilerplate:

```
var pod = new V1Pod();
pod.Group = "";
pod.Version = "v1";
pod.Kind = "Pod";
```

admilazz:
> fwiw, I'm not opposed to this client, but the group/version/kind info isn't something that developers typically think about when they create objects programattically.
> e.g. when you do new V1Pod(...) the GVK isn't populated so you're going to have a bunch of developers who need to do the boilerplate:
> var pod = new V1Pod();
> pod.Group = "";
> pod.Version = "v1";
> pod.Kind = "Pod";

True. In my own code, I have a simple generic helper function that instantiates a type and fills out those fields. I could add something like that here as well.

P.S. What do you think of New() for the request as opposed to, say, Request(). (When I originally wrote this, it was its own class and not a partial Kubernetes class, but now it has to share space with many other methods, so a more specific name may be better.)

src/KubernetesClient/Kubernetes.ConfigInit.cs
tg123: any detail reason for this change?

tg123:
After looking into the PR, 
My understanding is that this is something same as
`Azure management sdk Fluent` vs `Azure management sdk`

I suggest start a new nuget package `Kubenetes.Fluent` which depends on current SDK
The same way as Azure SDK.


The reason is that a SDK providing 2 different ways to use is really confused.
The sharing compoments are models only. 
In additional, fluent use a different httpclient from `autorest` which also contributes to inconsistency.

src/KubernetesClient/KubernetesRequest.cs
```diff
@@ -0,0 +1,777 @@
...
+using System.Threading.Tasks;
+using k8s.Models;
+using Microsoft.Rest;
+using Newtonsoft.Json;
+
+namespace k8s
+{
+    #region KubernetesRequest
+    /// <summary>Represents a single request to Kubernetes.</summary>
+    public sealed class KubernetesRequest : ICloneable
```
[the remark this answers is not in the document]
admilazz: What do you mean?

src/KubernetesClient/Kubernetes.ConfigInit.cs
admilazz: I changed it to save the entire config object rather than just two fields out of it, because it was useful to have it when instantiating a request - in particular to construct the credentials. Now that I use the ServiceCredentials object from the Kubernetes client, this can probably be reverted. But it's not like any information is being lost here. :-)

admilazz:
> I suggest start a new nuget package Kubenetes.Fluent which depends on current SDK
> The sharing compoments are models only.

Some significant code to construct credentials and manage SSL certificates is shared as well. In order to put this in a separate library, that code would have to be made public or else it would have to be duplicated into the new library, or else the fluent API would become a little less flexible. I'm not opposed to putting this in a separate library - and I probably will do that if it's not accepted here - but I am somewhat opposed to copying the code for credentials and SSL certificate handling.

It's not simply a different way to do the same thing, however. It does add some significant new capabilities. But I agree that there's a lot of overlap too.

> In additional, fluent use a different httpclient from autorest which also contributes to inconsistency.

Actually, it uses the same HttpClient as the Autorest client does.

src/KubernetesClient/KubernetesRequest.cs
```diff
@@ -0,0 +1,777 @@
...
+            return name;
+        }
+
+        static string NormalizeEmpty(string value) => string.IsNullOrEmpty(value) ? null : value; // normalizes empty strings to null
+    }
+#endregion
+
+#region KubernetesResponse
+    /// <summary>Represents a response to a <see cref="KubernetesRequest"/>.</summary>
+    public sealed class KubernetesResponse : IDisposable
```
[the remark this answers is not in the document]
tg123: move to a new file KubernetesResponse.cs?

brendandburns:
I agree with @tg123 that this should be a different package (it doesn't necessarily have to be a different Nuget, even just a different namespace seems ok to me)

Can you see how invasive it is to make the necessary SSL stuff public, I feel like something like an `Kubernetes.authorizeRequest(<request-object-interface>)` should work pretty decently as a public interface.

If it seems really terrible to make that stuff public, moving this into a different class `FluentKubernetes` is a reasonable compromise.

src/KubernetesClient/KubernetesRequest.cs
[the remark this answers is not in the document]
admilazz: Okay, I moved it.

admilazz:
> I agree with @tg123 that this should be a different package (it doesn't necessarily have to be a different Nuget, even just a different namespace seems ok to me)

Okay, I'll put it in a new namespace and use extension methods rather than a partial class.

> Can you see how invasive it is to make the necessary SSL stuff public

Well if it's just in a new namespace it can be internal. :-)

P.S. The blocking PR at the moment is the "convenient model extensions" one.

macsux:
> fwiw, I'm not opposed to this client, but the group/version/kind info isn't something that developers typically think about when they create objects programattically.

There is a new extension method that I've added to address this as part of other PR
https://github.com/macsux/kubernetes-client-csharp/blob/feature/crd/src/KubernetesClient/Extensions.cs#L31-L40

APIVersion and Kind should really be thought of as part of serialization type metadata then the actual payload.

macsux:
> Some significant code to construct credentials and manage SSL certificates is shared as well. In order to put this in a separate library, that code would have to be made public or else it would have to be duplicated into the new library, or else the fluent API would become a little less flexible. I'm not opposed to putting this in a separate library - and I probably will do that if it's not accepted here - but I am somewhat opposed to copying the code for credentials and SSL certificate handling.

Just create a new project and make it a friend of the main one, same way we did it for tests in https://github.com/kubernetes-client/csharp/pull/421

admilazz:
I guess I'll put this in a separate project, then.

macsux:
When I meant separate project I meant within the same SDK solution - moving towards a modular approach, not a separate GitHub project. Would love to see this code as part of main SDK

PR #408, opened 2020-04-10
Add an improved watch

admilazz:
This change adds improved watch support. There are some problems with the existing watch support.

* The design contains a pitfall that will trap users. It has public events like OnEvent but begins listening immediately in the constructor. This means that the usual pattern of creating an object and attaching your event handler will cause a race condition and lead to lost events. The only real way to avoid this is to pass the event handlers at the time you instantiate the watch. Given that watches are normally instantiated from Watch* methods that do not force you to provide the event handler up-front, the behavior is "broken by default".
* It internally uses a cancellation token but does not use it in the place that matters most - in the blocking IO call to read from the stream.
* If you dispose the watch while it's running, it often reports a spurious ObjectDisposedException.
* The watch offers only the lowest level support - reporting a stream of watch events from a single watch request. In practice, watches time out and need to be reopened, and when you reopen them you need to resume where you left off, which requires tracking the latest version, but if you're watching a list of items as opposed to a single item you can't easily do that because the correct version to resume at is not marked in the watch stream itself and must be retrieved from a list request. Support for resuming and for watching multiple items must be built on top, but the watch object itself doesn't have a very good interface for building on top of.
* It deserializes every event twice, which is unnecessary.

This change adds two main classes.

* WatchReader\<T\> -- This class functions similarly to the current Watch\<T\> but fixes the above problems. (I didn't want to make changes to Watch\<T\> itself in case anyone depends on its particular behavior.) It has a pull interface rather than an event-based interface:
      
      // x can be a Stream, HttpResponseMessage, or KubernetesResponse
      using(var wr = new WatchReader<V1Pod>(x))
      {
          WatchEvent<V1Pod> e;
          while((e = await wr.ReadAsync(cancelToken)) != null) /* handle event */
      }
      
* Watch\<T\> -- This class implements a high-level watch that tracks versions, resumes after disconnection, supports watch bookmarks, and can watch both lists and single items. It uses an event-based approach like the current Watcher\<T\>, but has more events:
  * Closed
  * EventReceived - Only raised for Added, Modified, and Removed events
  * Error
  * InitialListSent - Raised after the initial list of items was sent from opening the watch. This can be used to let a watch-based cache know when it has reached a synchronized state, for example.
  * Opened
  * Reset - Raised when the watch is opened without successfully resuming, i.e. when events may have been lost. This can be used to clear caches based on the watch, for example.
  
  ````    
  // request is a KubernetesRequest
  var w = new Watch<V1Pod>(request); // or use request.ToWatch<T>()
  w.EventReceived += ...;
  w.Run();
  …
  w.Dispose(); // or, give it a cancellation token and cancel it
  // if you need to know exactly when it stops you can await the task returned from Run()
  ````

**NOTE: This PR builds on and includes changes from two other PRs (https://github.com/kubernetes-client/csharp/pull/405 and https://github.com/kubernetes-client/csharp/pull/406).** I'm not sure how best to represent that in a PR, but the files actually added or changed in this specific PR are:

* Watch.cs
* Watcher.cs (trivial changes)
* KubernetesRequest.cs (minor changes)
* tests/KubernetesClient.Tests/FluentWatchTests.cs
* tests/KubernetesClient.Tests/PipeStream.cs

### DIFF
--- a/src/KubernetesClient/Watch.cs
+++ b/src/KubernetesClient/Watch.cs
@@ -440,7 +440,7 @@ namespace k8s
                 while(buffer.Length != 0 && (charRead != charWrite || await EnsureDataAsync().ConfigureAwait(false)))
                 {
                     int toRead = Math.Min(buffer.Length, charWrite-charRead);
-                    charBuffer.AsMemory(0, toRead).CopyTo(buffer);
+                    charBuffer.AsSpan(charRead, toRead).CopyTo(buffer.Span);
                     totalRead += toRead;
                     charRead += toRead;
                     if (!block) break;
@@ -449,17 +449,17 @@ namespace k8s
                 return totalRead;
             }
 #else
-            async Task<int> ReadAsync(char[] buffer, int index, int length, bool block)
+            async Task<int> ReadAsync(char[] buffer, int index, int count, bool block)
             {
                 int start = index;
-                while(length != 0 && (charRead != charWrite || await EnsureDataAsync().ConfigureAwait(false)))
+                while(count != 0 && (charRead != charWrite || await EnsureDataAsync().ConfigureAwait(false)))
                 {
-                    int toRead = Math.Min(length, charWrite-charRead);
-                    Array.Copy(charBuffer, 0, buffer, index, toRead);
+                    int toRead = Math.Min(count, charWrite-charRead);
+                    Array.Copy(charBuffer, charRead, buffer, index, toRead);
                     charRead += toRead;
                     index += toRead;
                     if (!block) break;
-                    length -= toRead;
+                    count -= toRead;
                 }
                 return index - start;
             }
--- a/src/KubernetesClient/Watch.cs
+++ b/src/KubernetesClient/Watch.cs
@@ -1,0 +1,451 @@
+using System;
+using System.IO;
+using System.Net;
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+using k8s.Models;
+using Newtonsoft.Json;
+
+namespace k8s
+{
+    #region Watch
+    /// <summary>Watches one or more Kubernetes objects, reopening the watch and managing versions as necessary.</summary>
+    public sealed class Watch<T> : IDisposable where T : IKubernetesObject, IMetadata<V1ObjectMeta>
+    {
+        /// <summary>Initializes a new watch from a <see cref="KubernetesRequest"/> that represents the item or list of items to watch.</summary>
+        /// <param name="request">A <see cref="KubernetesRequest"/> that selects the item or list of items to watch</param>
+        /// <param name="initialVersion">The resource version to resume from, or null to start from the current version</param>
+        /// <param name="isListWatch">The value of the <see cref="IsListWatch"/> property, or null to use the default</param>
+        public Watch(KubernetesRequest request, string initialVersion = null, bool? isListWatch = null)
+        {
+            this.req = request?.Clone() ?? throw new ArgumentNullException();
+            IsListWatch = isListWatch ?? req.Name() == null;
+            if (IsListWatch) req.AddQuery("allowWatchBookmarks", "true");
+            LastVersion = initialVersion;
+        }
+
+        /// <summary>Raised after the watch is closed.</summary>
+        public event Action<Watch<T>> Closed;
+
+        /// <summary>Raised when a non-bookmark watch event is received.</summary>
+        public event Action<Watch<T>, WatchEventType, T> EventReceived;
+
+        /// <summary>Raised when an error occurs, either an exception or a failure status.</summary>
+        public event Action<Watch<T>, Exception, V1Status> Error;
+
+        /// <summary>Raised after the initial list of item events has been sent when operating in list-watching mode (i.e.
+        /// when <see cref="IsListWatch"/> is true). If a resource version is passed to the constructor, the watch will attempt to
+        /// resume without obtaining an initial list, and this event may not be raised.
+        /// </summary>
+        public event Action<Watch<T>> InitialList;
+
+        /// <summary>Raised after the watch is opened.</summary>
+        public event Action<Watch<T>> Opened;
+
+        /// <summary>Raised after the watch is opened without a resource version or when events may have been lost.</summary>
+        public event Action<Watch<T>> Reset;
+
+        /// <summary>Gets or sets whether the watch should use list-watching semantics instead of item-watching semantics. This must
+        /// be true when watching multiple items and should be false when watching a single item and must be false when the request does
+        /// not return a list. By default, the property is initialized to true if the request <see cref="KubernetesRequest.Name()"/> is
+        /// null, but if you're using a request with a null name that is nonetheless guaranteed to return only one item, you can improve
+        /// efficiency by setting this to false.
+        /// </summary>
+        public bool IsListWatch { get; set; }
+
+        /// <summary>Gets or sets the version of the most recently received resource, or null if no resource has been received or
+        /// the version information is known to be out of date.
+        /// </summary>
+        public string LastVersion { get; private set; }
+
+        /// <summary>Gets or sets how long the watch will wait between retries after an error. The default is 15 seconds.</summary>
+        public TimeSpan OpenRetryTime { get; set; } = TimeSpan.FromSeconds(15);
+
+        /// <summary>Stops the watch.</summary>
+        public void Dispose() => cts?.Cancel();
+
+        /// <summary>Runs the watch. To stop it, call <see cref="Dispose"/> or cancel the <see cref="CancellationToken"/>.</summary>
+        /// <param name="shutdownToken">A <see cref="CancellationToken"/> that can be used to stop the watch</param>
+        public Task Run(CancellationToken shutdownToken = default)
+        {
+            if (task != null) throw new InvalidOperationException();
+            shutdownToken.ThrowIfCancellationRequested();
+            cts = shutdownToken.CanBeCanceled ?
+                CancellationTokenSource.CreateLinkedTokenSource(shutdownToken) : new CancellationTokenSource();
+            task = Task.Run(Run);
+            return task;
+        }
+
+        async Task Run()
+        {
+            while(!cts.IsCancellationRequested)
+            {
+                bool wasOpen = false; // used to determine whether we should raise OnClosed
+                try
+                {
+                    // when watching multiple items when we don't know the starting version, we first have to retrieve the list and then
+                    // watch for changes. if we just started watching, we'd get an initial mishmash of resource versions and wouldn't be
+                    // able to know what the latest version is. when listing, the appropriate version is part of the list metadata
+                    KubernetesList<T> list = null;
+                    if (IsListWatch && LastVersion == null)
+                    {
+                        using (KubernetesResponse listResponse = await req.WatchVersion(null).ExecuteAsync(cts.Token).ConfigureAwait(false))
+                        {
+                            if (listResponse.IsError)
+                            {
+                                await ReportErrorAndWait(listResponse).ConfigureAwait(false);
+                                continue;
+                            }
+                            list = await listResponse.GetBodyAsync<KubernetesList<T>>(failIfEmpty: true).ConfigureAwait(false);
+                            LastVersion = list.Metadata?.ResourceVersion;
+                        }
+                    }
+
+                    // now start the watch
+                    using (KubernetesResponse response = await req.WatchVersion(LastVersion ?? "").ExecuteAsync(cts.Token).ConfigureAwait(false))
+                    {
+                        if (response.IsError) // if the watch request failed...
+                        {
+                            // if our resource version is too old, reset it. otherwise report the error
+                            if (response.StatusCode == HttpStatusCode.Gone) LastVersion = null; // NOTE: in practice this seems to be reported as a watch event...
+                            else await ReportErrorAndWait(response).ConfigureAwait(false);
+                        }
+                        else // if the response was successful, then the watch is open...
+                        {
+                            if ((LastVersion == null || list != null) && Reset != null) Reset(this); // if we may have lost events, let the client know
+                            if (Opened != null) Opened(this);
+                            wasOpen = true; // since it was opened, we have to call OnClosed later
+                            // if we started a new watch based on a list, send out the initial events from the list.
+                            // otherwise, the initial event(s) will be sent from the watch
+                            if (list != null)
+                            {
+                                if (EventReceived != null)
+                                {
+                                    // annoyingly, Kubernetes may exclude the API version and kind from items in lists and the list itself
+                                    // only has the API version (of the list). the kind is the kind of the list (e.g. PodList). 
+                                    string guessedApiVersion, guessedKind;
+                                    if (!req.Scheme().TryGetVK(typeof(T), out guessedApiVersion, out guessedKind)) // so try to get them from the scheme
+                                    {
+                                        guessedApiVersion = list.ApiVersion; // if it's missing, assume the API version is the same as the list's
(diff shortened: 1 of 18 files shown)


tg123:
nice job
but why not `IEnumerable<T, Event>` ?

admilazz:
> why not IEnumerable<T, Event> ?

In WatchReader\<T\>? I guess for a couple reasons: 1) a watch spends most of its time blocked waiting for IO, which makes it a good candidate for async, and 2) watches don't just report events about type T - they also report errors which are of type V1Status rather than T. (It could be IEnumerable\<WatchEvent\<T\>\> but point 1 still applies.)

It could be an IAsyncEnumerable\<WatchEvent\<T\>\>, but that's only available in C# 8.0 and later.

Perhaps finally, the expected and main user interface is the high-level one, Watch\<T\>, not the low-level WatchReader\<T\>.

src/KubernetesClient/Watch.cs
```diff
@@ -0,0 +1,482 @@
...
+        /// </summary>
+        public async Task<WatchEvent<T>> ReadAsync(CancellationToken cancelToken = default)
+        {
+            if (IsError) return null;
+            cancelToken.ThrowIfCancellationRequested();
+            if (reader == null) CreateReader(await Response.Content.ReadAsStreamAsync().ConfigureAwait(false));
+            try
+            {
+                streamReader.CancelToken = cancelToken;
+#if NETCOREAPP2_1
```
andykernahan, line 284: I think this should be `#if !NET452` it's only `net452` that uses v9.0.1.

src/KubernetesClient/Watch.cs
```diff
@@ -0,0 +1,482 @@
...
+
+            // deserialize the watch event
+            var e = new WatchEvent<T>();
+            bool gotType = false;
+            while (true)
+            {
+                reader.Read(); // move to the next property, if any
+                if (reader.TokenType != JsonToken.PropertyName) break;
+                string name = (string)reader.Value;
+                reader.Read(); // move to the property value
```
andykernahan, line 303: Is is safe to ignore the return value of `Read`? There's no guarantee the property value can be read and it would (bizarrely) result in an `ArgumentException`:
```csharp
var r = new JsonTextReader(new StringReader("{\"type\":"));
Read();
Read();
Read();
Enum.Parse(typeof(WatchEventType), (string)r.Value, true);
void Read() => Console.WriteLine($"Read={r.Read()}, {r.TokenType}={r.Value}");
// Read=True, StartObject=
// Read=True, PropertyName=type
// Read=False, PropertyName=type
// System.ArgumentException: Requested value 'type' was not found.
```

src/KubernetesClient/Watch.cs
```diff
@@ -0,0 +1,482 @@
...
+                if (name == "type")
+                {
+                    e.Type = (WatchEventType)Enum.Parse(typeof(WatchEventType), (string)reader.Value, true);
+                    gotType = true;
+                }
+                else if (name == "object")
+                {
+                    if (!gotType) throw new JsonSerializationException("Expected type property before object.");
+                    if (e.Type != WatchEventType.Error) e.Object = serializer.Deserialize<T>(reader);
```
andykernahan, line 312: If I have understood correctly: in the (not unlikely) case an event spans the 4 KiB buffer, then `Deserialize` will result in blocking IO.  Also, I believe there are pathological cases where a disproportionate amount of IO becomes blocking (e.g. burst of events slightly smaller than some multiple of the buffer size).

It would be great if this were made async all the way.

src/KubernetesClient/Watch.cs
```diff
@@ -0,0 +1,482 @@
...
+
+            protected override void Dispose(bool disposing)
+            {
+                base.Dispose(disposing);
+                stream?.Dispose();
+            }
+
+            bool EnsureData()
+            {
+                if (charRead == charWrite)
```
andykernahan, line 383: Consider delegating to `EnsureDataAsync` instead:
```csharp
if (charRead == charWrite)
{
  try { return EnsureDataAsync().GetAwaiter().GetResult(); }
  catch (OperationCanceledException) { return false; } // non-async callers aren't expecting OperationCanceledException
}
```

src/KubernetesClient/Watch.cs
```diff
@@ -0,0 +1,482 @@
...
+                        if (response.IsError) // if the watch request failed...
+                        {
+                            // if our resource version is too old, reset it. otherwise report the error
+                            if (response.StatusCode == HttpStatusCode.Gone) LastVersion = null; // NOTE: in practice this seems to be reported as a watch event...
+                            else await ReportErrorAndWait(response).ConfigureAwait(false);
+                        }
+                        else // if the response was successful, then the watch is open...
+                        {
+                            bool reset = LastVersion == null || list != null;
+                            if (reset && Reset != null) Reset(this); // if we may have lost events, let the client know
```
andykernahan, line 138: There is a race condition here between the null check, `remove_Reset`, and the event being invoked.

Consider invoking all events using the null-conditional operator to eliminate the race condition.

src/KubernetesClient/Watch.cs
admilazz: I'll improve it.

src/KubernetesClient/Watch.cs
admilazz: The CancellableStreamReader can cancel even normal blocking IO by closing the stream when the token is canceled, though I'm unsure of what exception will be thrown if any in that case.

src/KubernetesClient/Watch.cs
admilazz: I considered it but liked having the fast path not create a task. But your suggestion of only doing it if we would do a read anyway is better. I'll do that.

src/KubernetesClient/Watch.cs
admilazz: Well, I really don't think people should be adding and removing event handlers on a running watch, since that's bound to create race conditions in whether or not they receive the events. But sure, just as a best practice to guard against that pathological case, I'll make the change.

src/KubernetesClient/Watch.cs
admilazz: Changed.

src/KubernetesClient/Watch.cs
andykernahan: Cancellation is not my concern, it's that blocking IO will result in blocked thread pool threads. Admittedly, threads may only be blocked for short periods, but in a system with many watchers it may cause the number of threads to grow large.

Can it not be made async all the way?

admilazz: I think that is sufficient, and furthermore I don't see an async version of Deserialize or an equivalent, but let me know if I'm missing something.

admilazz: I can make the calls to Read async. I don't know about Deserialize.

admilazz:
@andykernahan I appreciate the review. Thanks.

src/KubernetesClient/Watch.cs
```diff
@@ -0,0 +1,491 @@
...
+                return totalRead;
+            }
+#else
+            async Task<int> ReadAsync(char[] buffer, int index, int length, bool block)
+            {
+                int start = index;
+                while(length != 0 && (charRead != charWrite || await EnsureDataAsync().ConfigureAwait(false)))
+                {
+                    int toRead = Math.Min(length, charWrite-charRead);
+                    Array.Copy(charBuffer, 0, buffer, index, toRead);
```
andykernahan, line 458: `charRead` must be given for `sourceIndex`:
```csharp
Array.Copy(charBuffer, charRead, buffer, index, toRead);
```

src/KubernetesClient/Watch.cs
```diff
@@ -0,0 +1,491 @@
...
+            }
+
+#if NETCOREAPP2_1
+            async ValueTask<int> ReadAsync(Memory<char> buffer, bool block)
+            {
+                int totalRead = 0;
+                while(buffer.Length != 0 && (charRead != charWrite || await EnsureDataAsync().ConfigureAwait(false)))
+                {
+                    int toRead = Math.Min(buffer.Length, charWrite-charRead);
+                    charBuffer.AsMemory(0, toRead).CopyTo(buffer);
```
andykernahan, line 443: I think `charRead` must be given for `start`?
```csharp
charBuffer.AsMemory(charRead, toRead).CopyTo(buffer);
```
Also, I believe `AsSpan` be used instead as the span would remain on the stack.

src/KubernetesClient/Watch.cs
admilazz: Shame. Of course you're right.

src/KubernetesClient/Watch.cs
admilazz: Fixed.

src/KubernetesClient/Watch.cs
andykernahan: Here's a potential implementation for `!NET452`!
```csharp
try
{
  await reader.ReadAsync(cancelToken).ConfigureAwait(false);
  // Load.
  var json = await JObject.LoadAsync(reader, cancelToken).ConfigureAwait(false);
  var type = (string)json.Property("type");
  var obj = json.Property("object")?.Value;
  if (type == null || obj == null) throw new JsonSerializationException("The stream does not appear to contain watch events.");
  // Deserialize.
  var e = new WatchEvent<T>();
  e.Type = (WatchEventType)Enum.Parse(typeof(WatchEventType), type, true);
  if (e.Type != WatchEventType.Error) e.Object = obj.ToObject<T>(serializer);
  else e.Error = obj.ToObject<V1Status>(serializer);
  return e;
}
catch (IOException ex) when (ex.InnerException is WebException wex && wex.Status == WebExceptionStatus.RequestCanceled)
{
    throw new OperationCanceledException("The watch read was canceled.", ex.InnerException, cancelToken);
}
```
**NB** it was hacking this that identified the `charRead` issues.

src/KubernetesClient/Watch.cs
```diff
@@ -0,0 +1,491 @@
...
+                                        guessedApiVersion = list.ApiVersion; // if it's missing, assume the API version is the same as the list's
+                                        // and for the kind, we'll use a heuristic that if the list kind is FooList then the item kind is Foo
+                                        guessedKind = list.Kind != null && list.Kind.Length > 4 && list.Kind.EndsWith("List") ?
+                                            list.Kind.Substring(0, list.Kind.Length-4) : null;
+                                    }
+                                    foreach (T item in list.Items)
+                                    {
+                                        if (item.ApiVersion == null) item.ApiVersion = guessedApiVersion;
+                                        if (item.Kind == null) item.Kind = guessedKind;
+                                        EventReceived(this, WatchEventType.Added, item);
```
andykernahan, line 161: Last one: `EventReceived?.Invoke(...)`

admilazz: Right-o.

src/KubernetesClient/Watch.cs
admilazz: Hmm. I'm glad to know it's possible. I'm not sure which is better, but I'm not fond of how it goes from JSON -> JObject -> T (allocating a bunch of memory for the object tree) rather than just going from JSON -> T, especially since in practice the server isn't going to send half an event and then stop (so it shouldn't be blocked on IO for any significant length of time, which was the same rationale I had for the synchronous calls to Read()).

src/KubernetesClient/Watch.cs
```diff
@@ -0,0 +1,491 @@
...
+                }
+            }
+            if (!gotObject) throw new JsonSerializationException("The stream does not appear to contain watch events.");
+            return e;
+        }
+
+        #region CancelableStreamReader
+        // HACK: annoyingly, the TextReader interface doesn't support cancellation of async reads outside .NET core. furthermore, Json.NET
+        // doesn't support cancellation even in several versions of .NET core. so we have this hacky TextReader that attempts cancellation
+        sealed class CancelableStreamReader : TextReader
```
tg123, line 341: what is new vs `PeekableStreamReader + CancelableStream` in WatcherDelegatingHandler

src/KubernetesClient/Watch.cs
tg123: it is really a headache if we cannot do all the way async
the event json was separate by line ending inside the stream
another way is to create a buf for current line
then fill the buf asyncly.

due to json cannot be stream-parsed, you have to buf entire line to read json

tg123:
my concern is if there is a way to merge high level `watch` and `watcher`
or create a new `IWatcher`

I think it would confuse first time users

tests/KubernetesClient.Tests/PipeStream.cs
```diff
@@ -0,0 +1,413 @@
+using System;
+using System.Buffers;
+using System.Collections.Generic;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace k8s.Tests
+{
+    #region AsyncAutoResetEvent
```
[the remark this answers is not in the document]
admilazz: The problem with it is that auto reset event is async only. I needed one that handled both synchronous and asynchronous use.

tests/KubernetesClient.Tests/PipeStream.cs
```diff
@@ -0,0 +1,413 @@
...
+        }
+        #endregion
+
+        readonly LinkedList<Chunk> dataList = new LinkedList<Chunk>();
+    }
+    #endregion
+
+    #region PipeStream
+    /// <summary>Represents a read-only stream backed by a <see cref="Pipe"/>.</summary>
+    sealed class PipeStream : Stream
```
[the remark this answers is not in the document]
admilazz: A memory stream wouldn't work the same way. Instead of blocking at the end of it, it will return EOF. And it's not really suitable for simultaneous reading and writing.

src/KubernetesClient/Watch.cs
admilazz: It's not enough to have a cancelable stream. In fact, our underlying stream is already cancelable. What's needed is a cancelable TextReader, since the TextReader interface does not take cancellation tokens.

admilazz: The PeekableStreamReader doesn't support anything but peeking and reading lines, and doesn't support cancellation.

tests/KubernetesClient.Tests/PipeStream.cs
```diff
@@ -0,0 +1,413 @@
...
+
+        readonly StreamBuffer buffer = new StreamBuffer();
+        readonly AsyncAutoResetEvent readEvent = new AsyncAutoResetEvent();
+        bool disposed, finished;
+    }
+    #endregion
+
+    #region StreamBuffer
+    /// <summary>Implements a buffer into which data can be written and from which data can be read.</summary>
+    sealed class StreamBuffer : IDisposable
```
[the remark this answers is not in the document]
admilazz: A stream buffer has a single position pointer shared between reading and writing. It's not suitable for a producer/consumer use pattern.

src/KubernetesClient/Watch.cs
admilazz: What is the headache, by the way? The most important place to avoid blocking is while waiting for an event, since that can take an arbitrary amount of time. Once the server decides to send an event, however, it sends it in its entirety. (The server doesn't send half an event and then stop.) Thus, while it's crucial to have the ability to cancel the wait for an event and async helps avoid blocking a thread there, it's not so important when parsing an event that is already being sent over the network.

admilazz:
> my concern is if there is a way to merge high level watch and watcher
> or create a new IWatcher
> I think it would confuse first time users

I agree. If it was all up to me I would deprecate the existing watch, but I didn't want to be so presumptuous.

brendandburns:
I will review once this is rebased.

admilazz:
I guess I'll put this in a separate project, then.

PR #410, opened 2020-04-12
Implement exec over the SPDY protocol

admilazz:
This change adds improved exec support. The existing exec support uses web sockets, which [has a flawed protocol](https://github.com/kubernetes/kubernetes/issues/89899) leading to some commands running forever. This change helps address issue #397 .

I tried to replace the implementation of the existing exec support to provide a seamless upgrade. Unfortunately, many internal implementation details of the web socket approach (e.g. WebSocketNamespacedPodExecAsync, MuxedStream, StreamDemuxer, etc), were publicly exposed, and it would be a breaking change to remove them. I did replace the implementation of NamespacedPodExecAsync, however, which was generic enough.

This PR makes the following public API changes:

* The fluent API gains new members:
  * ExecCommandAsync - Execs the current resource, which does not necessarily have to be a pod
  * OpenSPDYAsync - Makes the request and upgrades it to a SPDY connection
  * OpenWebSocketAsync - Makes the request and upgrades it to a web socket connection

**NOTE: This PR builds on and includes changes from two other PRs (https://github.com/kubernetes-client/csharp/pull/405 and https://github.com/kubernetes-client/csharp/pull/406).** I'm not sure how best to represent that in a PR, but the files actually added or changed in this specific PR are:

* ByteBuffer.cs (to edit a comment)
* Kubernetes.ConfigInit.cs (to remove the automatic retry support added by the Microsoft.Rest.ClientRuntime)
* Kubernetes.Exec.cs (to replace the NamespacedPodExecAsync implementation)
* KubernetesRequest.cs (to add the three methods named above)
* PipeStream.cs
* SPDYExec.cs

### DIFF
--- a/src/KubernetesClient/ByteBuffer.cs
+++ b/src/KubernetesClient/ByteBuffer.cs
@@ -5,9 +5,11 @@ using System.Threading;
 
 namespace k8s
 {
-    // There may be already an async implementation that we can use:
-    // https://github.com/StephenCleary/AsyncEx/wiki/AsyncProducerConsumerQueue
-    // However, they focus on individual objects and may not be a good choice for use with fixed-with byte buffers
+    // Pipe could be used instead for an async implementation. However, since this class is public there's no telling whether somebody
+    // has referenced it, so removing this class would be a breaking change. Neither could we easily replace the implementation of
+    // this class with Pipe's implementation, since this class exposes various public members (like WriteWaterMark and the OnResize event)
+    // that don't make sense in Pipe's context. (It seems they are used for unit testing, but unfortunately they were made public rather
+    // than internal, so removing them is a breaking change.)
 
     /// <summary>
     /// Represents a bounded buffer. A dedicated thread can send bytes to this buffer (the producer); while another thread can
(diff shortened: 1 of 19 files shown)


brendandburns:
I'll review this after the dependent PRs are merged.

andykernahan:
@admilazz Are you planning to add tests for the new types and functionality? Similarly, the new [`SPDY.net`](https://github.com/adammil/SPDY.net) dependency was only very recently created (by you under @AdamMil?) and is not tested.

admilazz:
> @admilazz Are you planning to add tests for the new types and functionality? Similarly, the new SPDY.net dependency was only very recently created (by you under @AdamMil?) and is not tested.

Yes, but as for the priority of it I will see what happens to the other PRs. This is way down the stack.

admilazz:
I guess I'll put this in a separate project.

admilazz:
> the new SPDY.net dependency was only very recently created (by you under @AdamMil?) and is not tested.

P.S. I wouldn't say it's untested, though. :-)  I did run 20,000 execs of various kinds in a row with zero failures... But it's true I haven't written unit tests yet.

Issue #409, opened 2020-04-10
KubernetesClientConfiguration.BuildDefaultConfig does not handle multiple files specified by KUBECONFIG env-var


rynowak:
From the k8s docs:

> The KUBECONFIG environment variable holds a list of kubeconfig files. For Linux and Mac, the list is colon-delimited. For Windows, the list is semicolon-delimited. The KUBECONFIG environment variable is not required. If the KUBECONFIG environment variable doesn’t exist, kubectl uses the default kubeconfig file, $HOME/.kube/config.

> If the KUBECONFIG environment variable does exist, kubectl uses an effective configuration that is the result of merging the files listed in the KUBECONFIG environment variable.

It looks like this has just yet to be implemented. 

[This code](https://github.com/kubernetes-client/csharp/blob/master/src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs#L41) assumes that the KUBECONFIG env-var only contains a single filename. 

The result is that if you've got multiple files (on Windows at least) you'll get an exception saying that the file could not be found. (After attempting to find a file matching the env-var verbatim).

Originally reported here: https://github.com/dotnet/tye/issues/351

brendandburns:
Yeah, the `KUBECONFIG` merging logic is super-complicated and not defined anywhere. It's basically only defined in the golang code that actually does the merge.

What specific merging are you looking for? It wouldn't be hard to add some simplistic merging (e.g. merge-if-not-present) but it won't exactly match what the Golang client does.

jkotalik:
Probably just a simple merging (merge-if-not-present) is fine for now. I'd be willing to contribute this 😄 .

I'm reading some of the docs for the rules for merging kubeconfig with an environment variable, and it sounds like the goal of it was [to be a simple merge-if-not-present]( https://kubernetes.io/docs/concepts/configuration/organize-cluster-access-kubeconfig/#merging-kubeconfig-files).

>Otherwise, if the KUBECONFIG environment variable is set, use it as a list of files that should be merged. Merge the files listed in the KUBECONFIG environment variable according to these rules:
>  - Ignore empty filenames.
>  - Produce errors for files with content that cannot be deserialized.
>  - The first file to set a particular value or map key wins.
>  - Never change the value or map key. Example: Preserve the context of the first file to set current-context. Example: If two files specify a red-user, use only values from the first file’s red-user. Even if the second file has non-conflicting entries under red-user, discard them.

Has the Golang client evolved past these rules at this point?

[...]

PR #411, opened 2020-04-14
Allow KUBECONFIG environment variable to point to multiple files

jkotalik:
Fixes https://github.com/kubernetes-client/csharp/issues/409.

I wanted to draft this PR because I hit some places where I'd like to confirm whether we should have behavior changes or have API changes. 

From what I can tell, KubernetesClientConfiguration.BuildDefaultConfig() is the only API which reads the environment variable "KUBECONFIG". This feels odd because apis like LoadKubeConfig(Async) don't respect the [environment variable](https://github.com/kubernetes-client/csharp/blob/master/src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs#L479), they only respect the optional path or default. Also, there is no other way to explicitly load from the environment variable.

I'm wondering if changing the default behavior for BuildConfigFromConfigFile and LoadKubeConfigAsync is the right move here. This would be a breaking change and could be decently disruptive.

We could also take the approach of adding separate APIs to load directly from the environment variable, which would be less breaking.


jkotalik:
reverifying cla

tests/KubernetesClient.Tests/KubernetesClientConfigurationTests.cs
```diff
@@ -407,6 +409,75 @@ public void LoadKubeConfigStream()
...
+        }
+
+        [Fact]
+        public void LoadK8ConfigFromEnvironmentVariable()
+        {
+            var txt = File.ReadAllText("assets/kubeconfig.yml");
+            var expectedCfg = Yaml.LoadFromString<K8SConfiguration>(txt);
+
+            var filePath = Path.GetFullPath("assets/kubeconfig.yml");
+            Environment.SetEnvironmentVariable("KUBECONFIG", filePath);
```
davidfowl, line 432: This seems bad for parallel execution. These tests need to be part of a collection that don't run in parallel.

jkotalik: Yeah, I have two ways I could fix it, but both are based on what public API we want.

If we decided to keep the API where we we can pass in an environment variable name, I can make it so each test uses a unique variable name. 

Otherwise, I'll probably do some InternalsVisibleTo stuff to change the return value for the environment variable.

src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
```diff
@@ -560,6 +566,49 @@ public static K8SConfiguration LoadKubeConfig(Stream kubeconfigStream)
             return LoadKubeConfigAsync(kubeconfigStream).GetAwaiter().GetResult();
         }
 
+        /// <summary>
+        ///    Loads Kube Config from KUBECONFIG environment variable
+        /// </summary>
+        /// <param name="useRelativePaths">When <see langword="true"/>, the paths in the kubeconfig file will be considered to be relative to the directory in which the kubeconfig
+        /// <returns>Instance of the <see cref="K8SConfiguration"/> class</returns>
+        public static K8SConfiguration LoadKubeConfigFromEnvironmentVariable(string environmentVariable = "KUBECONFIG", bool useRelativePaths = true)
```
brendandburns, line 574: Does this really need to be a separate method?

src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
```diff
@@ -560,6 +566,49 @@ public static K8SConfiguration LoadKubeConfig(Stream kubeconfigStream)
...
+        /// </summary>
+        /// <param name="useRelativePaths">When <see langword="true"/>, the paths in the kubeconfig file will be considered to be relative to the directory in which the kubeconfig
+        /// <returns>Instance of the <see cref="K8SConfiguration"/> class</returns>
+        public static async Task<K8SConfiguration> LoadKubeConfigFromEnvironmentVariableAsync(string environmentVariable = "KUBECONFIG", bool useRelativePaths = true)
+        {
+            var kubeconfig = Environment.GetEnvironmentVariable(environmentVariable);
+
+            if (kubeconfig == null)
+            {
+                throw new KubeConfigException($"Environment variable KUBECONFIG is empty.");
```
brendandburns, line 590: Since the environment variable name is configurable, you should probably use `environmentVariable`'s value in the error message.

src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
```diff
@@ -560,6 +566,49 @@ public static K8SConfiguration LoadKubeConfig(Stream kubeconfigStream)
...
+            }
+
+            var configList = kubeconfig.Split(RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? ';' : ':');
+            if (configList.Length > 1)
```
brendandburns, line 594: This if is redundant with the `for` loop below, delete the if and the else and just leave the body of the if case.

src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
```diff
@@ -589,5 +638,95 @@ private static string GetFullPath(K8SConfiguration configuration, string path)
                 return Path.Combine(Path.GetDirectoryName(configuration.FileName), path);
             }
         }
+
+        private static void MergeKubeConfig(K8SConfiguration basek8SConfig, K8SConfiguration mergek8SConfig)
+        {
+            // For scalar values, prefer local values 
+            basek8SConfig.ApiVersion = basek8SConfig.ApiVersion ?? mergek8SConfig.ApiVersion;
+            basek8SConfig.CurrentContext = basek8SConfig.CurrentContext ?? mergek8SConfig.CurrentContext;
```
brendandburns, line 646: This is tricky (and an example of why merge is hard) I'm not sure first or last config wins is what people expect.

src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
```diff
@@ -589,5 +638,95 @@ private static string GetFullPath(K8SConfiguration configuration, string path)
                 return Path.Combine(Path.GetDirectoryName(configuration.FileName), path);
             }
         }
+
+        private static void MergeKubeConfig(K8SConfiguration basek8SConfig, K8SConfiguration mergek8SConfig)
+        {
+            // For scalar values, prefer local values 
+            basek8SConfig.ApiVersion = basek8SConfig.ApiVersion ?? mergek8SConfig.ApiVersion;
```
brendandburns, line 645: You can remove this, this field is deprecated anyway.

src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
```diff
@@ -589,5 +638,95 @@ private static string GetFullPath(K8SConfiguration configuration, string path)
...
+            basek8SConfig.CurrentContext = basek8SConfig.CurrentContext ?? mergek8SConfig.CurrentContext;
+            basek8SConfig.FileName = basek8SConfig.FileName ?? mergek8SConfig.FileName;
+            basek8SConfig.Kind = basek8SConfig.Kind ?? mergek8SConfig.Kind;
```
brendandburns, line 648: Should probably throw if these are different.

src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
```diff
@@ -589,5 +638,95 @@ private static string GetFullPath(K8SConfiguration configuration, string path)
...
+                }
+            }
+
+            // For IEnumerable, use name as unique identifier for whether we should insert or not.
+            if (mergek8SConfig.Contexts != null && mergek8SConfig.Contexts.Count() > 0)
+            {
+                var contexts = new List<Context>();
+                if (basek8SConfig.Contexts != null)
+                {
+                    contexts.AddRange(basek8SConfig.Contexts);
```
brendandburns, line 679: This is definitely going to run into problems (weird problems in kubectl b/c golang randomizes iteration) if contexts are duplicated. You can't just blindly add things into the lists, you need to make sure there are no duplicates.

brendandburns: oh, oops, nevermind, I see you are checking below.

src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
```diff
@@ -589,5 +638,95 @@ private static string GetFullPath(K8SConfiguration configuration, string path)
...
+                }
+
+                foreach (var context in mergek8SConfig.Contexts)
+                {
+                    if (basek8SConfig.Contexts?.FirstOrDefault(o => o.Name == context.Name) == null)
+                    {
+                        contexts.Add(context);
+                    }
```
brendandburns, line 687: Probably want `else { throw ... }` so that people aren't confused if things are silently dropped.

src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
```diff
@@ -589,5 +638,95 @@ private static string GetFullPath(K8SConfiguration configuration, string path)
...
+                    if (basek8SConfig.Clusters?.FirstOrDefault(o => o.Name == cluster.Name) == null)
+                    {
+                        clusters.Add(cluster);
+                    }
+                }
+
+                basek8SConfig.Clusters = clusters;
+            }
+
+            if (mergek8SConfig.Users != null && mergek8SConfig.Users.Count() > 0)
```
brendandburns, line 712: this logic feels redundant and it could be made generic somehow.

brendandburns:
Thanks for the PR, added some comments.

src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
davidfowl: Isn't last wins most intuitive? (Like the PATH)

jkotalik: According to https://kubernetes.io/docs/concepts/configuration/organize-cluster-access-kubeconfig/#merging-kubeconfig-files:

> The first file to set a particular value or map key wins.

I think preferring first wins should be implied here. However, for specific values like CurrentContext, this may vary?

jkotalik: Actually, I'm getting more confused now, because from what I can tell, [the go client is merging with overwrite for each kubeconfig](https://github.com/kubernetes/client-go/blob/master/tools/clientcmd/loader.go#L229).

src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
jkotalik: Hm, I'm curious about this. So if two kubeconfig.yml files define the same contexts (or Clusters, Users, Extensions or Preferences) or have them with the same name, it would be better to throw instead of ignore?

From what I can tell, that doesn't match what the go client does, but I'm not certain.

src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
davidfowl: Wait, first wins in the PATH 😄 ....

brendandburns: Yeah, this is my broader point about fancy merges being mostly only documented in how it is implemented in kubectl.

We can emulate that or we can just throw when we hit something we don't understand/don't spec.

src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
brendandburns: I guess my point is that if people have multiple different contexts, I feel like they're going to be pretty confused no matter what. My recollection is that if there are two identical things the go client picks randomly because map iteration is random in Go, but I'm not positive.

brendandburns:
@jkotalik let me know when this is ready for another review.

src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
```diff
@@ -589,5 +631,74 @@ private static string GetFullPath(K8SConfiguration configuration, string path)
...
+                foreach (var extension in mergek8SConfig.Extensions)
+                {
+                    if (basek8SConfig.Extensions?.ContainsKey(extension.Key) == false)
+                    {
+                        basek8SConfig.Extensions[extension.Key] = extension.Value;
+                    }
+                }
+            }
+
+            if (mergek8SConfig.Clusters != null && mergek8SConfig.Clusters.Count() > 0)
```
rynowak, line 665: worth noting (maybe a comment) that clusters are *map-like* in config despite being represented as a list here. https://github.com/kubernetes/client-go/blob/ede92e0fe62deed512d9ceb8bf4186db9f3776ff/tools/clientcmd/api/types.go#L238

src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
```diff
@@ -589,5 +631,74 @@ private static string GetFullPath(K8SConfiguration configuration, string path)
...
+            {
+                var clusters = new List<Cluster>();
```
rynowak, line 667: If you're going to make a temp data structure, it might be simpler to use a dictionary where the key is the cluster name.

src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
```diff
@@ -589,5 +631,74 @@ private static string GetFullPath(K8SConfiguration configuration, string path)
...
+                if (basek8SConfig.Clusters != null)
+                {
+                    clusters.AddRange(basek8SConfig.Clusters);
+                }
+
+                foreach (var cluster in mergek8SConfig.Clusters)
+                {
+                    if (basek8SConfig.Clusters?.FirstOrDefault(o => o.Name == cluster.Name) == null)
```
rynowak, line 675: can also use `.Any` here

src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
```diff
@@ -589,5 +631,74 @@ private static string GetFullPath(K8SConfiguration configuration, string path)
...
+
+        private static void MergeKubeConfig(K8SConfiguration basek8SConfig, K8SConfiguration mergek8SConfig)
+        {
+            // For scalar values, prefer local values 
+            basek8SConfig.CurrentContext = basek8SConfig.CurrentContext ?? mergek8SConfig.CurrentContext;
+            basek8SConfig.FileName = basek8SConfig.FileName ?? mergek8SConfig.FileName;
+            basek8SConfig.Kind = basek8SConfig.Kind ?? mergek8SConfig.Kind;
+
+            // For Dictionaries, prefer first if not present in second.
+            if (mergek8SConfig.Preferences != null && mergek8SConfig.Preferences.Count > 0)
```
rynowak, line 643: having a check on count for `Preferences` and `Extensions` doesn't seem right since we're already doing uniqueness based on key. They are *map-like* in the go code.

src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
jkotalik: Looking back, this is actually doing two passes with the merge, one forward and one backward. It still looks like first wins 😄

src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
jkotalik: Based on that, I think picking the first one is fine then if the Go client picks randomly rather than throwing.

src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
jkotalik: I'll make this internal.

jkotalik:
I'd like to write a few more tests to check merging of extensions and preferences, however I don't know/see any examples for how they are represented in config. Could someone point me to where that happens.

This is ready for re-review @brendanburns

brendandburns:
@jkotalik 

There are two options, this tip here:
https://github.com/kubernetes/kubernetes/issues/46381#issuecomment-303926031

Gives you a way to test/preview what `kubectl` does.

The go code for it is here:
https://github.com/kubernetes/client-go/blob/master/tools/clientcmd/loader.go#L199

Which ultimately calls the `mergo`:
https://github.com/kubernetes/client-go/blob/master/tools/clientcmd/loader.go#L230

Hope that helps. As I mentioned this is a twisty/poorly documented area of the project.

brendandburns:
Code looks good to me whenever you're happy with it.

jkotalik:
I verified my output for tests with the result of kubectl config view and it looked like it matched (minus ordering).

brendandburns:
@jkotalik are you ready to merge? Or did you want to add more validation?

jkotalik:
Yeah I feel good now, added a couple more tests for untested parts.

PR #394, opened 2020-04-01
Add support for informers

macsux:
This PR adds support for informers that form building block for creating controllers. It follows the spirit of original informers, but differs in implementation found in Go and Java in a few significant ways.

- The entire library is built on reactive paradigm using Reactive Extensions. RX is a natural fit as we're dealing with streams of events, allowing leverage powerful built in functional folds over live streams of object change data. This offers advantages in:

  - intuitive LINQ like operators to filter / transform and collect live information being streamed when attached to k8s
  - reuse of single connection for internal multicasting
  - scheduler abstractions to provide multithreaded synchronization and concurrency
  - virtual time for unit testing

- Flexible abstraction model based on IInformer interface that allows modeling observable resources over resources other then K8S. This allows observing a SET of resources of similar type while conforming to same logical interface, easily allowing connecting one resource type to another. IInformer abstracts out a List/Watch operations into a single observable stream of events. Each event in the stream includes a bitvise enum mask that describes event purpose and how receivers of event should treat it, as well the current state of the object. 

  - List operations are marked by event type "Reset", with first and last elements having ResetStart/ResetEnd flags. Reset also implies that any previously tracked cache state should be discard and new one used.
  - If list operation is empty and is followed by watch, ResetEmpty flag is used to denote transition point
  - Add/Modify/Delete flags denote changes to existing state
  - (Re)Sync events to denote periodic refreshes of same data already sent

  The intended vision is to allow reuseable set of informers for other APIs to be created and used as building blocks for creating controllers. For example a reusable informer can be written for twitter feed, leveraging push based websocket features. Legacy APIs without efficient change detection such an equivalent of Watch can be abstracted to provide push based events via polling. 

- Concrete implementations of IInformer follow individual API rules, as well as provide reasonable failure recovery models.

- Generic List operation for `Kubernates` object that can work with any existing object or CRD.

  - Future enhancement will add similar features for modifying the state

- Included K8S informer provides supports

  - List followed by Watch
  - Fault tolerance policy with default backoff retry
  - Seamless connection recovery over transient error codes, including support for bookmarks
  - Server side support for namespace filtering
  - Future todo: server side label filtering

[...]

PR #412, opened 2020-04-14
Generic methods

macsux:
- Generic methods for Kubernetes
- Friend assembly tags
- Timeout handler for when using external HttpClient with Autorest
- Support for bookmarks in the watcher

This PR is pre-req for informers PR https://github.com/kubernetes-client/csharp/pull/394


brendandburns:
Please resolve conflicts then I will review.

macsux:
@brendandburns Resolved

src/KubernetesClient/Kubernetes.Generic.cs
```diff
@@ -0,0 +1,1131 @@
...
+            cancellationToken.ThrowIfCancellationRequested();
+            httpResponse = await HttpClient.SendAsync(httpRequest, cancellationToken).ConfigureAwait(false);
+            if (shouldTrace)
+            {
+                ServiceClientTracing.ReceiveResponse(invocationId, httpResponse);
+            }
+            var statusCode = httpResponse.StatusCode;
+            cancellationToken.ThrowIfCancellationRequested();
+            string responseContent = null;
+            if ((int)statusCode != 200 && (int)statusCode != 401)
```
brendandburns, line 161: This is inconsistent with the `OK` below, please use constants everywhere for consistency.

src/KubernetesClient/Kubernetes.Generic.cs
```diff
@@ -0,0 +1,1131 @@
...
+            }
+            // Create HTTP transport objects
+            var httpRequest = new HttpRequestMessage();
+            HttpResponseMessage httpResponse = null;
+            httpRequest.Method = new HttpMethod("GET");
+            httpRequest.RequestUri = new System.Uri(url);
+            // Set Headers
+
+
+            if (customHeaders != null)
```
brendandburns, line 280: A bunch of this logic is repeated in multiple places, please refactor into common methods.

src/KubernetesClient/Kubernetes.Generic.cs
```diff
@@ -0,0 +1,1131 @@
...
+            bool? export = default,
+            string pretty = default,
+            Dictionary<string, List<string>> customHeaders = null,
+            CancellationToken cancellationToken = default)
+        {
+
+            // Tracing
+            var shouldTrace = ServiceClientTracing.IsEnabled;
+            string invocationId = null;
+            if (shouldTrace)
```
brendandburns, line 241: Though this logic is a little different than previous, I think you could pretty easily refactor it into a helper method that was something like:

```
if (shouldTrace)
{
    addTraces(["exact", exact], ["export", export], ....);
}

File: src/KubernetesClient/Kubernetes.Generic.cs [hunk 4]
```diff
@@ -0,0 +1,1131 @@
...
+                tracingParameters.Add("timeoutSeconds", timeoutSeconds);
+                tracingParameters.Add("watch", watch);
+                tracingParameters.Add("namespaceParameter", namespaceParameter);
+                tracingParameters.Add("pretty", pretty);
+                tracingParameters.Add("cancellationToken", cancellationToken);
+                ServiceClientTracing.Enter(invocationId, this, "ListNamespacedPod", tracingParameters);
+            }
+            var url = GetUrlPathForResource(type, namespaceParameter);
+            var queryParameters = new List<string>();
+            if (allowWatchBookmarks != null)
```
brendandburns, line 79: There is logic for building query strings here:

https://github.com/kubernetes-client/csharp/blob/master/src/KubernetesClient/Utilities.cs#L9

That will clean this up a lot.

src/KubernetesClient/Kubernetes.Generic.cs
```diff
@@ -0,0 +1,1131 @@
...
+            }
+            // Create Result
+            var result = (HttpOperationResponse)Activator.CreateInstance(typeof(HttpOperationResponse<>).MakeGenericType(type));
+            result.Request = httpRequest;
+            result.Response = httpResponse;
+            // Deserialize Response
+
+            if (statusCode == OK)
+            {
+                responseContent = await httpResponse.Content.ReadAsStringAsync().ConfigureAwait(false);
```
brendandburns, line 344: I think this logic can be extracted too.

src/KubernetesClient/Kubernetes.Generic.cs
```diff
@@ -0,0 +1,1131 @@
...
+            cancellationToken.ThrowIfCancellationRequested();
+            httpResponse = await HttpClient.SendAsync(httpRequest, cancellationToken).ConfigureAwait(false);
+            if (shouldTrace)
+            {
+                ServiceClientTracing.ReceiveResponse(invocationId, httpResponse);
+            }
+            var statusCode = httpResponse.StatusCode;
+            cancellationToken.ThrowIfCancellationRequested();
+            string responseContent = null;
+            if (statusCode != OK && statusCode != Accepted && statusCode != Unauthorized)
```
brendandburns, line 686: Delete can actually return Accepted and a `Status` object instead of the deleted object. We should probably handle that case here too.

src/KubernetesClient/Extensions.cs
```diff
@@ -0,0 +1,36 @@
...
+using System.Reflection;
+using System.Text.RegularExpressions;
+using k8s.Models;
+
+namespace k8s
+{
+    public static class Extensions
+    {
+
+        public static KubernetesEntityAttribute GetKubernetesTypeMetadata<T>(this T obj) where T : IKubernetesObject =>
```
brendandburns, line 11: We now have an `IMetadata` interface:

https://github.com/kubernetes-client/csharp/blob/master/src/KubernetesClient/IMetadata.cs

I think it makes sense to use that here.

brendandburns:
This looks reasonable.

There's a bunch of repeated code that I think could be refactored into shared methods, and some comments about using the shared `IMetadata` interface.

Thanks!

src/KubernetesClient/Kubernetes.Generic.cs
macsux: This code was taken directly from generated methods of Kubernetes object (hence the duplicate code - I wasn't trying to pretty it up beyond making it generic). 

Personally I'm of the opinion that once these methods are standardized, 90% of Kubernetes class autogenerated methods can be made to just call these generic methods.

src/KubernetesClient/Extensions.cs
macsux: This is different metadata as in this context I mean grab the type's `KubernetesEntityAttribute` instead of Metadata field. I guess I can rename it to something else - I can see the confusion

macsux:
This should be much cleaner now, I've consolidated pretty much all requests into a single method, and added integration tests (currently disabled). 

@brendandburns I've added a test fixture to spin up Kind cluster to do integration testing. Did we wanna add this as part of pipeline? I'm sure we have other tests that can benefit from this.

src/KubernetesClient/KubernetesEntityAttribute.cs.orig
```diff
@@ -0,0 +1,78 @@
+using System;
```
brendandburns, line 1: I think this file should be deleted?

brendandburns:
This generally looks good to me. It needs a rebase and I think there is an extraneous file that was left over from development.

brendandburns:
Oh, and I think the Kind testing is awesome!

macsux:
Please hold off on merging this just yet, as if there's a green light to do what is propsed in https://github.com/kubernetes-client/csharp/issues/449 I would like to refactor this as it's own class with extension methods for KubernetesClient. The target implementation would look like this (and removed from Steeltoe if accepted as part of KubernetesClient) if the above is approved:

https://github.com/steeltoeoss-incubator/kubernetes/blob/informers/src/KubernetesBase/Kubernetes.Generic.cs

PR #414, opened 2020-04-15
Add WatchEventType.Bookmark

admilazz:
This change adds a Bookmark value to the WatchEventType enum, to support watch bookmarks.


tg123:
/LGTM

thanks for smaller changes,
my brain's memory is too ..... (mem overflow)

Issue #417, opened 2020-04-16
Autogenerated OpenAPI schema classes don't work as expected


macsux:
CustomResourceDefinition includes support for specifying OpenAPI schema. The generated classes that define the model for OpenAPI seem to be having issues with some types of common schema semantics. Here's an OpenAPI schema as generated by NJsonSchema

```
{
  "$schema": "http://json-schema.org/draft-04/schema#",
  "title": "MyCustomResource",
  "type": "object",
  "additionalProperties": false,
  "properties": {
    "spec": {
      "x-nullable": true,
      "oneOf": [
        {
          "$ref": "#/definitions/MyCustomResourceSpec"
        }
      ]
    },
    "status": {
      "x-nullable": true,
      "oneOf": [
        {
          "$ref": "#/definitions/MyCustomResourceStatus"
        }
      ]
    }
  },
  "definitions": {
    "MyCustomResourceSpec": {
      "type": "object",
      "additionalProperties": false,
      "properties": {
        "value": {
          "type": "string",
          "x-nullable": true
        },
        "item": {
          "type": "string",
          "x-nullable": true
        },
        "Replicas": {
          "type": "integer",
          "format": "int32"
        }
      }
    },
    "MyCustomResourceStatus": {
      "type": "object",
      "additionalProperties": false,
      "properties": {
        "Replicas": {
          "type": "integer",
          "format": "int32"
        }
      }
    }
  }
}
```
When we try to deserialize this into V1beta1JSONSchemaProps ex.`JsonConvert.DeserializeObject<V1beta1JSONSchemaProps>(schema.ToJson())`, it generates a single item in `Properties["spec"].OneOf` with null value. This is incorrect and should have resulted in value being another `V1beta1JSONSchemaProps` with `RefProperty` set to what we see as part of the `$ref` in the schema.

This is currently acting as a bottleneck for a PR I was getting ready to submit that allows generating and installing CRDs, but I suspect will cause the same issue when trying to read `CustomResourceDefinition` from API server as it will fail correct deserialization (and can potentially corrupt the CRD if used to modify its structure).

brendandburns:
Take a look at the spec here:
https://raw.githubusercontent.com/kubernetes/kubernetes/master/api/openapi-spec/swagger.json

I think the issue is that v1beta1.JSONSchemaProps is JSONSchema, not OpenAPI.

v1.JSONSchemaProps is OpenAPI v3

There's a big discussion here:
[...]

PR #418, opened 2020-04-16
Support for defining and installing CRDs

macsux:
This adds support for generating CRD out of CLR types. It includes a builder class that allows combining multiple versions of the same CR into a single logical CRD. This is currently not working 100% as expected, specifically, the schema doesn't get installed into API server due to a blocking bug as per https://github.com/kubernetes-client/csharp/issues/417

I would like to get feedback on this before I cover it with unit tests. Few questions for core committers:
1. Should this be isolated as own package (due to new dependency on NJsonSchema)
2. I'm leaning towards introducing mapper functionality to avoid transitioning between versions of same object (v1 <--> v1beta1). This will avoid explicit methods that call for specific version of object to be created as currently seen in the PR. The plan is to use AutoMapper to map transitions between versions, and include maps for core types as well. This aligns well with converter functionality already offered by K8S via webhook endpoint. 

Given the two items above I propose introducing `KubernetesClient.CustomResources` and `KubernetesClient.Versioning` packages.


brendandburns:
Once we get the mapper PR merged, I will review this PR. Apologies for the delay.

brendandburns:
Ok, the mapping PR is merged. Once this is rebased, I will review.

PR #419, opened 2020-04-16
Remove Microsoft.AspNetCore.WebUtilities dependency

admilazz:
This change removes the Microsoft.AspNetCore.WebUtilities dependency (which pulls in several other unnecessary dependencies as well). The library was only used to manipulate query strings, but that can be done simply without it. Some minor bugs found in query-string handling were also fixed.


src/KubernetesClient/KubernetesClient.csproj
```diff
@@ -32,12 +32,11 @@
...
     <PackageReference Include="Nerdbank.GitVersioning" Version="3.0.50" PrivateAssets="all" />
     <PackageReference Include="Portable.BouncyCastle" Version="1.8.1.3" />
-    <PackageReference Include="Microsoft.AspNetCore.WebUtilities" Version="1.1.2" Condition="'$(TargetFramework)' != 'netstandard2.0'" />
-    <PackageReference Include="Microsoft.AspNetCore.WebUtilities" Version="2.2.0" Condition="'$(TargetFramework)' == 'netstandard2.0'" />
     <PackageReference Include="Microsoft.Rest.ClientRuntime" Version="2.3.10" />
     <PackageReference Include="Newtonsoft.Json" Version="9.0.1" Condition="'$(TargetFramework)' != 'netstandard2.0' and '$(TargetFramework)' != 'netcoreapp2.1'" />
     <PackageReference Include="Newtonsoft.Json" Version="12.0.2" Condition="'$(TargetFramework)' == 'netstandard2.0' or '$(TargetFramework)' == 'netcoreapp2.1'" />
     <PackageReference Include="YamlDotNet" Version="6.0.0" />
+    <PackageReference Include="System.Buffers" Version="4.5.1" Condition="'$(TargetFramework)' != 'netcoreapp2.1'" />
```
admilazz, line 39: This is one of the indirect dependencies pulled in by Microsoft.AspNetCore.WebUtilities, and it is being used.

brendandburns:
Please add a unit test for the query string generator then this is ready to merge.

src/KubernetesClient/Kubernetes.WebSocket.cs
```diff
@@ -226,14 +222,13 @@ public Task<WebSocket> WebSocketNamespacedPodAttachAsync(string name, string @na
...
 
-            uriBuilder.Query = QueryHelpers.AddQueryString(string.Empty, new Dictionary<string, string>
-            {
-                { "container", container},
-                { "stderr", stderr ? "1": "0"},
-                { "stdin", stdin ? "1": "0"},
-                { "stdout", stdout ? "1": "0"},
-                { "tty", tty ? "1": "0"}
-            }).TrimStart('?');
+            var query = new StringBuilder();
```
tg123, line 225: I would prefer not to manipulate strings
Maybe copy the code from WebUtilities?

QueryHelpers is more readable

admilazz:
> Please add a unit test for the query string generator then this is ready to merge.

I tried to but I want to stop - or at least not contribute to - the exposure of implementation details publicly which I think is hampering changes to the client by making changes breaking when they really shouldn't be.

Is there some way to share internal types with the tests? Normally you use the [InternalsVisibleTo] attribute, but that requires a signed assembly. I remember seeing some comments about this in the past but I don't know where that went...

src/KubernetesClient/Kubernetes.WebSocket.cs
admilazz: How about if I:
1. Call Utilities.AddQueryParameter for each line rather than using Append, or
2. Add Utilities.AddQueryParameters which can taken multiple parameters, and use that?

admilazz:
@brendandburns Okay, I added a test, which involved signing the test assembly so it can access the internals of the main client assembly.

Hopefully future implementation details can remain "internal" so they can change without needing a breaking change.

tg123:
link #421
personally, I would prefer existing wheel instead of creating a new one.

admilazz:
> link #421
> personally, I would prefer existing wheel instead of creating a new one.

Seems we made the same change at the same time. I merged in yours.

PR #420, opened 2020-04-16
Support for conversions between models of different versions

macsux:
- Allows conversions between models of different versions (v1 <--> v1beta1)
- Explicit type conversion via type operators

The use of automapper allows ensuring that all field mappings are accounted for. This work is a precursor to introducing API methods that automatically select the endpoint and model based on the server k8s version.


src/KubernetesClient/Versioning/KubernetesVersionComparitor.cs
```diff
@@ -0,0 +1,58 @@
+using System;
+using System.Collections.Generic;
+using System.Text.RegularExpressions;
+
+namespace k8s.Versioning
+{
+    public class KubernetesVersionComparer : IComparer<string>
```
tg123, line 7: `System.Version`?
you may want to reuse it

src/KubernetesClient/Versioning/VersionConverter.cs
```diff
@@ -0,0 +1,179 @@
...
+        {
+            var attr = type.GetCustomAttribute<KubernetesEntityAttribute>();
+            if(attr == null)
+                throw new InvalidOperationException($"Type {type} does not have {nameof(KubernetesEntityAttribute)}");
+            return attr;
+        }
+
+        internal static void GetKubernetesMaps(IMapperConfigurationExpression cfg)
+        {
+            cfg.DisableConstructorMapping();
```
tg123, line 69: generated?

macsux: I generated it in LinqPad, but few of these maps had to be manually realigned. I guess I can put the autogeneration logic where there are no conflicts into the gen project and augment it with manual maps for where there are conflicts.

src/KubernetesClient/Versioning/KubernetesVersionComparitor.cs
macsux: > is this to support api call to lower version k8s server?

Yes, there's an open PR for creating and installed CRDs from me. I've ran into this problem when installing into my test cluster which is actually using an older version of CRDs. I'm refactoring that code to be based on this one, and automatically do conversion based on server version. I'm partly blocked on that one until we figure out what to do with https://github.com/kubernetes-client/csharp/issues/417

brendandburns:
@macsux this seems like a reasonable start. Is this in a state where you'd like to merge it?

@tg123 any further comments on this PR?

macsux:
Yes, it's good to go as far as I'm concerned

PR #422, opened 2020-04-16
Timeout handler. 

macsux:
This is needed to allow KubernetesClient not timeout during watch requests when initialized with externally managed HttpClient

### DIFF
--- a/src/KubernetesClient/WatcherDelegatingHandler.cs
+++ b/src/KubernetesClient/WatcherDelegatingHandler.cs
@@ -12,12 +12,49 @@ namespace k8s
     /// <summary>
     /// This HttpDelegatingHandler is to rewrite the response and return first line to autorest client
     /// then use WatchExt to create a watch object which interact with the replaced http response to get watch works.
+    ///
+    /// This handler should be added when using externally managed <see cref="HttpClient"/> with <see cref="Kubernetes"/>.
+    /// It replaces the default Timeout behavior by ensuring that any request that is a watch does not timeout.
+    /// When using this handler, the <see cref="HttpClient.Timeout"/> should be set <see cref="Timeout.InfiniteTimeSpan"/>
+    /// and the desired actual timeout for non-watch requests should be passed as part of the <see cref="TimeoutHandler"/> constructor.
     /// </summary>
     internal class WatcherDelegatingHandler : DelegatingHandler
     {
+        private readonly TimeSpan _timeout;
+        private static TimeSpan DefaultTimeout = TimeSpan.FromSeconds(100); // 100 is the default for HttpClient
+
+        public WatcherDelegatingHandler() : this(DefaultTimeout)
+        {
+        }
+
+        public WatcherDelegatingHandler(TimeSpan timeout)
+        {
+            _timeout = timeout;
+        }
+
+        public WatcherDelegatingHandler(HttpMessageHandler innerHandler, TimeSpan timeout = default) : base(innerHandler)
+        {
+            _timeout = timeout != default ? timeout : DefaultTimeout;
+        }
+
         protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage request,
             CancellationToken cancellationToken)
         {
+            var isWatch = false;
+            if (request.Method == HttpMethod.Get)
+            {
+                string query = request.RequestUri.Query;
+                int index = query.IndexOf("watch=true");
+                if (index > 0 && (query[index - 1] == '&' || query[index - 1] == '?'))
+                    isWatch = true;
+            }
+
+            if (!isWatch)
+            {
+                var cts = new CancellationTokenSource(_timeout);
+                cancellationToken = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, cts.Token).Token;
+            }
+
             var originResponse = await base.SendAsync(request, cancellationToken).ConfigureAwait(false);
 
             if (originResponse.IsSuccessStatusCode && request.Method == HttpMethod.Get) // all watches are GETs, so we can ignore others
--- a/src/KubernetesClient/WatcherDelegatingHandler.cs
+++ b/src/KubernetesClient/WatcherDelegatingHandler.cs
@@ -46,7 +46,9 @@ namespace k8s
                 string query = request.RequestUri.Query;
                 int index = query.IndexOf("watch=true");
                 if (index > 0 && (query[index - 1] == '&' || query[index - 1] == '?'))
+                {
                     isWatch = true;
+                }
             }
 
             if (!isWatch)
--- a/src/KubernetesClient/Kubernetes.ConfigInit.cs
+++ b/src/KubernetesClient/Kubernetes.ConfigInit.cs
@@ -5,6 +5,7 @@ using System.Net;
 using System.Net.Http;
 using System.Net.Security;
 using System.Security.Cryptography.X509Certificates;
+using System.Threading;
 using k8s.Exceptions;
 using k8s.Models;
 using Microsoft.Rest;
@@ -233,6 +234,7 @@ namespace k8s
 
             AppendDelegatingHandler<WatcherDelegatingHandler>();
             HttpClient = new HttpClient(FirstMessageHandler, false);
+            HttpClient.Timeout = Timeout.InfiniteTimeSpan; // timeout logic will be handled by WatcherDelegatingHandler
         }
 
         /// <summary>
--- a/src/KubernetesClient/KubernetesClientConfiguration.HttpClientHandler.cs
+++ b/src/KubernetesClient/KubernetesClientConfiguration.HttpClientHandler.cs
@@ -54,5 +54,6 @@ namespace k8s
         }
 
         public static DelegatingHandler CreateWatchHandler() => new WatcherDelegatingHandler();
+        public static DelegatingHandler CreateWatchHandler(TimeSpan timeout) => new WatcherDelegatingHandler(timeout);
     }
 }
--- a/src/KubernetesClient/ModelExtensions.cs
+++ b/src/KubernetesClient/ModelExtensions.cs
@@ -80,8 +80,14 @@ namespace k8s.Models
             else
             {
                 int slash = obj.ApiVersion.IndexOf('/');
-                if (slash < 0) (group, version) = (string.Empty, obj.ApiVersion);
-                else (group, version) = (obj.ApiVersion.Substring(0, slash), obj.ApiVersion.Substring(slash + 1));
+                if (slash < 0)
+                {
+                    (group, version) = (string.Empty, obj.ApiVersion);
+                }
+                else
+                {
+                    (group, version) = (obj.ApiVersion.Substring(0, slash), obj.ApiVersion.Substring(slash + 1));
+                }
             }
         }
 
@@ -162,6 +168,7 @@ namespace k8s.Models
             {
                 throw new ArgumentNullException(nameof(obj));
             }
+
             if (predicate == null)
             {
                 throw new ArgumentNullException(nameof(predicate));
@@ -172,7 +179,10 @@ namespace k8s.Models
             {
                 for (int i = 0; i < ownerRefs.Count; i++)
                 {
-                    if (predicate(ownerRefs[i])) return i;
+                    if (predicate(ownerRefs[i]))
+                    {
+                        return i;
+                    }
                 }
             }
             return -1;
--- a/src/KubernetesClient/ModelExtensions.cs
+++ b/src/KubernetesClient/ModelExtensions.cs
@@ -53,7 +53,7 @@ namespace k8s.Models
             if (obj.ApiVersion != null)
             {
                 int slash = obj.ApiVersion.IndexOf('/');
-                return slash < 0 ? obj.ApiVersion : obj.ApiVersion.Substring(slash+1);
+                return slash < 0 ? obj.ApiVersion : obj.ApiVersion.Substring(slash + 1);
             }
             return null;
         }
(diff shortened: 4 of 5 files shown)


src/KubernetesClient/TimeoutHandler.cs
```diff
@@ -0,0 +1,34 @@
+using System;
+using System.Linq;
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.WebUtilities;
+
+namespace informers
+{
+    public class TimeoutHandler : DelegatingHandler
```
brendandburns, line 10: Can you document what this is used for?

macsux: Added description to the class that explains its purpose and how to use

admilazz, line 16: Seems like it could use a better class name, since it sounds like a generic timeout handler but it's watch-specific. Also, is the namespace ("informers") correct?

macsux: I took a second look and realized all this logic belongs in the existing `WatcherDelegatingHandler`. Code has been updated

brendandburns:
This looks good to me, but it needs a rebase.

macsux:
Rebased, should be good to merge now

src/KubernetesClient/WatcherDelegatingHandler.cs
```diff
@@ -12,12 +12,48 @@ namespace k8s
...
+        public WatcherDelegatingHandler(HttpMessageHandler innerHandler, TimeSpan timeout = default) : base(innerHandler)
+        {
+            _timeout = timeout != default ? timeout : DefaultTimeout;
+        }
+
         protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage request,
             CancellationToken cancellationToken)
         {
+            var isWatch =  false;
+			if(request.Method == HttpMethod.Get)
```
brendandburns, line 44: Sorry for the nit, but formatting is off here.

brendandburns:
Sorry one nit related to formatting (we should really add a pre-push hook to validate code format)

macsux:
I sometimes don't notice when tabs get inserted during merge vs spaces, and
they look same in my merge tool. Though I'm against strict lint validation
unless we have a standard tool that is ide neutral to reformat code. I've
been on projects where format compliance was such a chore that a third of
time was spent fixing irrelevant things (extra whitespace at end of lines,
etc). Indentation should be consistent though

brendandburns:
yep, I will be adding in `dotnet format` as a validation so we can just run `dotnet format` in the future.

macsux:
Spacing should be fixed now

tg123:
:) i also add stlyecop to check tabs

src/KubernetesClient/WatcherDelegatingHandler.cs
```diff
@@ -12,19 +12,56 @@ namespace k8s
...
+
+            if (!isWatch)
+            {
+                var cts = new CancellationTokenSource(_timeout);
+                cancellationToken = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, cts.Token).Token;
+            }
+
             var originResponse = await base.SendAsync(request, cancellationToken).ConfigureAwait(false);
 
             if (originResponse.IsSuccessStatusCode && request.Method == HttpMethod.Get) // all watches are GETs, so we can ignore others
```
tg123, line 60: `isWatch` here?

admilazz:
It seems like the effect of this change is that HttpClient.Timeout no longer "works", in that for all non-watch requests you're going to override it with the the value from the timeout handler (defaulting to 100 seconds).

HttpClient.Timeout is something you can control on a per-request basis if you like, whereas changing the timeout via the handler requires constructing a whole new Kubernetes client with a new handler. And actually, since the WatchDelegatingHandler is always added by the Kubernetes client with its default timeout value, there may be no way to increase the timeout anymore. (?)

If the HttpClient is "externally managed", can't the external manager, whoever that is, ensure the timeout and handlers on it are correct? This change doesn't seem to affect externally managed HttpClients, but only the default HttpClient (since the timeout handler is not added to externally provided HttpClients).

src/KubernetesClient/WatcherDelegatingHandler.cs
macsux: Not sure what you meant by your comment here. We're only creating a new timeout CancellationToken if it is NOT watch. Watch requests are meant to be non-expiring.

admilazz: I think he meant to reuse the work you already did to compute isWatch to avoid doing that work a second time for watch requests.

macsux:
> It seems like the effect of this change is that HttpClient.Timeout no longer "works", in that for all non-watch requests you're going to override it with the the value from the timeout handler (defaulting to 100 seconds).

The behavior just "moves" the timeout logic form HttpClient.Timeout property into the handler, but the same semantics apply. You can always set it to Infinite in the handler, and use your own cancellation token for "per request" fine-grained control. However, a request with `watch` already implies that the request is streaming and should be excluded from standard timeout rules, so a dumb logic like the one used in HttpClient.Timeout should not be allowed.

> HttpClient.Timeout is something you can control on a per-request basis if you like, whereas changing the timeout via the handler requires constructing a whole new Kubernetes client with a new handler. And actually, since the WatchDelegatingHandler is always added by the Kubernetes client with its default timeout value, there may be no way to increase the timeout anymore. (?)

That's not true. [See documentation](https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclient.timeout?view=netframework-4.8)
> The same timeout will apply for all requests using this HttpClient instance. You may also set different timeouts for individual requests using a CancellationTokenSource on a task. Note that only the shorter of the two timeouts will apply.


> If the HttpClient is "externally managed", can't the external manager, whoever that is, ensure the timeout and handlers on it are correct? This change doesn't seem to affect externally managed HttpClients, but only the default HttpClient (since the timeout handler is not added to externally provided HttpClients).

Externally managed HttpClient such as the ones created using HttpClientFactories are expected to add `WatcherDelegatingHandler` manually (this is why it was recently exposed via `CreateWatchHandler` static method).

admilazz:
> The behavior just "moves" the timeout logic form HttpClient.Timeout property into the handler, but the same semantics apply.

Right, that's what I mean by "HttpClient.Timeout no longer 'works'".

> You can always set it to Infinite in the handler, and use your own cancellation token for "per request" fine-grained control.

I suspect most people won't know this, and it seems like a breaking change.

> > The same timeout will apply for all requests using this HttpClient instance.

Yes, but you can can change HttpClient.Timeout at any time, so you can change it between requests. You can't change the handler's timeout between requests. You'd have to create a new Kubernetes client with a new handler.

> Externally managed HttpClient such as the ones created using HttpClientFactories are expected to add `WatcherDelegatingHandler` manually (this is why it was recently exposed via `CreateWatchHandler` static method).

Right, but if this change is specifically to support externally managed HttpClients, why does it have to change the behavior of the default, internally managed HttpClient? Would it be possible to provide some utility that people managing their own HttpClients can use without changing the behavior of the default HttpClient? (I think you were aiming for something like that in your original version.)

macsux:
> Yes, but you can can change HttpClient.Timeout at any time, so you can change it between requests. You can't change the handler's timeout between requests. You'd have to create a new Kubernetes client with a new handler.

[HttpClient is intended to be a singleton](https://aspnetmonsters.com/2016/08/2016-08-27-httpclientwrong/) for a given "service type" (which is why HttpClientFactory was built in the first place). The Timeout should NEVER be altered if your code is already using HttpClient as other threads might be trying to change it at the same time.

> Right, but if this change is specifically to support externally managed HttpClients, why does it have to change the behavior of the default, internally managed HttpClient? Would it be possible to provide some utility that people managing their own HttpClients can use without changing the behavior of the default HttpClient? (I think you were aiming for something like that in your original version.)

So this is the part that deeply confuses me because the watch handler actually doesn't timeout in current code, even though it should. [This test](https://github.com/kubernetes-client/csharp/blob/8e7bf0b6f24d7daadf070a29cd170e4c861ab7a5/tests/KubernetesClient.Tests/WatchTests.cs#L343-L414) actually tests for this scenario, and it should fail. I've spent about 2 hours trying to understand WHY it passes right now, and haven't found an answer. There's nothing in existing code that changes watch timeouts. If you change [this line ](https://github.com/kubernetes-client/csharp/blob/8e7bf0b6f24d7daadf070a29cd170e4c861ab7a5/tests/KubernetesClient.Tests/WatchTests.cs#L364) to initialize Kubernetes with external HttpClient, this test will fail. 

But the bottom line is anyone using this library would expect watch requests to not timeout.

admilazz:
> [HttpClient is intended to be a singleton](https://aspnetmonsters.com/2016/08/2016-08-27-httpclientwrong/) for a given "service type" (which is why HttpClientFactory was built in the first place). The Timeout should NEVER be altered if your code is already using HttpClient as other threads might be trying to change it at the same time.

Well, it's completely up to you (i.e. the user) whether there will be other threads trying to change it at the same time. You could, for example, do this:

    c = new Kubernetes(config);
    c.HttpClient.Timeout = x;
    ... use the client as much as you want on whichever threads you want ...

You'd expect it to use timeout `x` but it won't anymore after your change. Or you could use the client in a single-threaded style.

> So this is the part that deeply confuses me because the watch handler actually doesn't timeout in current code, even though it should. [This test](https://github.com/kubernetes-client/csharp/blob/8e7bf0b6f24d7daadf070a29cd170e4c861ab7a5/tests/KubernetesClient.Tests/WatchTests.cs#L343-L414) actually tests for this scenario, and it should fail. I've spent about 2 hours trying to understand WHY it passes right now, and haven't found an answer.

From my vague memory, I thought the reason is that:
1. The timeout only applies to the client reading data from the connection stream, not the connection stream itself.
2. The watch handler captures the connection stream and replaces it with a fake connection stream.
3. The watch handler does some complicated stuff to write the first line of the server response into that fake stream. This line of data is gets read by the generated service client code. The generated client code is happy and tries to deserialize it as a list of items or whatever. And the timeout no longer applies. (Really I think it should skip the complicated stuff and just write a constant "{}" to the fake stream, which should be just as effective.)

macsux:
> From my vague memory, I thought the reason is that:
> 
> The timeout only applies to the client reading data from the connection stream, not the connection stream itself.
> The watch handler captures the connection stream and replaces it with a fake connection stream.
> The watch handler does some complicated stuff to write the first line of the server response into that fake stream. This line of data gets read by the generated service client code. The generated client code is happy and tries to deserialize it as a list of items or whatever. And the timeout no longer applies. (Really I think it should skip the complicated stuff and just write a constant "{}" to the fake stream, which should be just as effective.)
> The watch continues using the captured connection stream for as long as it wants.

This doesn't explain why an externally created HttpClient with a registered `WatcherDelegatingHandler ` times out during watch requests - this is how I encountered this issue in the first place.

admilazz:
> This doesn't explain why an externally created HttpClient with a registered `WatcherDelegatingHandler ` times out during watch requests - this is how I encountered this issue in the first place.

Is the stream-switcheroo hack inside WatcherDelegatingHandler is actually being executed? I.e. if you set a breakpoint inside LineSeparatedHttpContent.SerializeToStreamAsync, is it hit?

macsux:
Ok, I've repeated the test with external HttpClient and it seems to be working fine now, I'm not sure why it works now, but this was definitely an issue when I was writing Informers code. Gonna close this PR as it's not needed at this point

macsux:
I've confirmed that the issue is still present when used with HttpFactory. This code is failing after 100 seconds. The WatchDelegatingHandler is being loaded, but doesn't work. When examining handler stack, HttpClient inserts LogHandler between innermost and Watch, which might be what's causing the underlying issue. I would expect that the below code is how most modern code would attempt to use the library.

```
services.AddHttpClient("DefaultName")
                .AddTypedClient<IKubernetes>((httpClient, serviceProvider) =>
                {
                    return new Kubernetes(config, httpClient);
                })
                .AddHttpMessageHandler(KubernetesClientConfiguration.CreateWatchHandler)
                .ConfigurePrimaryHttpMessageHandler(config.CreateDefaultHttpClientHandler);
```

shaikatz:
Any news about that? we seeing the same behavior with terminated watcher after 100 seconds.

brendandburns:
close/reopen tro try to retrigger the build

brendandburns:
To get this merged, the author needs to run `dotnet format` and then I think it's good to go.

admilazz:
Well, I personally don't like how it breaks HttpClient.Timeout. It seems like something that should be opted into. Since the scenario is intended for "externally managed HttpClients", I'd say that code using external HTTP clients should take the responsibility to set the handlers in a way that works (perhaps using a helper method as in the original iteration of this PR), rather than changing the behavior of the default HttpClient (which presumably doesn't need this change to work?).

brendandburns:
I think I agree w/ @admilazz that I'd rather not change the default behavior, it seems like we can make this opt-in without much loss of user experience.

macsux:
Here's my take on this

The way it currently works is very much a hack that seems to trick
httpclient, and seems doesn't work as expected in all scenarios. I'm OK
reverting it to original version of PR, however it puts burden on external
consumers to understand internal implementation so they properly configure
httpclient (set timeout to infinite and use handler).

Imo external management of httpclient is preferred these days. If user
spawns up kubernetes client in transient kind of way, it will create new
instances of httpclient which has been shown to have serious negative
affects on app performance due to socket exhaust. This is the whole reason
why httpfactory was created by Microsoft.

I feel this will negatively affect much larger percentage of users then
whatever negative effect would be of highjacking timeout property, and it
doesn't surface immediately due to long default timeout . I would rather
put bulk of users on happy path then have them discover this quirk they try
to push to prod.

@brendandburns should I still go ahead and refactor to original version of
this PR?

PR #427, opened 2020-04-20
Fix issue with WatchDispose test

macsux:
This fixes an incorrect assertion which assumed there are no received events after Dispose on Watch is called. This is incorrect because the way cancellation token is used doesn't throw, so it's possible to have items being processed and published as events even after Dispose is called.

### DIFF
--- a/tests/KubernetesClient.Tests/WatchTests.cs
+++ b/tests/KubernetesClient.Tests/WatchTests.cs
@@ -263,7 +263,6 @@ namespace k8s.Tests
                     await Task.Yield();
                 }
 
-                Assert.Empty(events);
                 Assert.False(watcher.Watching);
                 Assert.True(connectionClosed.IsSet);
             }


fejta:
prow, this is already merged

tg123:
no event should rise in watcher after disposed
should fix watcher rather than removing assert from testcase?

macsux:
I'm confused as to why you need to offer guarantee on this. You're using IDisposable for what should have been a CancellationToken (internally it is actually). CancellationTokens always had "best effort" rather than immediate guarantees. The only way to make this guarantee is to use locks, because even if you were to ask for a TaskCancelledException, it might still be in middle of marshalling and cause new event to be emitted millisecond after you call dispose. 

This is why both TPL dataflow, and System.Threading.Channels use a combination of `CancellationToken` and `Completed` task you can await on to offer this kind of guarantees, which imo we don't need here.

Issue #433, opened 2020-04-23
Code styling issues


macsux:
With the recent introduction of code style rules, I wanted to bring up a few issues I hope we can address:

1. Some types of checks are not reformatted using `dotnet format`. This causes a significant loss in productivity trying to fix up formatting issues. Examples of where auto formatted doesn't work is a requirement for extract lines after braces. If there's another tool that can "auto-format", I think it should be put into README.md as guidance
2. I disagree with `SA1503: Braces should not be ommited` for `if` statements. A lot of validation checks such as `if (type == null) throw new ArgumentNullException(nameof(type));` turn into 4 lines of code
```
            if (type == null)
            {
                throw new ArgumentNullException(nameof(type));
            }

```
which only detracts from the readability of code. I propose this rule is removed
3. There are rules in there right now to validate things like namespace naming convention which `k8s` is failing
4. Legitimate warnings are being drowned in the formatting "noise" on build atm, some of which we won't be able to address (like namespace naming). 
5. I'm also noticing a VERY significant degradation of IDE performance with the latest style changes, as I suspect there's a ton of background threads trying to run Roslyn analyzers. 
6. In some cases, the code actually becomes less efficient. Example: JObject.Contains is defined explicitly in the newer version but older the one that targets 4.5.2. The current rule makes you cast it to IDictionary as you're not supposed to access it implicitly, meaning newer targets are using less optimized code path. 

**I think there needs to be a balance struck on which rules to enforce, as previous experience with projects that enforced very strict linting rules without the ability to automatically reformat the code has discouraged contributions as it turned into a major chore.**

macsux:
So on a personal note, this is big enough of an issue for me to not want to contribute going forward. I'm fine with some styling rules, but this just ridiculous. I just spent half a day going through some existing PRs fixing line breaks and other things to make it conform to the new rules, but at this point 
1. My IDE is slow
2. Much of the existing code became less readable
3. Build is full of warnings that are unrelated to my code or something that can or should be addressed
4. Existing style rules of Rider/Resharper conflict with those in the project - it formats code in a way that conflicts

tg123:
sorry and I totally understand your feeling
absolutely we can discuss any enforced rules or code style

this project was once with resharper code check but removed due to build system reasons #274 

I am moving some MS internal practice to this project as we have many styles in this project. those enforced rules are/will base on those internal rules.

1 I am working to clean up remaining warning (~4K) and have them enforced, warning or suppressed.

[...]

PR #435, opened 2020-04-23
make sa1503 rule default

tg123:
First of first 
developers must be happy

I think we can disable this first and think about it carefully
sorry for enforce my own code style option and experience 

is it better with you @macsux? #433


brendandburns:
Let's hold on this. I actually believe this is a good style rule.

PR #440, opened 2020-05-06
add TokenFileAUth which reloads token if it expires

### LINKED ISSUE #293: client should periodically reload InClusterConfig token
See https://github.com/kubernetes/kubernetes/issues/70679

zshihang:
fixes #293. by default, this will set an expiration period of 1 minutes on service account token. as we are graduating the Beta feature Projected Service Account Volume to GA, we expect clients to reload token before it expires.


src/KubernetesClient/Authentication/TokenFileAuth.cs
```diff
@@ -0,0 +1,31 @@
...
+        {
+            _token_file = token_file;
+        }
+
+        public async Task<AuthenticationHeaderValue> GetAuthenticationHeaderAsync(CancellationToken cancellationToken)
+        {
+            if (_token_expires_at < DateTime.UtcNow)
+            {
+                _token = File.ReadAllText(_token_file).Trim();
+                _token_expires_at = DateTime.UtcNow.AddSeconds(60);
```
brendandburns, line 26: Please add a comment about why 60 seconds is the right value.

src/KubernetesClient/Authentication/TokenFileAuth.cs
```diff
@@ -0,0 +1,31 @@
+using System;
+using System.Net.Http.Headers;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Rest;
+
+namespace k8s.Authentication
+{
+    public class TokenFileAuth : ITokenProvider
```
brendandburns, line 10: Please add a unit test for this file.

brendandburns: oh, you already did, nevermind.

tests/KubernetesClient.Tests/TokenFileAuth.cs
```diff
@@ -0,0 +1,31 @@
...
+using System.Threading.Tasks;
+using FluentAssertions;
+using k8s.Authentication;
+using Xunit;
+
+namespace k8s.Tests
+{
+    public class TokenFileAuthTests
+    {
+        [OperatingSystemDependentFact(Exclude = OperatingSystem.OSX | OperatingSystem.Windows)]
```
brendandburns, line 12: Why is this excluded on OSX or Windows? This test should definitely work on all platforms.

brendandburns:
Couple of small comments.

src/KubernetesClient/Authentication/TokenFileAuth.cs
zshihang: done

tests/KubernetesClient.Tests/TokenFileAuth.cs
zshihang: removed. i only had linux at hand to verify the test.

PR #442, opened 2020-05-10
Quoted non-string values remain quoted after serialization

### LINKED ISSUE #387: Environment variable values are not coerced to strings during serialization
Hello, I am currently manipulating yaml configurations and ran into an issue where environment variable values containing ints and bools were not correctly being coerced to strings.

I have created a failing test case here: https://github.com/ridhoq/csharp/blob/201e2be09b1fb30fe86cc68ff65aaeb4f1141a21/tests/KubernetesClient.Tests/YamlTests.cs#L325

The test output results in:
 ```
k8s.Tests.YamlTests.EnvVariableShouldBeStrings

Assert.Equal() Failure
                                  ↓ (pos 111)
Expected: ···: PORT\n      value: "3000"\n    image: vish/stress\n    name: c···
Actual:   ···: PORT\n      value: 3000\n    image: vish/stress\n    name: cpu···
                                  ↑ (pos 111)
   at k8s.Tests.YamlTests.EnvVariableShouldBeStrings() in /Users/ridwan/git/csharp/tests/KubernetesClient.Tests/YamlTests.cs:line 343
```
In this case, the integer value in an environment variable value will break yaml validation.

1. Can a maintainer help me confirm that this is actually a bug?
2. I am happy to submit a PR if it is a bug. In order to submit a fix, I need to better understand how `YamlDotNet` is being used in the `SaveToString<T>(T value)` method. Does anybody have any pointers on how to proceed debugging this?

**EDIT** - clarifying that the issue is for environment variables specifically

ridhoq:
As seen in https://github.com/aaubry/YamlDotNet/issues/33, quoted `null`/`int`/`bool` values are not coerced to strings correctly. @gabriel-samfira authored an upstream  fix to this in https://github.com/cloudbase/powershell-yaml/pull/41 that allows values that match the above types to be serialized correctly.

This PR adapts @gabriel-samfira's work into this library to ensure these values are serialized correctly into valid kubernetes YAML configuration.

Closes #387

### DIFF
--- a/src/KubernetesClient/StringQuotingEmitter.cs
+++ b/src/KubernetesClient/StringQuotingEmitter.cs
@@ -6,6 +6,7 @@ using YamlDotNet.Serialization.EventEmitters;
 
 namespace k8s
 {
+    // adapted from https://github.com/cloudbase/powershell-yaml/blob/master/powershell-yaml.psm1
     public class StringQuotingEmitter : ChainedEventEmitter
     {
         // Patterns from https://yaml.org/spec/1.2/spec.html#id2804356
--- a/src/KubernetesClient/StringQuotingEmitter.cs
+++ b/src/KubernetesClient/StringQuotingEmitter.cs
@@ -26,7 +26,7 @@ namespace k8s
             switch (typeCode)
             {
                 case TypeCode.Char:
-                    if (char.IsDigit((char) eventInfo.Source.Value))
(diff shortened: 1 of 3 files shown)


tg123:
why not update YAML lib to get it fixed?
we can keep testcase for sure

brendandburns:
@tg123 afaik there's no fix in YamlDotNet, only the documentation of the issue.

PR #444, opened 2020-05-11
Fix reading incorrect stream in ExecuteExternalCommand

Tyrrrz:
Also, reading the streams sequentially may result in a deadlock.

Consider using [CliWrap](https://github.com/Tyrrrz/CliWrap) for handling external processes.


brendandburns:
/lgtm
/approve

git hub actions appear hung

Issue #396, opened 2020-04-01
Failure to parse AWS generated kube config


seanamosw:
I saw the initial implementation for supporting the `user/exec` config was done in https://github.com/kubernetes-client/csharp/pull/359

However, it appears that it is still unable to parse the `exec/env` section.

This appears to be how the model is specified in https://github.com/kubernetes-client/csharp/blob/master/src/KubernetesClient/KubeConfigModels/ExternalExecution.cs

```csharp
        [YamlMember(Alias = "env")]
        public IDictionary<string, string> EnvironmentVariables { get; set; }
```

That would parse yaml with this format:
```yaml
env:
  name1: value1
  name2: value2
```

However the actual format is:
```yaml
env:
   - name: name1
     value: value1
   - name: name2
     value: value2
```
I believe this needs to at least be changed to something like this:

```csharp
        [YamlMember(Alias = "env")]
        public IList<IDictionary<string, string>> EnvironmentVariables { get; set; }
```

Might want to further change it to something like:
```csharp
        public IList<NameValuePair> EnvironmentVariables { get; set; }
```

brendandburns:
We'd be happy to have a PR to fix this.

Thanks!

PR #359, opened 2020-02-20
Out-of-tree client authentication providers (UserCredentials exec option) for asp.net core applications

talboren:
after facing the same problem described in: https://github.com/kubernetes-client/csharp/issues/340
I've tried to create a version that will allow me to receive the token using an external command execution (as kubeconfig supports).

Implemented according to:
https://github.com/kubernetes/community/blob/master/contributors/design-proposals/auth/kubectl-exec-plugins.md
and with inspiration from the kubernetes python implementation (https://github.com/kubernetes-client/python-base/blob/master/config/exec_provider.py)

**1. I haven't written tests
2. Token refreshment hasn't been implemented meaning that once the token is expired a new token has to be generated, I am currently using my kubernetes "access layer" from the DI with a scoped life time cycle.
3. I've tested it running on Linux (_Amazon Linux 2_) only, not sure if it works properly on Windows/any other creature ;)**

Would appreciate help finishing and wrapping this up for usage.

PR #451, opened 2020-05-22
Removed NETSTANDARD2_0 IFDefs and added Brackets

KyleJFischer:
Removed NETSTANDARD2_) IfDefs to allow later versions of .net to be able to use the external execution feature. 

Also added { because it won't let me build if I didn't fix them. Can remove them if we want it the old way.


KyleJFischer:
@k8s-ci-robot /check-cla

KyleJFischer:
Sorry about all the extra commits, first time contributing. Let me know if there is anything else you need me to change!

src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
```diff
@@ -411,7 +407,6 @@ public static string RenewAzureToken(string tenantId, string clientId, string ap
             throw new KubeConfigException("Refresh not supported.");
         }
 
-#if NETSTANDARD2_0
```
NasAmin, line 414: Not sure what impact this will have in terms for backwards compatibility.
Will this cause any harm if this block of code runs for non .Net core clients?

KyleJFischer: I'll try running some more testing on this in a bit. If others could test different setups as well, that would probably be the most ideal.

tg123:
I think it is OK to remove this build condition.

add in #359
ping @talboren for more info

talboren:
> I think it is OK to remove this build condition.
> 
> add in #359
> ping @talboren for more info

At the time I had several problems with these tests that required the preprocessor directives, not quite sure what will be the impact of removing these but it seems that build passes. 

@tg123 you mentioned before that you're working on migrating the codebase to .netstandard2.0, how's that going? anything we can help with?

btw, not sure it has something to do with this specific pull request but @KyleJFischer please bare in mind the following issue: https://github.com/kubernetes-client/csharp/issues/396 (although I haven't suffered from the problems mentioned in there)

src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
```diff
@@ -432,14 +427,12 @@ public static string ExecuteExternalCommand(ExternalExecution config)
...
-            process.StartInfo.Environment.Add("KUBERNETES_EXEC_INFO",
-                JsonConvert.SerializeObject(execInfo));
-
+            process.StartInfo.EnvironmentVariables.Add("KUBERNETES_EXEC_INFO", JsonConvert.SerializeObject(execInfo));
             if (config.EnvironmentVariables != null)
             {
                 foreach (var configEnvironmentVariableKey in config.EnvironmentVariables.Keys)
                 {
-                    process.StartInfo.Environment.Add(key: configEnvironmentVariableKey,
+                    process.StartInfo.EnvironmentVariables.Add(key: configEnvironmentVariableKey,
```
tg123, line 435: I checked the doc
EnvironmentVariables === Environment

just curious why this was changed

tg123:
/LGTM

already moved to netstandard2
thanks @talboren

src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
KyleJFischer: If you are asking why I changed it, I had read above the target .net framework for this client was 4.5.2 which unfortunately doesn't have the Environment. However, EnvironmentVariables does exist in 4.5.2 and is in the latest preview of .net 5. [Source](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.processstartinfo.environmentvariables?view=netcore-3.1)

If you are asking why it changed in general, well, I can't help you there :) 

Hopefully that clears up any confusion.

NasAmin:
Is there anything else specific this PR is waiting on? I'd really appreciate if it can merged.
@KyleJFischer I have tried your PR changes locally to pull deployments and namespaces from multiple clusters on .NET Core 3.1 project. It works perfectly.

Issue #416, opened 2020-04-15
No implementation for kubectl top node / top pod


RSBlek:
For simple logging / monitoring there should be a method to get the cpu and ram metrics of nodes and pods that are provided by the kubectl top node &kubectl top pod commands.

I couldn't find a direct call to that command. Is there a possible workaround that i can use to get the metrics of a node?

brendandburns:
There's no specific API call for `kubectl top` it's actually a pretty complicated piece of logic inside of kubectl.

You can use the CustomObjects API to pull the metrics directly from the metrics server.

There's a discussion here:
https://github.com/kubernetes-client/java/issues/675

with pointers to the relevant `kubectl` code.

We'd take a PR for this if you want to write it.

RSBlek:
@brendanburns Thanks, the CustomObjects API in conjunction with the metrics server was just what i was looking for. 

For readers:

The following can be used as "quickfix" to get the node metrics. I think that reading the pod metrics will work the same

```
    public class NodeMetrics
    {
        [JsonProperty(PropertyName = "metadata")]
        public V1ObjectMeta Metadata { get; set; }

        [JsonProperty(PropertyName = "timestamp")]
        public DateTime Timestamp { get; set; }

        [JsonProperty(PropertyName = "window")]
        public String Window { get; set; }

        [JsonProperty(PropertyName = "usage")]
        public IDictionary<string, ResourceQuantity> Usage { get; set; }
    }
```
```
    public class NodeMetricsList
    {
        [JsonProperty(PropertyName = "apiVersion")]
        public string ApiVersion { get; set; }

        [JsonProperty(PropertyName = "kind")]
        public string Kind { get; set; }

        [JsonProperty(PropertyName = "metadata")]
        public V1ObjectMeta Metadata { get; set; }

        [JsonProperty(PropertyName = "items")]
        public IEnumerable<NodeMetrics> Items { get; set; }
    }
```
```
Kubernetes client = ...
JObject customObject = (JObject)await client.GetClusterCustomObjectAsync("metrics.k8s.io", "v1beta1", "nodes", "");
NodeMetricsList nodeMetricsList = customObject.ToObject<NodeMetricsList>();
```

brendandburns:
Cool, glad it worked! If you want to get this into a PR, we'd happily take the addition.

alarco-l:
If it's up to grab, i'm willing to help on this one as first contribution.

[...]

PR #466, opened 2020-07-25
Nodes and pods metrics

alarco-l:
Implementation for usage metrics pulled from metrics server.

Try to address #416

Using as [metrics server API reference](https://github.com/kubernetes/metrics/blob/master/pkg/apis/metrics/v1beta1/types.go)

 - [x] Nodes metrics
 - [x] Pods metrics
 - [x] Pods metrics by namespace
 - [x] Add sample
 - [x] Add tests


brendandburns:
Super cool! Thanks for the PR!

It is simple enough that I'm ok to merge w/o tests and add them in a second PR, or you can add them to this PR. Your call.

alarco-l:
You can merge it and I will open another one for the tests.

Glad to be able to contribute.

alarco-l:
@brendandburns, I could add the tests on this one.

However, not sure this is the right way to test it.
If you have some insights on that, it could help me.

brendandburns:
Let's merge and then add tests.

Basically you will want to mock out the web server and then test that the client code works correctly to retrieve metrics.

/lgtm
/approve

PR #471, opened 2020-08-08
Nodes and pods metrics tests

alarco-l:
Add more tests for nodes and pods metrics as discussed with @brendandburns.

Use of `MockKubeApiServer` and copy payload response form `minikube`.

See #466


brendandburns:
/lgtm
/approve

Thanks for adding more tests!

Issue #464, opened 2020-07-17
Is there a C-Sharp method/code-snippet to upload a file from my local machine to a pod?


Mahmoud-Karam:

brendandburns:
There's not a utility currently, the equivalent Java code is here:

https://github.com/kubernetes-client/java/blob/master/util/src/main/java/io/kubernetes/client/Copy.java#L41

Which should be pretty easy to adapt to C#.

We would gladly take a PR adding this support too.

Mahmoud-Karam:
Thanks a lot @brendanburns

lazarkrstic:
@Mahmoud-Karam  Can you post your code snippet for C#?

Mahmoud-Karam:
> @Mahmoud-Karam Can you post your code snippet for C#?

@lazarkrstic I didn't work on it unfortunately.

lazarkrstic:
Solved in https://github.com/kubernetes-client/csharp/issues/949#issuecomment-1190083366

PR #472, opened 2020-08-09
Copy to/from pod

alarco-l:
Implementation for copy from/to a pod.

Important note, to make a copy, it requires that the `tar binary` is present in your container image, like the `kubectl cp` command.

Try to address #464 

- [ ] Copy file from a pod
- [ ] Copy directory from a pod
- [ ] Copy file to a pod
- [ ] Copy directory to a pod

### DIFF
--- a/src/KubernetesClient/IKubernetes.Copy.cs
+++ b/src/KubernetesClient/IKubernetes.Copy.cs
@@ -140,10 +140,10 @@ namespace k8s
         /// The container in which the source directory exist.
         /// </param>
         /// <param name="sourceFolderPath">
-        /// The source file path.
+        /// The source directory path.
         /// </param>
         /// <param name="destinationFolderPath">
-        /// The destination file path.
+        /// The destination directory path.
         /// </param>
         /// <param name="cancellationToken">
         /// A <see cref="CancellationToken"/> which can be used to cancel the asynchronous operation.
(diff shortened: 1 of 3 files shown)


brendandburns:
cool, excited to see this. Let me know when it is ready to review!

txr150430:
I have a similar requirement. Thanks Ludovic Alarcon and @brendanburns  for getting this effort into our K8S c# client.

alarco-l:
@brendandburns I got stuck on an issue since few days, I hope you can enlighten me.
In `CopyDirectoryToPodAsync` or `CopyFileToPodAsync` when I try to read if there is some error on stdErr

```csharp
using (var reader= new StreamReader(stdError))
{
    if (errorReader.Peek() != -1)
    {
      .....
    }
}
```

The StreamReader hangs on `Peek()`, same with `ReadLineAsync()` and `ReadToEndAsync()`.

If I don't try to check from error, the hang will happen in `Kubernetes.Exec.cs` also on the StreamReader.

I cannot figure out where is my mistake...

alarco-l:
Hello @brendandburns,

I'm stuck on the copy to pod part, my actual code get stuck when it try to use the `peek()` method from stream.

In my understanding, it's because I write the tar input on the `standard input` by doing a pipe with `base64` command. 
On a Linux machine the way to tell it the end, it's to do a `ctrl + d`.
If I understand correctly, when a stream is close and/or disposed it's equivalent to `ctrl + d`.
But I cannot close the standard input the the `exec callback`.

If the process is kill after being stuck either small or big and text or binary, the copy is successful.

So for unblocking the process, I tried unsuccessfully:
- Write `0x04` two times, on the stream (the ascii code for `ctrl + d`) but stream still stuck
- Use `base64 -d << EOT | tar xzmf - -C FOLDER` but ending with a broken pipe (both tty and not tty mode)
- Use tty and write two times `0x04` but it only works with small files, big files ends with `truncated input`.
- I tried to put a delay between writing data and write `0x04`

Do you have any input that can help ?
Am I trying to copy the right way ? (In my understanding of kubectl and java client code I think it's correct)

I'm out of idea... I think I made a stupid mistake but I'm too deep inside the code to see it

hol430:
Hi all, I need this feature for a project which I'm working on. I've checked out this branch and was able to get the CopyFileToPodAsync() function to work by removing the base64 encoding/decoding layer and avoiding the invocation of the shell altogether. As far as I can tell, the kubectl doesn't seem to encode to base64 before sending it off to the pod, while the java version does do this - not sure why. I'm sure I could get it working with more fiddling, but if it's unnecessary then I don't see much point. When I tried to use the shell I ran into the problems mentioned above.

I'm not sure what the process is for reviving this PR - could someone let me know? I've attached my patch to this message if anyone's interested. I haven't tested the copying of entire directory, or the copying of files/directories from pod to local system, however I believe this approach could be suitable.

[copy.patch.txt](https://github.com/kubernetes-client/csharp/files/6866966/copy.patch.txt)

brendandburns:
@hol430 please feel free to craft/send a new PR if you'd like.

Thanks!

hol430:
Thanks @brendandburns - I don't seem to have push access on this branch, so I've opened #671, which should supersede this one.

tg123:
close as moved to #671

PR #479, opened 2020-09-17
Parse clientCertificateData/clientKeyData from auth plugins

Frassle:
Currently ExecuteExternalCommand crashes with a key not found error if
it runs a certificate based (rather than token based) plugin. This
commit will now return either the token string, or the certificate and
key strings which are then used to set the relevant fields on the
configuration object.

### DIFF
--- a/src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
+++ b/src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
@@ -389,7 +389,7 @@ namespace k8s
                     throw new KubeConfigException("External command execution missing ApiVersion key");
                 }
 
-                ExecuteExternalCommand(userDetails.UserCredentials.ExternalExecution, out var accessToken, out var clientCertificateData, out var clientCertificateKeyData);
+                var (accessToken, clientCertificateData, clientCertificateKeyData) = ExecuteExternalCommand(userDetails.UserCredentials.ExternalExecution);
                 AccessToken = accessToken;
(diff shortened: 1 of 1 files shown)


Frassle:
I signed it

Frassle:
I signed it

src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
```diff
@@ -458,8 +460,10 @@ public static Process CreateRunnableExternalProcess(ExternalExecution config)
         /// https://github.com/kubernetes-client/python-base/blob/master/config/exec_provider.py
         /// </summary>
         /// <param name="config">The external command execution configuration</param>
-        /// <returns>The token received from the external command execution</returns>
-        public static string ExecuteExternalCommand(ExternalExecution config)
+        /// <param name="token">The token received from the external command execution</param>
+        /// <param name="clientCertificateData">The client certificate data received from the external command execution</param>
+        /// <param name="clientKeyData">The client key data received from the external command execution</param>
+        public static void ExecuteExternalCommand(ExternalExecution config, out string token, out string clientCertificateData, out string clientKeyData)
```
tg123, line 466: prefer new return style return`(string, string, string)`
seems more readable

tg123 (approved):

src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
Frassle: Sure

PR #480, opened 2020-09-19
Fix format of ClientCertificate[Key]Data from auth plugins

Frassle:
ClientCertificateData and ClientCertificateKeyData are expected to be base64 encoded strings by CertUtils.GeneratePfx. However auth plugins return the actual "--BEGIN X---...---END X---" strings, this commit base64 encodes that to match as if it was read from the config file.

### DIFF
--- a/src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
+++ b/src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
@@ -391,8 +391,11 @@ namespace k8s
 
                 var (accessToken, clientCertificateData, clientCertificateKeyData) = ExecuteExternalCommand(userDetails.UserCredentials.ExternalExecution);
                 AccessToken = accessToken;
-                ClientCertificateData = clientCertificateData;
-                ClientCertificateKeyData = clientCertificateKeyData;
+                // When reading ClientCertificateData from a config file it will be base64 encoded, and code later in the system (see CertUtils.GeneratePfx)
+                // expects ClientCertificateData and ClientCertificateKeyData to be base64 encoded because of this. However the string returned by external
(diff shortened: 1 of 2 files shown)


Frassle:
I missed this distinction of how these strings we're formatted in my last PR (#479).

/assign brendandburns

brendandburns:
This looks good, can you add a unit test to validate the behavior and protect against future breakage?

Thanks!

Frassle:
> can you add a unit test 

I did have a look at this quickly and wasn't sure the best way to unit test. We're calling into System.Diagnostics.Process and I couldn't think of any easy way to mock that out for a test. I could make a config that has command set to cmd.exe or /bin/sh with arguments that just cause it to echo back an expected json blob, but that seemed a big ugly.

brendandburns:
In other contexts, I have just used `echo` to echo back the json blob. I think that's actually the right way to do it.

Thanks!

Frassle:
I've added another test to the auth tests copying bits of the external token test and the certificate tests.

N.B The external auth test looks like it would never run because it was ifdefed for NETSTANDARD_2_0 but the tests either run as netcoreapp2.0 or netcoreapp2.1 so I've removed that ifdef

brendandburns:
Thanks, looks like the test is failing though with some JSON parse error...

Frassle:
Looking in to it, I only checked it on my windows machine (removing the exclude clause to get it to run, seemed fine but I'm going to trust the comment that cert use is intermittent)

Frassle:
Seems the problem is these external tests weren't ever running so the echo handling is wrong. I've raised https://github.com/kubernetes-client/csharp/pull/484 as a smaller PR to just enable the one existing external test and get it working. After that's merged I'll merge master into here and it should fix up this PR as well.

brendandburns:
Cool, thanks for digging in here. I merged #484 so this should be ready once you rebase.

PR #493, opened 2020-10-07
WIP: Update generated code, throw on 401 errors

### LINKED ISSUE #491: No error returned when credentials are incorrect
When calling a method with basic authentication (login and password), if the credentials are incorrect, there is no exception. An example with the following will just produce a `ns` variable with a `null` value:

```
KubernetesClientConfiguration config = new KubernetesClientConfiguration
{
    Username = "xxx",
    Password = "yyy",
    Host = "https://zzz:6443",
    SkipTlsVerify = true
};
Kubernetes client = new Kubernetes(config);
V1Namespace ns = client.ReadNamespace("test");
```

brendandburns:
Fixes #491 

cc @tg123


brendandburns:
Hrm, it looks like this reverts `V1Patch` to `object` I think that shouldn't happen, I'll need to fix that before we can merge this.

tg123:
swagger.json changed?

brendandburns:
Closing this since #505 merged and has the same changes.

PR #504, opened 2020-10-24
Update YamlDotNet to 8.1.2

brendandburns:


brendandburns:
Looks like the default serialization behavior for YamlDotNet changed in 8.0.0, I fixed the serializer to emit nulls (which is the expected behavior for Kubernetes objects)

Issue #476, opened 2020-08-24
PatchNamespacedCustomObjectWithHttpMessagesAsync fails with the latest nuget 2.0.29.


aronip:
As per PR #385, there has been a change in PatchNamespacedCustomObjectWithHttpMessagesAsync method to use JsonPatch as the PatchType, instead of MergePatch which was the behavior before.

![image](https://user-images.githubusercontent.com/2344772/91052196-c07d0100-e63e-11ea-9988-343c0e1cc765.png)

Due to this in our existing implementation, we are hitting the below error while patching a namespaced custom object:
{"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"json: **cannot unmarshal object into Go value of type jsonpatch.Patch**","reason":"BadRequest","code":400}.

The client expects that PatchType must be JsonPatch now, but in our scenario we need the support for MergePatch. Would it be possible to revert to the existing behavior, prior to #385? Or, can PatchType be added as an input for these methods?

brendandburns:
Yeah, this is unfortunate, I would have suggested passing in the `Content-Type` in the custom headers parameter, but looking at the generated code, the content-type is overridden if if it is passed in.

That seems like a bug in AutoRest (the code generator in question) since at this point you can't customize the content-type. I'll follow up with that project and see if we can fix things.

gaurav137:
@brendandburns any update on AutoRest project making this fix?

brendandburns:
I checked w/ the AutoRest team, it looks like a fix may not be forthcoming quickly, so I will see about manually editing this file to revert to the default behavior while we look to see if there is a more permanent fix.

PR #506, opened 2020-10-27
Revert the change from `json-patch` to `merge-patch`

brendandburns:
Fixes https://github.com/kubernetes-client/csharp/issues/476 for now.


tg123:
so fixed by `s/json-path/merge-path/g` ?
this is due to autorest?

maybe by adding it to gen

brendandburns:
@tg123 I'm thinking about adding it to the `gen` repo, but I'm trying to sort out a better answer w/ the autorest folks, this is a temporary patch until I figure out what we're doing.

(and yes `perl -pi -s/json-patch/merge-patch/g Kubernetes.cs` is how I made this change)

tg123:
/LGTM

btw, there was `sed` in https://github.com/kubernetes-client/gen/blob/master/openapi/csharp.sh#L62 already

PR #513, opened 2020-11-01
E2E test with minikube on gh action

tg123:
This PR enable tests against a minikube server to cover some scenarios cannot easily simulate in unittest .

Here the tests introduced with this PR are very simple, just create a pod.
More are coming for recent issues


brendandburns:
/lgtm
/approve

Thanks!

PR #526, opened 2020-11-23
move integration test to minikube proj

tg123:


brendandburns:
Looks good to me, but it looks like you need to run `dotnet format`

tg123:
fixed

PR #535, opened 2020-12-09
[WIP] Proposal: deprecate net452 for 4.0

tg123:
.net client supports `net452` for a long time #32 
however, it creates some maintenance problems:
- strong name https://github.com/dotnet/runtime/blob/master/docs/project/strong-name-signing.md
- some websocket api missing
- some certificate api missing

older sdks also introduce some difficulty to create even more amazing features. 

net452 will be eol with Windows Server 2012 on Oct 10, 2023
legacy users can still use 3.0 .net client (generated based on K8S 1.19) till 2023

this change is planned to be shipped together with client 4.0 (K8S 1.20)


brendandburns:
Works for me.

Issue #538, opened 2020-12-13
It would be good to have an example showing how to handle custom resources


akanso:
It would be good to have a custom resource example showing how the C# client is:
 `creating/deleting/updating/listing` custom resource

brendandburns:
Closing because #540 is merging.

PR #539, opened 2020-12-13
CR example

akanso:
targeting issue: #538

This example creates, lists, Gets, updates and deletes a custom resource.


brendandburns:
Thanks for the PR. We need you to sign the CLA and then we can merge this.

akanso:
> Thanks for the PR. We need you to sign the CLA and then we can merge this.

Thanks, I already did.

I realized my commits where under a different email, that's why there is one requirement still missing

examples/customResource/customResource.sln
```diff
@@ -0,0 +1,18 @@
+﻿
```
tg123, line 1: please add this proj to root sln

tg123 (approved):

examples/customResource/customResource.sln
akanso: done!

brendandburns:
@akanso this is still not quite right, I don't think. One of the commits has 'Ubuntu' as the author, I suspect that is the problem.

akanso:
closing in favor of #540

PR #543, opened 2021-01-05
fix delay

tg123:
cc @timmydo

### DIFF
--- a/src/KubernetesClient/LeaderElection/LeaderElector.cs
+++ b/src/KubernetesClient/LeaderElection/LeaderElector.cs
@@ -189,7 +189,7 @@ namespace k8s.LeaderElection
 
         private async Task AcquireAsync(CancellationToken cancellationToken)
         {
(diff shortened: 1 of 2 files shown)


src/KubernetesClient/LeaderElection/LeaderElector.cs
```diff
@@ -189,11 +189,11 @@ private async Task<bool> TryAcquireOrRenew(CancellationToken cancellationToken)
 
         private async Task AcquireAsync(CancellationToken cancellationToken)
         {
+            var delay = config.RetryPeriod.Milliseconds;
```
timmydo: you can remove `.Milliseconds`, then use `delay *= JitterFactor`

tg123: https://docs.microsoft.com/en-us/dotnet/api/system.timespan.op_multiply

`*` was introduced after netstandard 2.1
have to revert :(

Issue #524, opened 2020-11-19
User: user@domain.com does not have appropriate auth credentials in kubeconfig


daniv-msft:
Hello,

I'm working on the [Bridge to Kubernetes](https://github.com/Microsoft/mindaro) product, and one of our customers reported an exception recently that seems to be coming from the C# KubernetesClient:
```
k8s.Exceptions.KubeConfigException: {
	"StackTrace":"
		at k8s.KubernetesClientConfiguration.SetUserDetails(K8SConfiguration k8SConfig, Context activeContext)
		at k8s.KubernetesClientConfiguration.InitializeContext(K8SConfiguration k8SConfig, String currentContext)
		at k8s.KubernetesClientConfiguration.GetKubernetesClientConfiguration(String currentContext, String masterUrl, K8SConfiguration k8SConfig)
		at k8s.KubernetesClientConfiguration.BuildConfigFromConfigObject(K8SConfiguration k8SConfig, String currentContext, String masterUrl)
		at Microsoft.DevSpaces.Common.Kubernetes.K8sClientFactory.CreateFromKubeConfig(K8SConfiguration kubeConfig)",
	"Message":"User: user@domain.com does not have appropriate auth credentials in kubeconfig",
}
```

The customer is using an OIDC authprovider ([keycloak](https://www.keycloak.org/)).
The kubeconfig used seems to be valid. Please note that the PII data has been removed (same thing above in the exception):
```
apiVersion: v1
clusters:
- cluster:
    certificate-authority-data: <certificate-authority-data>
    server: https://api.test.<cluster-domain>
  name: test.<cluster-domain>
contexts:
    cluster: test.<cluster-domain>
    namespace: <namespace>
    user: <user@domain.com>
  name: test.<cluster-domain>
current-context: test.<cluster-domain>
kind: Config
preferences: {}
users:
- name: <user@domain.com>
  user:
    auth-provider:
      config:
        client-id: kubernetes
        client-secret: ""
        id-token: <id-token>
        idp-issuer-url: https://<auth-domain>/auth/realms/master
        refresh-token: <refresh-token>
      name: oidc
```
Also, the user mentions that running `kubectl` commands work with this same kubeconfig.

Could you please let me know if you have an idea of what might cause this?

tg123:
oidc is not supported at the moment :(

daniv-msft:
All right, thanks @tg123 for your quick reply and clarifying this.

PR #544, opened 2021-01-07
OIDC support

arbielsk:
**Feature**
Add support for OpenID Connect. This includes a new OidcTokenProvider which is able to refresh tokens. A new dependency is introduced in the form of [IdentityModel.OidcClient](https://github.com/IdentityModel/IdentityModel.OidcClient).
While at it, I fixed a few typos in comments in src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs

**Motivation**
At my company we are avid users of [Bridge-To-Kubernetes](https://github.com/microsoft/mindaro) which uses the C# K8S Client under the hood. We currently have multiple teams that are blocked from using Bridge-To-Kubernetes since the [C# K8S client currently does not support OIDC](https://github.com/kubernetes-client/csharp/issues/524). I therefore created this PR with the intent of adding OIDC support to C# K8S client so it can be integrated into Bridge-To-Kubernetes.

**Disclaimer**
I am not a C# developer. In fact this is the first C# code I have ever written. I therefore would like to ask for a thorough review; I am more than happy to update the PR to conform with any best practices, style guides and testing coverage as you see fit. I would be very grateful for any feedback!

I am also currently seeing the following warning and due to my unfamiliarity with C# I am unsure how to solve it optimally: 

> csharp/src/KubernetesClient/Extensions.cs(8,25): warning CA1724: The type name Extensions conflicts in whole or in part with the namespace name 'Microsoft.Extensions'. Change either name to eliminate the conflict.

**Thank you for taking the time to review and consider this PR!** 
My team is very much looking forward to seeing this feature downstream in Bridge-To-Kubernetes as soon as possible, as this will significantly reduce our development efforts on multiple projects.

### DIFF
--- a/kubernetes-client.ruleset
+++ b/kubernetes-client.ruleset
@@ -301,6 +301,6 @@
     <Rule Id="CS1573" Action="None" />
     <Rule Id="CS1574" Action="None" />
     <!-- Rename k8s.Extensions type to mitigate conflict with Microsoft.Extensions namespace https://github.com/kubernetes-client/csharp/pull/544#issuecomment-759230655-->
-    <Rule Id="CA1723" Action="None" />
+    <Rule Id="CA1724" Action="None" />
   </Rules>
 </RuleSet>
--- a/kubernetes-client.ruleset
+++ b/kubernetes-client.ruleset
@@ -300,6 +300,7 @@
     <Rule Id="CS1591" Action="None" />
     <Rule Id="CS1573" Action="None" />
     <Rule Id="CS1574" Action="None" />
-
+    <!-- Rename k8s.Extensions type to mitigate conflict with Microsoft.Extensions namespace https://github.com/kubernetes-client/csharp/pull/544#issuecomment-759230655-->
+    <Rule Id="CA1723" Action="None" />
   </Rules>
 </RuleSet>
--- a/src/KubernetesClient/Authentication/OidcTokenProvider.cs
+++ b/src/KubernetesClient/Authentication/OidcTokenProvider.cs
@@ -32,7 +32,7 @@ namespace k8s.Authentication
 
         public async Task<AuthenticationHeaderValue> GetAuthenticationHeaderAsync(CancellationToken cancellationToken)
         {
-            if (_expiry == null || DateTime.UtcNow.AddSeconds(30) > _expiry)
+            if (_expiry == null || _accessToken == null || DateTime.UtcNow.AddSeconds(30) > _expiry)
             {
                 await RefreshToken().ConfigureAwait(false);
             }
--- a/src/KubernetesClient/Authentication/OidcTokenProvider.cs
+++ b/src/KubernetesClient/Authentication/OidcTokenProvider.cs
@@ -58,7 +58,7 @@ namespace k8s.Authentication
             }
             catch (Exception ex)
             {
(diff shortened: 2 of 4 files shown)


arbielsk:
CNCF ICLA accepted.

arbielsk:
There seem to be non-deterministic tests in LeaderElectionTests.cs

**Example**
Run: https://github.com/kubernetes-client/csharp/actions/runs/468548275 

Tests finish successfully for Ubuntu, but fail for MacOS with same code. On my local machine (MacOS) the tests run successfully (3/3 runs, no test failures).

Also without any code changes, the result for tests on the Ubuntu build differ between these two runs (I added an empty commit to re-trigger Github Actions):
https://github.com/kubernetes-client/csharp/actions/runs/468492411
https://github.com/kubernetes-client/csharp/actions/runs/468548275

I don't believe this is in the scope of this PR as it should be independent of the LeaderElection feature if I am interpreting the code correctly.

tg123:
seems `Microsoft.Extensions` was introduce by new dependency

arbielsk:
> seems `Microsoft.Extensions` was introduce by new dependency

Yes it definitely was. However I am unsure what the correct response would be:
- Rename Extensions in K8S Client?
- Try to find a OIDC client that doesn't rely on Microsoft.Extensions (and is possibly lower quality)?
- Something I am not aware of?

brendandburns:
Can you just use explicit full package names (e.g. `k8s.Extensions`) to eliminate the abiguity?

tg123:
you may want rebase master to get rid of flaky testcases

i am fine with allow `CA1723` in our repo, see `kubernetes-client.ruleset` in our repo.

arbielsk:
> Can you just use explicit full package names (e.g. `k8s.Extensions`) to eliminate the abiguity?

@brendandburns I've spent a little more time educating myself on the issue. As far as I understand it the warning is produced because the **type name** `Extensions` in namespace `k8s` that has a partial conflict with the **namespace** `Microsoft.Extensions` which is introduced by the OidcClient dependency. I therefore believe (with my limited knowledge of C#) that your suggestion is not applicable in this context as this is not about two type names being in conflict.

I therefore understand that there are the following possible resolutions:
- accept the warning as suggested by @tg123
- rename the `Extensions` type to something different that seems appropriate. Maybe `MetadataExtensions` as the code seems to be related to that? I've tried renaming the class and file and the tests are still passing, including the VersionConverter Tests that are relying on this functionality. 

@tg123 @brendanburns Please let me know what your preferred way of solving this would be. I have a commit prepared to rename the type name to `MetadataExtensions` but have not pushed it, as I believe the strategy for handling this warning should be your call.

Thank you again for taking the time to review my PR!

brendandburns:
Accepting the warning for now is fine with me.

If we want to make the change, we need to make it across a major version since it is a breaking change.

tg123:
please add CA1723  to kubernetes-client.ruleset to disable warning 
then LGTM

arbielsk:
Thanks again, @brendanburns @tg123 for getting back to me so quickly. I've added **CA1724** to the exceptions and added a comment explaining the situation.
Looking forward to seeing this merged :) 

Thanks again and have a great night!

tg123 (approved):
/LGTM

arbielsk:
Thank you, @tg123, @brendandburns!

PR #548, opened 2021-01-20
Add HttpClient property to the IKubernetes interface

qmfrederik:
The `HttpClient` property is a public property of the `Kubernetes` class, but not present on the `IKubernetes` interface.

In my code, I prefer to take a dependency on the `IKubernetes` interface instead of the `Kubernetes` class, because the methods on the `Kubernetes` class are not virtual and do not allow me to mock the Kubernetes API in unit tests.

When writing extension methods for the Kubernetes API, I need to access the `HttpClient` property so that I can send send HTTP requests to the API server. This PR adds that property to the interface, in addition to it being available on the class.


brendandburns:
Can you change the name to `IKubernetes.HttpClient` or somesuch?

qmfrederik:
@brendanburns I guess you're referring to the file name, right? If so - done!

PR #549, opened 2021-01-20
Make LineSeparatedHttpContent public

qmfrederik:
If you call `ListNamespaced*WithHttpMessagesAsync(..., watch:true)`, you'll get a `HttpOperationResponse<*List>` object in return.

However, it's currently not easy to access the raw HTTP watch data, because `response.Content.Content` is of type `LineSeparatedHttpContent` which is internal.

This PR makes `LineSeparatedHttpContent` public and have it expose a generic `TextReader` which allows you to read the watch data.

This allows the community to provide custom implementations of the Watch algorithm.

### DIFF
--- a/src/KubernetesClient/WatcherDelegatingHandler.cs
+++ b/src/KubernetesClient/WatcherDelegatingHandler.cs
@@ -142,7 +142,7 @@ namespace k8s
             }
         }
 
-        internal class LineSeparatedHttpContent : HttpContent
+        public class LineSeparatedHttpContent : HttpContent
         {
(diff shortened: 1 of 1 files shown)


qmfrederik:
This repo contains a couple of issues related to the behavior of the Watch method not matching the expectations. Making this property public allows the community to experiment with alternative implementations, while the Watch API in this repository is being ironed out.

tg123:
/LGTM

btw  my idea to watch related issues is to hack into generator side to make watch return something easier to handle.
no handler anymore maybe

PR #550, opened 2021-01-20
Support wildcard IPv4 and IPv6 addresses

qmfrederik:
Some tools can generate kubeconfig files which use wildcard IPv4 or IPv6 addresses. For example, using k3d with `--api-server=https://0.0.0.0:6433/` would generate a kubeconfig file like this:

```
apiVersion: v1
clusters:
- cluster:
    certificate-authority-data: (...)
    server: https://0.0.0.0:6433
  name: k3d-k3s-default
```

Standard Kubernetes tools (like kubectl or Helm) correctly parse the 0.0.0.0 IP address and transform it 127.0.0.1; 3rd party tools like curl or wget will do the same on Unix systems.

This is default behavior on Unix but not on Windows. As a result, the .NET Kubernetes client will fail to work with kubeconfig files like this and you'll get HTTP exceptions.

Go has explicit workarounds for this (see https://github.com/golang/go/commit/1a0b1cca4c26d41fe7508ffdb355de78b4ea2a19), and this PR attemps to replicate these workarounds in the .NET client.

### DIFF
--- a/src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
+++ b/src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
@@ -10,6 +10,7 @@ using System.Threading.Tasks;
 using k8s.Authentication;
 using k8s.Exceptions;
 using k8s.KubeConfigModels;
+using System.Net;
 
 namespace k8s
 {
@@ -271,6 +272,22 @@ namespace k8s
             if (!Uri.TryCreate(Host, UriKind.Absolute, out var uri))
             {
                 throw new KubeConfigException($"Bad server host URL `{Host}` (cannot be parsed)");
+            }
+
+            if (IPAddress.TryParse(uri.Host, out var ipAddress))
+            {
+                if (IPAddress.Equals(IPAddress.Any, ipAddress))
+                {
+                    var builder = new UriBuilder(Host);
+                    builder.Host = $"{IPAddress.Loopback}";
+                    Host = builder.ToString();
+                }
+                else if (IPAddress.Equals(IPAddress.IPv6Any, ipAddress))
+                {
+                    var builder = new UriBuilder(Host);
+                    builder.Host = $"{IPAddress.IPv6Loopback}";
+                    Host = builder.ToString();
+                }
             }
 
             if (uri.Scheme == "https")
--- a/tests/KubernetesClient.Tests/KubernetesClientConfigurationTests.cs
+++ b/tests/KubernetesClient.Tests/KubernetesClientConfigurationTests.cs
@@ -381,6 +381,33 @@ namespace k8s.Tests
             var cfg = await KubernetesClientConfiguration.BuildConfigFromConfigFileAsync(new FileInfo(path)).ConfigureAwait(false);
         }
 
+        [Fact]
+        public void ContextWithWildcardIPv4()
+        {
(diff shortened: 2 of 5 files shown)


qmfrederik:
This is the exception which you get with the current implementation:

```
    System.Net.Http.HttpRequestException : IPv4 address 0.0.0.0 and IPv6 address ::0 are unspecified addresses that cannot be used as a target address. (Parameter 'hostName') (0.0.0.0:6433)
    ---- System.ArgumentException : IPv4 address 0.0.0.0 and IPv6 address ::0 are unspecified addresses that cannot be used as a target address. (Parameter 'hostName')
```

src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
```diff
@@ -273,6 +273,19 @@ private void SetClusterDetails(K8SConfiguration k8SConfig, Context activeContext
                 throw new KubeConfigException($"Bad server host URL `{Host}` (cannot be parsed)");
             }
 
+            if (uri.Host == "0.0.0.0")
```
tg123, line 276: IPAddress.Any and IPAddress.IPv6Any

src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
```diff
@@ -273,6 +273,19 @@ private void SetClusterDetails(K8SConfiguration k8SConfig, Context activeContext
...
+            {
+                var builder = new UriBuilder(Host);
+                builder.Host = "127.0.0.1";
+                Host = builder.ToString();
+            }
+            else if (uri.Host == "[::]" || uri.Host == "[0:0:0:0:0:0:0:0]")
+            {
+                var builder = new UriBuilder(Host);
```
tg123, line 284: IPAddress.Loopback

tg123 (commented):
suggest use ip const

brendandburns:
I don't like this. This is a bug in the kubeconfig generator imho 0.0.0.0 != 127.0.0.1 and I don't think we should work around a bug in k3s

qmfrederik:
It's a weird corner case, but I don't think it is a bug per se. On Unix, 0.0.0.0 _always_ resolves to 127.0.0.1, kind of "by convention". It works with standard Unix utilities like cURL, wget.

The behavior was important enough so that golang decided to replicate the same behavior on Windows (golang/go@1a0b1cc).

It's very weird to end up with a kubeconf file which works fine on Linux, works fine on Windows with the k8s tools, but have the .NET client barks at it.

I think it's reasonable to expect the .NET k8s client accept kubeconf files which are accepted by the rest of the ecosystem.

brendandburns:
meh, I still don't like it, but I agree with the principle that if it works with `kubectl` it should work with this library, so ok.

Please fix the constant comments from @tg123

qmfrederik:
Thanks, @brendandburns . I updated the code so that the IP address is parsed and then compared with the `IPAddress.Any` or  `IPAddress.IPv6Any` values, hope that's OK.

qmfrederik:
Build failure is #556

tg123:
/LGTM

i used to `curl 0:xxx`, never know it was converted by client.

brendandburns:
@qmfrederik the fix to #556 has merged, can you rebase and then we can merge this?

Issue #533, opened 2020-12-06
Watch Method stops watching


pbalexlear:
When using a watch method it appears to stop watching after a certain time period, not sure if this is a task deadlock or a stream timeout but after around 30 minutes new events are no longer triggered even if changes are made on the cluster, also the watch object says its still watching but when debugging and looking at the tasks they are awaiting the streamreader.readline

pbalexlear:
Having spent a lot of time looking into this i believe its due to a bug in the Ssl stream, potentially a deadlock but essentially the await calls on the read stream method just hand indefinitely and never get called back.

tg123:
did you server close the connection?

pbalexlear:
Not that i can see, either way were that the case i would expect and exception to be thrown which doesnt happen.

brendandburns:
It's quite possible also that something in the network path timed out. You need to send keep-alives in many cases to keep the TCP stream open. (or you need to timeout the watch at a lower number like 5 minutes)

pbalexlear:
Maybe, but im using the default implementation of this client and by default it isnt working, it doesn't throw an exception just seems to get stuck in an await cycle.
i did try changing the HttpClient inside this library to have an infinite timeout but this didnt seem to change this behaviour.

eskaufel:
@pbalexlear I think I have the same issue. Just out of curiosity, are you using Windows or Linux?

pbalexlear:
@eskaufel 

I was doing this on mac initially but as the debugger for tasks is better in VS for windows i tried on windows with the same results but was able to see more in detail what was happening. Although I've since deleted the windows VM so no longer have the information, given this happens to me every-time after maybe 5mins of no events i would have thought it would be fairly easy to reproduce. 

It does look to be like its potentially a bug in SSL streams as when you do a kube proxy and watch locally using the http address this behaviour doesn't happen.

pbalexlear:
Ideally there should be some exception thrown and then i could restart the watcher but as there isnt i cannot capture the event and as the indicator 'Watching' is set in the same method thats stuck awaiting a task result this never gets changed either so i cannot capture if a watcher truly is watching or not.

tg123:
could you please create a test case in e2e project?

and i will try to understand and fix it

pbalexlear:
@tg123 Could you explain a bit more by what you mean by create a test case in e2e project? I'm happy to help assist in any way i can although i think the best way for you to understand this issue is to recreate it.

[...]

Issue #555, opened 2021-01-26
Supporting SocketsHandler for KubeClient


tg123:
keep track of msft-internal question on `Supporting SocketsHandler for KubeClient`

PR #560, opened 2021-02-12
add tcp keep alive hack to net5 target

tg123:
this is the temp fix for #533 and net5 only
net5 users will have tcp keepalive by default

### DIFF
--- a/src/KubernetesClient/Kubernetes.ConfigInit.cs
+++ b/src/KubernetesClient/Kubernetes.ConfigInit.cs
@@ -184,7 +184,6 @@ namespace k8s
                 var ip = Dns.GetHostEntry(context.DnsEndPoint.Host).AddressList.First();
                 var ep = new IPEndPoint(ip, context.DnsEndPoint.Port);
 
-                // Console.WriteLine(ep);
                 await socket.ConnectAsync(ep, token).ConfigureAwait(false);
 
 
--- a/src/KubernetesClient/Kubernetes.ConfigInit.cs
+++ b/src/KubernetesClient/Kubernetes.ConfigInit.cs
@@ -171,6 +171,18 @@ namespace k8s
 
 
 #if NET5_0
+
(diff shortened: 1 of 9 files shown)


tg123:
#555

src/KubernetesClient/Kubernetes.ConfigInit.cs
```diff
@@ -166,6 +168,27 @@ private void AppendDelegatingHandler<T>()
...
+                    NoDelay = true,
+                };
+
+                socket.SetSocketOption(SocketOptionLevel.Socket, SocketOptionName.KeepAlive, true);
+
+                await socket.ConnectAsync(context.DnsEndPoint.Host, context.DnsEndPoint.Port, token).ConfigureAwait(false);
+                return new NetworkStream(socket, ownsSocket: true);
+            };
+
+            var p = HttpClientHandler.GetType().GetField("_underlyingHandler", BindingFlags.NonPublic | BindingFlags.Instance);
```
brendandburns, line 188: Add a comment here to explain where this magic field name came from

brendandburns:
Looks good, one small comment on documenting things.

(separately, looks like we should exclude tests from the codeql scanning)

brendandburns:
We should perhaps only add keep-alives for sockets used for watch.

brendandburns:
I think we're hitting this:

https://github.com/dotnet/runtime/issues/24917

brendandburns:
Perhaps we should switch to HTTP/2 pings?

https://github.com/dotnet/runtime/pull/40257

tg123:
I tested this when I first impl the tcp keepalive
at first I think it is `tcp keepalive` but I am still being kicked out after enable this.

> Perhaps we should switch to HTTP/2 pings?
> 
> [dotnet/runtime#40257](https://github.com/dotnet/runtime/pull/40257)

PR #567, opened 2021-02-22
fix(yaml parser): Add IgnoreUnmatchedProperties to yaml deserializer

### LINKED ISSUE #566: Exception when kube config contains additional properties
Hey there :)

Someone just found out in relation to my operator sdk (https://github.com/buehler/dotnet-operator-sdk/issues/149) that the construction of the kube config fails if it contains additional properties.

I looked into your code and discovered, that the `Yaml.LoadAllFromString` method does include the `.IgnoreUnmatchedProperties` call on the deserializer builder, but the `public static T LoadFromString<T>` does not.

To make this more resilient and possible for users with AWS kubernetes (which adds additional properies to the kube config) it would be nice to include the `IgnoreUnmatchedProperties` call to the load from string method as well.

Regards

buehler:
This fixes #566.

It only adds the `IgnoreUnmatchedProperties` switch to the yaml deserializer in `LoadAllFromString` and `LoadFromString` methods.


buehler:
Hmm.. @brendanburns the failed test seems to be shakey

src/KubernetesClient/Yaml.cs
```diff
@@ -158,10 +159,11 @@ public static T LoadFromString<T>(string content)
         {
             var deserializer =
                 new DeserializerBuilder()
-                    .WithNamingConvention(CamelCaseNamingConvention.Instance)
+                    .WithNamingConvention(new YamlRenameCamelCaseConvention())
```
tg123, line 162: better to have a static singleton

src/KubernetesClient/Yaml.cs
```diff
@@ -190,6 +192,27 @@ public static string SaveToString<T>(T value)
             return stringBuilder.ToString();
         }
 
+        private class YamlRenameCamelCaseConvention : INamingConvention
+        {
+            private readonly INamingConvention _camelCase = CamelCaseNamingConvention.Instance;
+
+            private readonly IDictionary<string, string> _rename = new Dictionary<string, string>
+            {
+                { "namespaceProperty", "namespace" },
```
tg123, line 201: any reason for this renaming?
seems weird if my config was renamed after loaded

buehler: Yes, sadly when using `IgnoreUnmatchedProperties`, the `namespace` property will not be matched anymore. If discovered this while testing. With this renaming strategy, the properties are matched (again) correctly.

src/KubernetesClient/Yaml.cs
buehler: 👌🏾

buehler: Using static singleton now

src/KubernetesClient/Yaml.cs
brendandburns: Do we understand why this happens? I'd rather not make changes w/o understanding the reason for it.

brendandburns:
Change looks reasonable, but I'm wondering about why the renaming is required.

buehler:
@brendandburns the generated classes for the Kubernetes client contain `[JsonProperty]` attributes with the corresponding (correct) name. But there are no corresponding attributes for the yaml serializer / deserializer. So by default, the yaml serializer searches for same name properties (namespace <-> Namespace).

Since `namespace` is a reserved word in C#, the generated names used for those reserved words are suffixed with `Property` (e.g. `NamespaceProperty`. Sadly, this is not found by the deserializer when using `IgnoreUnmatchedProperties`. The renaming strategy aims to use the normal CamelCaseStrategy in the first place and then map the irregular keyword names like `Namespace`, `object` and so on. This renaming is only relevant for yaml, because of the reason stated above (the attributes).

The "best" solution in my opinion would be to adjust the generation templates - the ones that generate the C# classes out of the Kubernetes definitions, but as I saw, those are T4 templates and that's way out of my league - to add an equivalent to the jsonproperty attribute for yaml.

I use the same strategy in my Kubernetes operator SDK (https://github.com/buehler/dotnet-operator-sdk) because the re-serialization of crds (for example) contains wrong names (i.e. `namespaceProperty` instead of `namespace`).

tg123:
i do think rename is a good idea
just checked yamldotnet, not sure if we can add `WithAttributeOverride` in order that make YAML honor `JsonProperty`

buehler:
@tg123 I also had a look into this overwrite method, it seems to be to overwrite a specific property of a type. So either the client generate would generate this, or we use this renaming strategy.

As said, the cleanest way would be to introduce the change into the client generator template (i.e. generating the yamlproperty attributes).

alexmg:
I have also run into the same issue as @buehler with additional properties in YAML files. I added the official [examples](https://github.com/kubernetes/examples) repository as a submodule to my project and ran all the YAML files through the deserializer and found a number that contained properties not defined in the models.

It appears that the models are generated with AutoRest and I am not sure how difficult it would be to have additional attributes added to the generated types, but given the existing `JsonProperty` attributes are already a reliable source of the truth regarding model property names, an alternative approach is to utilize those within the existing `Yaml` class by creating a set or mirrored `YamlMember` attributes dynamically.

I have opened PR #574 to explore this alternative approach. The PR contains the same unit tests with the addition of an extra roundtrip test for the `NamespaceProperty` property on `V1ObjectMeta` type as that is a prime example of the issue.

A bonus of utilizing the reliable code generated `JsonProperty` attributes for the YAML deserialization is that the `AutoRestTypeInspector` is no longer required because all known properties are mapped explicitly. This mapping is completely internal and not visible to consumers of the `Yaml` class.

buehler:
Actually, I've to say that the solution by @alexmg is definitely more elegant than mine!
I really like the solution and I'm closing this merge request in favor of his!

Also, thanks @alexmg for the idea, I'm going to use it in the operator sdk for serializing things. I've got the same problem there.

tg123:
A new model generator is on the way, still struggling some blocking issues
will for sure include yaml attrib

buehler:
@tg123 Very nice to hear!
If I may: do you think about reworking the generator to use `records` insted of `classes`? This may be an improvement in terms of immutability and cloning / equality checking of the generated resources.

PR #574, opened 2021-02-27
Support IgnoreUnmatchedProperties in YAML serialization

### LINKED ISSUE #566: Exception when kube config contains additional properties

alexmg:
This fixes #566 as an alternative to #567 by mapping `JsonPropertyAttribute` instances to `YamlMemberAttribute` instances dynamically.

- During buliding of the deserializer the `DeserializerBuilder` has a `WithAttributeOverride` call made for every `JsonPropertyAttribute` found on the generated models.
- The `AutoRestTypeInspector` has been removed as it is no longer required with the explicit attribute mappings.
- The deserializer is configured with `IgnoreUnmatchedProperties` to avoid exceptions when reading YAML files with additional properties.
- Serializer and deserializer instances are created statically and cached instead of being constructed and configured on every call.


src/KubernetesClient/Yaml.cs
```diff
@@ -105,38 +126,23 @@ public static List<object> LoadAllFromString(string content, Dictionary<string,
                 throw new ArgumentNullException(nameof(typeMap));
             }
 
-            var deserializer =
```
tg123, line 108: I think the reason of `new` here is that the deserializer might not be thread safe

tg123: seems I am wrong after we updated to 8.1.2
https://github.com/aaubry/YamlDotNet/blob/2e4e5bb402cafee22f36abd65423d60e60221b40/releases/8.1.0.md

should double confirm then

Issue #584, opened 2021-03-13
Improve Watcher Pattern


mpoettgen:
Initially I've had a bit of trouble understanding the Watcher pattern, so I've looked at the implementation and found a few things that may be worthwhile to be changed/added.

# Current situation

## The `Watcher` object is not 'cancellable' and/or 'awaitable'

If I understand things correctly the consumer is supposed to keep the `Watcher` around for as long as he wants to watch a resource and then dispose it. In an async context it would be nicer to be able to signal termination using a `CancellationToken` and have an awaitable `Task` somewhere (like the [ExecuteTask](https://github.com/dotnet/runtime/blob/6791d05494f5aa4d1a789525c2708e775aafdc07/src/libraries/Microsoft.Extensions.Hosting.Abstractions/src/BackgroundService.cs#L24) property) that completes when the `Watcher` is done watching. This would be similar to the pattern in the `BackgroundWorker`.

## Putting the `WatcherLoop` onto the thread pool may not always be necessary/desired

If your `Watcher` is already running on a thread pool tread, then all that this thread will be doing is wait for the watcher loop to finish in another thread. It can do that asynchronously, but anyways...

## API usage is not really obvious

As I said, it took me a while to understand how it works.

# Proposal

1. Add `StartWatch` extension methods, similar to the `Watch` extension methods, but getting an extra `CancellationToken` parameter and returning a `Task` object.
2. Add private constructors to the `Watcher` class that take the extra `CancellationToken`, but do not run the `WatcherLoop`. The `CancellationToken` can be registered with the `CancellationTokenSource`.
3. Add static `StartWatch` methods to the `Watcher` class that construct `Watcher` objects and then execute the `WatcherLoop` returning the `Task`. The `Watcher` objects can be disposed automatically, when the `WatcherLoop` is done.

If this is something that would be considered, I would be willing to create a pull request.

tg123:
PRs are welcomed for sure

Please take a look here 
https://github.com/lodejard/WatchExample/blob/9d6934d665f1c53162a6e5e4c66874ba66374320/Worker.cs

I am trying to add awaitable and enumerable APIs to watcher object.
still testing the changes and some glitches need to be fixed

#581 

cc @lodejard

davidfowl:
IAsyncEnumerable is the answer to this as it is model 2 but with compiler support (await foreach)

mpoettgen:
[...]

Issue #581, opened 2021-03-12
enumerable for watcher


tg123:
caller-driven style api for watcher

proposal:
https://github.com/lodejard/WatchExample/blob/9d6934d665f1c53162a6e5e4c66874ba66374320/Worker.cs#L147

PR #586, opened 2021-03-14
expose IAsyncEnumerable API for watcher

tg123:
#584
#581 

require C# 8.0 to use this API
`netstandard2.0` is supported with `Microsoft.Bcl.AsyncInterfaces`

_Note_: here `onError` to `CreateWatchEventEnumerator` is to handle non-fatal errors:
 * server yields a bad line, cannot be deserialized
 * server yields a `error` event

the enumerator will remain alive when `errros` above happens


src/KubernetesClient/WatcherExt.cs
```diff
@@ -59,5 +65,12 @@ public static class WatcherExt
         {
             return Watch(Task.FromResult(response), onEvent, onError, onClosed);
         }
+
+        public static IAsyncEnumerable<(WatchEventType, T)> WatchAsync<T, L>(
```
tg123, line 69: doc will be added after API style get reviewed

tg123:
cc @davidfowl @lodejard

src/KubernetesClient/Watcher.cs
```diff
@@ -174,6 +150,70 @@ private async Task WatcherLoop(CancellationToken cancellationToken)
...
+                if (!task.IsCompleted)
+                {
+                    // here to pass cancellationToken into task
+                    return task.ContinueWith(t => t.GetAwaiter().GetResult(), cancellationToken);
+                }
+
+                return task;
+            }
+
+            using var streamReader = await AttachCancellationToken(streamReaderCreator()).ConfigureAwait(false);
```
davidfowl, line 169: This looks a bit convoluted. It's because StreamReader doesn't support cancellation token right?

src/KubernetesClient/Watcher.cs
```diff
@@ -174,6 +150,70 @@ private async Task WatcherLoop(CancellationToken cancellationToken)
             }
         }
 
+        internal static async IAsyncEnumerable<(WatchEventType, T)> CreateWatchEventEnumerator(
+            Func<Task<TextReader>> streamReaderCreator,
+            Action<Exception> onError = null,
+            [EnumeratorCancellation] CancellationToken cancellationToken = default)
+        {
+            Task<TR> AttachCancellationToken<TR>(Task<TR> task)
```
davidfowl, line 158: What's the goal of this method?

src/KubernetesClient/Watcher.cs
```diff
@@ -174,6 +150,70 @@ private async Task WatcherLoop(CancellationToken cancellationToken)
...
+            }
+
+            using var streamReader = await AttachCancellationToken(streamReaderCreator()).ConfigureAwait(false);
+
+            for (; ; )
+            {
+                // ReadLineAsync will return null when we've reached the end of the stream.
+                var line = await AttachCancellationToken(streamReader.ReadLineAsync()).ConfigureAwait(false);
+
+                if (cancellationToken.IsCancellationRequested)
```
davidfowl, line 176: Typically better to use `cancellationToken.ThrowIfCancellationRequested()`

src/KubernetesClient/Watcher.cs
tg123: disposing watcher will dispose the stream which would cancel the `readlineasync` in the original watcher pattern
both `streamReaderCreator()` and `readlineasync` does not honor the cancellation token, here is a workaround to attach it

src/KubernetesClient/Watcher.cs
tg123: agree 
but here is to be compatible with old watcher pattern
otherwise, signal a cancel will raise an exception.

brendandburns:
I'm going to defer to @davidfowl and @lodejard to do the first reviews on this. Once they're happy I'll do a final review.

src/KubernetesClient/Watcher.cs
lodejard: Interesting, haven't seen this ContinueWith/CancellationToken pattern before. I'll look into that a little closer before commenting.

lodejard: Looking into this it seems the underlying task remains active even though the wrapping task becomes cancelled. I think you'd be okay without this pattern.

In a perfect world ReadLineAsync would take a CT... Short of that, to cover the `ReadLineAsync` it might be better to register the CT as a call to stream.Dispose. That will have the same effect as before - ensuring the underlying call waiting on network unwinds immediately.

```csharp
using var streamReader = await AttachCancellationToken(streamReaderCreator()).ConfigureAwait(false);
using var cancellationTokenRegistration = cancellationToken.Register(streamReader.Dispose);
```

For the first call down to content `ReadAsStreamAsync` --- that one doesn't take a CT either. Good news is it doesn't seem like that code path enters a wait state? In the original code it seems like that wasn't covered by cancellation either. So should be okay to leave the `streamReaderCreator()` unguarded by CT.

src/KubernetesClient/Watcher.cs
lodejard: For callers using `await foreach(...WatchAsync(CT))` it could be expected when a cancel is requested the await will throw? Might not be a mandatory convention, but it would indicate if the loop ended normally from response ending vs ending from a client-side cancellation teardown.

Ah - also the await keyword in the loop will also throw the exception if the ReadLineAsync task enters a cancelled state, so you'll need to catch for that as well for old watcher pattern.

How about if the `WatcherLoop` method has an additional `catch (OperationCancelledException)` case which does not call OnError? 

That would stop cancel exceptions from reaching the old pattern, but enable you to use `ThrowIfCancellationRequested` from the async enumerable function and allow `await foreach` callers to have task.Canceled status propagate back.

src/KubernetesClient/Watcher.cs
tg123: yup 
`streamReaderCreator` is under our control and easy to make it take CT

I am thinking that `AttachCancellationToken` is a waste if I use it only once :)

tg123:
any idea about `Action<Exception> onError` parameter, I do not like the design

mpoettgen:
@tg123,

> any idea about `Action<Exception> onError` parameter, I do not like the design

How about making `WatchEvent` a standalone, generic class `WatchEvent<T>` instead of a nested type and adding a third property `Error` of type `Exception`. `WatchAsync` should then return an `IEnumerable<WatchEvent<T>>` and you can either `yield return` the de-serialized watch event or a `WatchEvent` with `WatchEventType.Error` and the corresponding `Exception` in the `Error` property.

The other option would be to make the returned tuple a triple, but to me that would be a bit awkward.

tg123:
ping @davidfowl @lodejard

davidfowl:
Will take another look

tg123:
netstandard2.0 support dropped

davidfowl:
> netstandard2.0 support dropped

That's a big breaking change, not sure you should do that.

tg123:
> > netstandard2.0 support dropped
> 
> That's a big breaking change, not sure you should do that.

netstarnd2 dropped here #603

src/KubernetesClient/Watcher.cs
```diff
@@ -174,6 +155,67 @@ private async Task WatcherLoop(CancellationToken cancellationToken)
...
+                    yield break;
+                }
+
+                WatchEvent @event = null;
+
+                try
+                {
+                    var genericEvent = SafeJsonConvert.DeserializeObject<Watcher<KubernetesObject>.WatchEvent>(line);
+
+                    if (genericEvent.Object.Kind == "Status")
```
mpoettgen, line 194: @tg123, not sure, but if I'm not mistaken, if the list/watch operation actually returns existing objects, then those will be read (again) by the first ReadLineAsync above. The `line` will not deserialize properly, as the line represents the object list, not a watch event. Therefore a `NullReferenceException` will be thrown, caught below and reported via the `onError` handler if and only if an `onError` handler has been provided. Test cases do not cover this. I don't think this is intentional!?

The `LineSeparatedHttpContent` only peeks at the first line, so it is read again here. It doesn't know, whether it should discard the line as it doesn't know, whether it is reading an object list or a watch event.

tg123: i noticed the npe during dev and did some changes to testcase https://github.com/kubernetes-client/csharp/pull/586/files/c8506d22c52caf3e827a51a9f4de477659557759#diff-b8e0f043f040cdeca7bd66470c6f4cf8dc42b39a1f514cf412e284309284e431L196

Yet, server will never return a `listobj` if `watch=true` and `LineSeparatedHttpContent` will install to `watch=true` only.

In the old days, (1.6 or 1.7 i believe), there is a `/watch/xxx` api which return `listobj` which will cause the issue you mentioned. but it was deprecated and `get` with `watch=true` will yield `watch obj` only
that is the reason i changed the testcase removing the first list obj.

mpoettgen: Then could you please have a look at the TestWatchWithHandlers? That's the test I stumbled on.

tg123: yup
`await WriteStreamLine(httpContext, MockKubeApiServer.MockPodResponse).ConfigureAwait(false);`
should be removed 
it is outdated, but i did not notice it as test passed

mpoettgen: That is because there is no `onError` handler attached. I'm experimenting with an alternative implementation that encapsulates both success and error results in a `WatchEvent<T>`. That's why I noticed.

brendandburns:
Where is this PR at? I was waiting for the discussion to settle. Is it ready for a final review?

brendandburns:
Checking in on the status of this PR again :)

Woazboat:
Also interested in the status of this PR (since the current watcher api is extremely cumbersome to use)

davidfowl (approved):

tg123:
talked with @davidfowl and got approval from him
cc @brendanburns

tg123:
ping @brendanburns

davidfowl:
image

The gray approval 😄

PR #599, opened 2021-04-01
Default type map for YAML deserialization

### LINKED ISSUE #598: Cohesion issue: Yaml.LoadAllFrom* requires an explicit type map, even for built-in types
As shown in [`YamlTests.LoadAllFromString`](https://github.com/kubernetes-client/csharp/blob/59352954c4d6bee95126e85aaa9a3d7ed8097242/tests/KubernetesClient.Tests/YamlTests.cs#L26-L28), it's a bit cumbersome that I need to supply an explicit type mapping dictionary for all of the generated types in `k8s.Model`.

Why should my code need to know how to resolve this type mapping correctly? Using this in my code requires me to either maintain a static type mapping, or to code reflective logic that inspects the types in the `KubernetesClient` assembly to find model types and populate my dictionary at runtime. Curating this mapping for those default model types feels like it should be the job of the library itself.

Fortunately I've got a PR ready to go to fix this, so I'll go ahead and submit that in just a moment.

benjamincburns:
* Fixes #598
* Affects Yaml.LoadAllFrom* methods
* Doesn't require user to explicitly pass a mapping for known types
* Allows user to specify a mapping for custom types, if required

### DIFF
--- a/src/KubernetesClient/Yaml.cs
+++ b/src/KubernetesClient/Yaml.cs
@@ -39,6 +39,19 @@ namespace k8s
                 .WithOverridesFromJsonPropertyAttributes()
                 .BuildValueSerializer();
 
+        private static readonly IDictionary<string, Type> ModelTypeMap = typeof(KubernetesEntityAttribute).Assembly
+            .GetTypes()
+            .Where(t => t.GetCustomAttributes(typeof(KubernetesEntityAttribute), true).Any())
+            .ToDictionary(
+                t =>
+                {
+                    var attr = (KubernetesEntityAttribute)t.GetCustomAttribute(
+                        typeof(KubernetesEntityAttribute), true);
+                    var groupPrefix = string.IsNullOrEmpty(attr.Group) ? "" : $"{attr.Group}/";
+                    return $"{groupPrefix}{attr.ApiVersion}/{attr.Kind}";
+                },
+                t => t);
+
         public class ByteArrayStringYamlConverter : IYamlTypeConverter
         {
             public bool Accepts(Type type)
@@ -84,10 +97,11 @@ namespace k8s
         /// The stream to load the objects from.
         /// </param>
         /// <param name="typeMap">
-        /// A map from apiVersion/kind to Type. For example "v1/Pod" -> typeof(V1Pod)
+        /// A map from apiVersion/kind to Type. For example "v1/Pod" -> typeof(V1Pod). If null, a default mapping will
+        /// be used.
         /// </param>
         /// <returns>collection of objects</returns>
-        public static async Task<List<object>> LoadAllFromStreamAsync(Stream stream, Dictionary<string, Type> typeMap)
+        public static async Task<List<object>> LoadAllFromStreamAsync(Stream stream, IDictionary<string, Type> typeMap = null)
         {
             var reader = new StreamReader(stream);
             var content = await reader.ReadToEndAsync().ConfigureAwait(false);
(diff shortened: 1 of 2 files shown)


benjamincburns:
Fixing up the formatting issues, will `push -f` in just a moment

benjamincburns:
The failing test (`LeaderElectionTest`) passes fine when I run it locally, and [it seems that it's had some issues with flakiness in the past](https://github.com/kubernetes-client/csharp/blob/59352954c4d6bee95126e85aaa9a3d7ed8097242/tests/KubernetesClient.Tests/LeaderElection/LeaderElectionTests.cs#L258-L271). I'll try and regenerate my commit hash to force a rerun, otherwise I think one of the project admins will need to do that for me.

benjamincburns:
Failed in `LeaderElectionTests` again, but this time at an entirely different line. I await advice from the maintainers, as I'd rather not keep resubmitting until it passes.

benjamincburns:
Oh, I missed the bot commands... I'll give it one more shot...

/retest

tg123:
i will help you with the flaky tests

brendandburns:
@benjamincburns Does this actually work? The reason that I ask is that the type metadata fields are no filled in the objects as far as I know. They're only there when you deserialize them from the server (or a YAML file on disk)

So I'm not certain that this will build the mapping you expect.

On the other hand you appear to have added tests to validate the behavior, so I'm not sure what is going on there.

Ultimately, the right way to do this is to add API type discovery logic similar to what is in the Java client here:
https://github.com/kubernetes-client/java/blob/master/util/src/main/java/io/kubernetes/client/Discovery.java

I'm ok to merge this, but I'm not positive it will work in practice.

benjamincburns:
> the type metadata fields are not filled in the objects as far as I know

I'm not sure I understand your meaning there. The annotations are in the generated `ModelExtensions.cs` file. In C# annotations are included at runtime by default, so they can be inspected to build the mapping just fine. The other source of your confusion may be that all of the model types appear to be defined as partial classes, so the classes in `ModelExtensions.cs` are actually partial definitions of those types, and they're merged in with the other definitions at compile time.

brendandburns:
@benjamincburns The problem is that for an object that you create (e.g. `new V1Pod()`) the fields that those extension methods refer to are empty strings.

e.g. the extension code here: https://github.com/kubernetes-client/csharp/blob/master/src/KubernetesClient/ModelExtensions.cs#L38

pulls out the ApiGroup in an object, but unless that object has been instantiated from a JSON blob from the server, or a JSON/YAML that is on file, the apiGroup field will be empty string. So you won't build the mappings that you expect.

Unless I'm missing something. The only place the api groups and versions and kinds are defined is on the Kubernetes server.

benjamincburns:
I'm not sure I follow entirely. In the case of `V1Pod`, my code is using the annotation shown here to add the type to the mapping.

https://github.com/kubernetes-client/csharp/blob/edaeb1443a394dcfc5981199402c7578e84b907c/src/KubernetesClient/generated/ModelExtensions.cs#L576-L582

You'll note that the group is an empty string, but I was under the impression that this was intentional and correct for this type. I was under the impression that the group was only included in the version string in cases where a kind might be ambiguous for a given API version, such as the `Event` kind.

https://github.com/kubernetes-client/csharp/blob/edaeb1443a394dcfc5981199402c7578e84b907c/src/KubernetesClient/generated/ModelExtensions.cs#L480-L486

vs

https://github.com/kubernetes-client/csharp/blob/edaeb1443a394dcfc5981199402c7578e84b907c/src/KubernetesClient/generated/ModelExtensions.cs#L704-L710

benjamincburns:
> but unless that object has been instantiated from a JSON blob from the server, or a JSON/YAML that is on file, the apiGroup field will be empty string.

This statement is especially confusing to me. This PR is strictly pertaining to deserialization from one or more YAML documents into the generated model types provided by this library. I'm not touching anything to do with JSON here, and any default values for mandatory fields such as kind and API version will be written as part of the deserialization process as they already are in the existing code.

Can you suggest a test case that you think might not work were this PR to be accepted?

brendandburns:
Ah, thank you for that direct pointer into ModelExtensions, I should have looked deeper into that file.

That provides the info that I was looking for, apologies for the confusion.

/lgtm
/approve

benjamincburns:
No worries, @brendanburns!

PR #604, opened 2021-04-14
Update LICENSE

AshenW0lf:
Update the copyright notice within the licence agreement, as per the instructions under "APPENDIX: How to apply the Apache License to your work."

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -186,7 +186,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright 2021 brendandburns and krabhishek8260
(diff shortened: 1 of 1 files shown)


AshenW0lf:
Contributor License Agreement (CLA), has now been signed

brendandburns:
Hrm, I would rather not have to update this every year and the authors are not just me.

Perhaps:

`{year} the Kubernetes Project`

AshenW0lf:
You don't need to update the year, is meant to be the year of first publication so could be a date in the past and the project name would work just as well as the "owners" of the copyright.

Issue #607, opened 2021-04-20
BuildDefaultConfig: env parsing for multiple KUBECONFIG paths doesn't account for double quotes (")


ChadJessup:
When calling ```KubernetesClientConfiguration.BuildDefaultConfig();``` if the ```KUBECONFIG``` env variable multiple paths, the split code doesn't trim the double quote. This makes some of the resulting ```FileInfo``` invalid:

```
KUBECONFIG=C:\Users\X\.kube\config;c:\Users\X\.kube\prod
```

```csharp
var configList = kubeconfig.Split(RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? ';' : ':')
    .Select((s) => new FileInfo(s));
```

The current code splits the path like this (note the quotes):
```
"C:\Users\X\.kube\config
c:\Users\X\.kube\prod"
```

The first path when provided to a ```FileInfo``` causes the path to be appended to the cwd, making an invalid path.
Trimming ```"``` before and after seems to solve the problem without any noticeable side-effects.

I'll be submitting a fix trimming the double quotes here in a second.

tg123:
seems this is because " in the env var

ChadJessup:
> seems this is because " in the env var

Kind of!
Windows seems to have special logic involved when a semi-colon is in the value.

Attached image is what a user sees when viewing the vars, only the ```KUBECONFIG``` kvp has quotes when viewed in debugger.
If I manually edit it and add a ; to the end, the quotes go away, but this value was written by k8s tooling awhile back.

![image](https://user-images.githubusercontent.com/618397/115439746-c625e900-a1c3-11eb-9035-71c03f9a3bfd.png)

tg123:
close as #608 
However, please also file a bug to the tool which put " in env

PR #608, opened 2021-04-20
Trimming double quote from parsed KUBECONFIG path.

ChadJessup:
Addresses #607 - trimming double quote when creating ```FileInfo```.


ChadJessup:
I signed it.

brendandburns:
I think it's kind of weird to have quotes in your environment variable value, but I don't see any harm in doing this.

/lgtm
/approve

Issue #618, opened 2021-05-11
Support for provideClusterInfo and installHint in KubeConfig Models under ExternalExecution.


namra98:
Hi Team,
Thanks for this great library.

Kubernetes has recently added support in kubeConfig for provideClusterInfo and installHint properties recently.
Link : https://kubernetes.io/docs/reference/access-authn-authz/_print/#configuration

These are not yet implemented in csharp library.
We were trying to deserialize this updated kubeconfig to C# kubernetes model, but this is failing with error as below:
**provideClusterInfoProperty not found on type k8s.KubeConfigModels.ExternalExecution.**

Current workaround is to remove this property manually from kubeconfig and then do deserialzation.

I can raise a PR for adding these properties in ExternalExecution Model.
Let me know if this sounds good.

brendandburns:
A PR to add support for this would be great.

Thanks!

namra98:
Fixed by https://github.com/kubernetes-client/csharp/pull/620

PR #620, opened 2021-05-14
Adding new properties in ExternalExecution model.

namra98:
**Overview**
Kubernetes added support for these two new parameters in ExternalExecution.
Which denotes whether to pass cluster data to newly created process as part of environment variables or not.
reference : https://kubernetes.io/docs/reference/access-authn-authz/authentication/#configuration

**Requirement**
Bug #618 

**Changes**
Added these two new fields as part of kubeconfig.

**ToDo**
Make changes to pass Cluster info to new process if ProvideClusterInfo is true.


brendandburns:
/lgtm
/approve

Thank you!

Issue #624, opened 2021-05-18
Getting "Unathorized" response when kubectl is working? 


amsoedal:
Hi, we have a customer who's been reporting that our CLI returns "unathorized" when they try to interact with their cluster, but when they use other kubernetes tools (kubectl, or the node kubernetes client) that everything works fine. Since our CLI relies on the c# kubernetes client, is there a possible issue related to this? They say they doesn't have access to the default namespace, but we're trying to list pods in a specified (non-default) namespace on their behalf.

tg123:
`var config = KubernetesClientConfiguration.BuildDefaultConfig();` to have same config loading pattern as kubectl

brendandburns:
It's always possible that we don't support some authentication method. Can you paste the details of the kubeconfig file (with secrets redacted, obviously)

jarrett-confrey:
@brendanburns i am the one whos seeing the issue, this is what my kubeeconfig looks like
```yaml
apiVersion: v1
clusters:
- cluster:
    certificate-authority-data: ...
    server: ...
  name: ...
contexts:
- context:
    cluster: ...
    namespace: ...
    user: ...
  name: ...
current-context: ...
kind: Config
preferences: {}
users:
- name: ...
  user:
    auth-provider:
      config:
        client-id: gangway
        client-secret: ...
        id-token: ...
        idp-issuer-url: https://dex...
        refresh-token: ...
      name: oidc
```

brendandburns:
The OIDC provider is here:

https://github.com/kubernetes-client/csharp/blob/master/src/KubernetesClient/Authentication/OidcTokenProvider.cs

and loaded here:

https://github.com/kubernetes-client/csharp/blob/2cbad3d42926d58729e9bb537de2200d746b9d90/src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs#L412

I don't see anything that jumps out as being wrong, but I suspect that there is some sort of problem with the intersection of the client and your OIDC provider.

brendandburns:
Note that OIDC support was only added in January 2021, so if you are using an old version of the client library that may be the cause.

amsoedal:
@brendandburns that definitely sounds possible. I'll need to double check.

amsoedal:
I checked with my team, and apparently the last time we updated was explicitly so that we could pull in the OIDC support (we're on version 4.0.16).

jarrett-confrey:
[...]

PR #633, opened 2021-05-20
fix: oidc

jarrett-confrey:
this pull request
- fixes an issues where the oicd expiration was not being properly checked causing the token to alway be refreshed
  - for me the issue causes an attempted and failed token refresh that prevents successful oicd authorization
    - some tokens are long lived and unable to be refreshed depending on the type of connector
      - https://github.com/dexidp/dex/#connectors
  - we now look at the expiration from the token and only attempt a refresh when needed similar to the javascript client
    - https://github.com/kubernetes-client/javascript/blob/master/src/oidc_auth.ts
  - we now use the id token instead of the access token for authentication
    - https://kubernetes.io/docs/reference/access-authn-authz/authentication/#openid-connect-tokens
- adds better error for failed token refreshes
  - all i was seeing was `Unauthorized` but the actual error i was facing turned out to be
    - `invalid_request: Refresh token is invalid or has already been claimed by another client.`

related issues
- https://github.com/microsoft/mindaro/issues/80#issuecomment-842406752
- https://github.com/microsoft/mindaro/issues/174
- https://github.com/kubernetes-client/csharp/issues/624
- https://github.com/kubernetes-client/csharp/pull/544

### DIFF
--- a/src/KubernetesClient/Authentication/OidcTokenProvider.cs
+++ b/src/KubernetesClient/Authentication/OidcTokenProvider.cs
@@ -40,8 +40,8 @@ namespace k8s.Authentication
             var handler = new JwtSecurityTokenHandler();
             var token = handler.ReadJwtToken(_accessToken ?? _idToken);
             var expiry = token.Payload.Exp ?? 0;
-            var dateTime = DateTimeOffset.FromUnixTimeSeconds(expiry);
-            return dateTime;
+            var dateTimeOffset = DateTimeOffset.FromUnixTimeSeconds(expiry);
+            return dateTimeOffset;
         }
 
         private OidcClient getClient(string clientId, string clientSecret, string idpIssuerUrl)
--- a/src/KubernetesClient/Authentication/OidcTokenProvider.cs
+++ b/src/KubernetesClient/Authentication/OidcTokenProvider.cs
@@ -6,7 +6,6 @@ using System.IdentityModel.Tokens.Jwt;
 using Microsoft.Rest;
 using IdentityModel.OidcClient;
 using k8s.Exceptions;
-using Serilog;
 
 namespace k8s.Authentication
 {
@@ -53,18 +52,6 @@ namespace k8s.Authentication
                 ClientSecret = clientSecret ?? "",
                 Authority = idpIssuerUrl,
             };
-
-            var shouldTrace = ServiceClientTracing.IsEnabled;
-            if (shouldTrace)
-            {
-                var serilog = new LoggerConfiguration()
-                    .MinimumLevel.Verbose()
-                    .Enrich.FromLogContext()
-                    .WriteTo.LiterateConsole(outputTemplate: "[{Timestamp:HH:mm:ss} {Level}] {SourceContext}{NewLine}{Message}{NewLine}{Exception}{NewLine}")
-                    .CreateLogger();
-
-                options.LoggerFactory.AddSerilog(serilog);
-            }
 
             return new OidcClient(options);
         }
--- a/src/KubernetesClient/Authentication/OidcTokenProvider.cs
+++ b/src/KubernetesClient/Authentication/OidcTokenProvider.cs
@@ -2,9 +2,11 @@ using System;
 using System.Net.Http.Headers;
 using System.Threading;
 using System.Threading.Tasks;
+using System.IdentityModel.Tokens.Jwt;
 using Microsoft.Rest;
 using IdentityModel.OidcClient;
 using k8s.Exceptions;
+using Serilog;
 
 namespace k8s.Authentication
 {
@@ -14,23 +16,33 @@ namespace k8s.Authentication
         private string _idToken;
         private string _refreshToken;
         private string _accessToken;
-        private DateTime _expiry;
+        private DateTimeOffset _expiry;
 
         public OidcTokenProvider(string clientId, string clientSecret, string idpIssuerUrl, string idToken, string refreshToken)
         {
             _idToken = idToken;
             _refreshToken = refreshToken;
             _oidcClient = getClient(clientId, clientSecret, idpIssuerUrl);
+            _expiry = getExpiryFromToken();
         }
 
         public async Task<AuthenticationHeaderValue> GetAuthenticationHeaderAsync(CancellationToken cancellationToken)
         {
-            if (_accessToken == null || DateTime.UtcNow.AddSeconds(30) > _expiry)
+            if ((_accessToken == null && _idToken == null) || DateTimeOffset.UtcNow.AddSeconds(30) > _expiry)
             {
                 await RefreshToken().ConfigureAwait(false);
             }
 
-            return new AuthenticationHeaderValue("Bearer", _accessToken);
+            return new AuthenticationHeaderValue("Bearer", _accessToken ?? _idToken);
+        }
+
+        private DateTimeOffset getExpiryFromToken()
+        {
+            var handler = new JwtSecurityTokenHandler();
+            var token = handler.ReadJwtToken(_accessToken ?? _idToken);
+            var expiry = token.Payload.Exp ?? 0;
+            var dateTime = DateTimeOffset.FromUnixTimeSeconds(expiry);
+            return dateTime;
         }
 
         private OidcClient getClient(string clientId, string clientSecret, string idpIssuerUrl)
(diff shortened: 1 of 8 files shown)


jarrett-confrey:
@arbielsk would you be able to verify that this doesnt break anything for you from https://github.com/kubernetes-client/csharp/pull/544?

brendandburns:
A couple of build steps failed, can you try a rebase?

I'm happy to also take this PR forward if you'd rather not or don't have time. Let me know.

Thanks!

jarrett-confrey:
@brendandburns im happy to take this forward myself and contribute to this repo with your review. i dont have any c# experience so please let me know if there are improvements you see that i could make. i have rebased by branch, let me know if that helps with the build issues. thank you!

src/KubernetesClient/KubernetesClient.csproj
```diff
@@ -36,10 +36,13 @@
     <PackageReference Include="Portable.BouncyCastle" Version="1.8.1.3" />
     <PackageReference Include="Microsoft.Rest.ClientRuntime" Version="2.3.10" />
     <PackageReference Include="prometheus-net" Version="4.1.1" />
+    <PackageReference Include="Serilog.Extensions.Logging" Version="2.0.2" />
```
tg123, line 39: can you use default trace instead of serilog
want to keep less dependencies for end users

src/KubernetesClient/Authentication/OidcTokenProvider.cs
```diff
@@ -14,23 +16,33 @@ public class OidcTokenProvider : ITokenProvider
         private string _idToken;
         private string _refreshToken;
         private string _accessToken;
-        private DateTime _expiry;
+        private DateTimeOffset _expiry;
```
tg123, line 19: any other concern why change this to datetimeoffset? just wondering. 

i do see
```
bumps IdentityModel.OidcClient
this causes the change from DateTime to DateTimeOffset
``

but did not understand why

File: src/KubernetesClient/Authentication/OidcTokenProvider.cs [hunk 3]
```diff
@@ -14,23 +16,33 @@ public class OidcTokenProvider : ITokenProvider
...
             _idToken = idToken;
             _refreshToken = refreshToken;
             _oidcClient = getClient(clientId, clientSecret, idpIssuerUrl);
+            _expiry = getExpiryFromToken();
         }
 
         public async Task<AuthenticationHeaderValue> GetAuthenticationHeaderAsync(CancellationToken cancellationToken)
         {
-            if (_accessToken == null || DateTime.UtcNow.AddSeconds(30) > _expiry)
+            if ((_accessToken == null && _idToken == null) || DateTimeOffset.UtcNow.AddSeconds(30) > _expiry)
```
tg123, line 31: any case that access token is null and idtoken is not null?
and i think we need to double check if idtoken has access to resources

src/KubernetesClient/KubernetesClient.csproj
jarrett-confrey: it doesnt seem like `OidcClient` is compatible with `ServiceClientTracing` so i have removed the verbose logging. if you see a way to make this work please let me know and ill add it back
https://github.com/IdentityModel/IdentityModel.OidcClient/blob/main/src/OidcClient/OidcClient.cs

src/KubernetesClient/Authentication/OidcTokenProvider.cs
jarrett-confrey: it looks like this was changed to resolve an issue with cross timezone information
- https://github.com/IdentityModel/IdentityModel.OidcClient/issues/273
- https://github.com/IdentityModel/IdentityModel.OidcClient/pull/274

src/KubernetesClient/Authentication/OidcTokenProvider.cs
jarrett-confrey: >any case that access token is null and idtoken is not null?

yes, that is the case where i was having an issue. my access token was null but i had an unexpired id token and the client would try to refresh when i didnt need to refresh. the unneeded refresh would fail due to our policies and no successful request could be made

>and i think we need to double check if idtoken has access to resources

im not sure i understand. how would we go about checking that? why would that check need to be added now when it wasnt needed before? is the javascript client doing this kind of checking?

jarrett-confrey:
>BTW anyway to test it? we can integrate it into ci

@tg123 the way im testing is just running `dotnet run --project examples/simple` and ensuring that i no longer get an `Unauthorized` error. i think you would need to have oidc auth setup in your kube config. one where the id token is not expired and then ensuring that unexpired id token is used to authenticate requests instead of attempting to refresh the token. my kube config looks like this https://github.com/kubernetes-client/csharp/issues/624#issuecomment-843651298. hopefully that helps somewhat, let me know if theres any more info i can provide

src/KubernetesClient/Authentication/OidcTokenProvider.cs
tg123: not 100% sure if the `id token` can be used as `bearer`
checked <https://openid.net/specs/openid-connect-core-1_0.html> and did not see if the `idtoken` can be used as the `bearer`

however, I also checked <https://github.com/kubernetes-client/javascript/blob/681af0941f306a1f6ac6a4d7622a5901c5ebb0e3/src/oidc_auth.ts#L68>
seems js is using idtoken only

my concern is if we are following the spec

tg123:
build failure: `/home/runner/work/csharp/csharp/tests/KubernetesClient.Tests/KubernetesClient.Tests.csproj : error NU1605:  KubernetesClient.Tests -> KubernetesClient -> IdentityModel.OidcClient 4.0.0 -> Microsoft.Extensions.Logging (>= 5.0.0)`

tg123:
> > BTW anyway to test it? we can integrate it into ci
> 
> @tg123 the way im testing is just running `dotnet run --project examples/simple` and ensuring that i no longer get an `Unauthorized` error. i think you would need to have oidc auth setup in your kube config. one where the id token is not expired and then ensuring that unexpired id token is used to authenticate requests instead of attempting to refresh the token. my kube config looks like this [#624 (comment)](https://github.com/kubernetes-client/csharp/issues/624#issuecomment-843651298). hopefully that helps somewhat, let me know if theres any more info i can provide

Any way to create a OIDC server and do test here? not a must but good to have
https://github.com/kubernetes-client/csharp/tree/master/tests/E2E.Tests

brendandburns:
I think it's going to be pretty hard to test this (and definitely beyond the scope of this PR)

Let's TODO the integration test and do the best we can in unit tests with mocks.

src/KubernetesClient/Authentication/OidcTokenProvider.cs
jarrett-confrey: a lot of these concepts are new to me but from what i can tell the javascript implementation of only using id token is correct. i started down this path initially but then was worried about breaking things for @arbielsk who implemented using the access token. im fairly confident now that we can remove the access token
>To identify the user, the authenticator uses the `id_token` (not the `access_token`) from the OAuth2 token response as a bearer token.

https://kubernetes.io/docs/reference/access-authn-authz/authentication/#openid-connect-tokens

src/KubernetesClient/Authentication/OidcTokenProvider.cs
```diff
@@ -14,23 +15,33 @@ public class OidcTokenProvider : ITokenProvider
...
             }
 
-            return new AuthenticationHeaderValue("Bearer", _accessToken);
+            return new AuthenticationHeaderValue("Bearer", _accessToken ?? _idToken);
+        }
+
+        private DateTimeOffset getExpiryFromToken()
+        {
+            var handler = new JwtSecurityTokenHandler();
+            var token = handler.ReadJwtToken(_accessToken ?? _idToken);
```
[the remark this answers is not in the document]
jarrett-confrey: i have removed the access token from OidcTokenProvider and now only use the id token

jarrett-confrey:
i appreciate the feedback! please let me know what else i can do to improve this pr

>build failure: /home/runner/work/csharp/csharp/tests/KubernetesClient.Tests/KubernetesClient.Tests.csproj : error NU1605: KubernetesClient.Tests -> KubernetesClient -> IdentityModel.OidcClient 4.0.0 -> Microsoft.Extensions.Logging (>= 5.0.0)

this has been corrected. i ended up reverting the `IdentityModel.OidcClient` bump as i was having trouble getting things to work with the same version of `Microsoft.Extensions.Logging` and i did not actually need any changes from this bump

>Any way to create a OIDC server and do test here? not a must but good to have

i looked into the e2e test a bit, it seems we would need to setup minikube with oidc
https://minikube.sigs.k8s.io/docs/tutorials/openid_connect_auth/

>I think it's going to be pretty hard to test this (and definitely beyond the scope of this PR)
>Let's TODO the integration test and do the best we can in unit tests with mocks.

i agree that the e2e test effort might be out of scope for this pr. i have added some unit tests

tg123:
thanks for updating, I understand id token better

/LTGM

brendandburns:
/lgtm
/approve

Many thanks for this PR!

jarrett-confrey:
thanks everyone!

Issue #634, opened 2021-05-21
Could we make the `KubeConfigDefaultLocation`  public?


WeihanLi:
Could we public `KubeConfigDefaultLocation` in
<https://github.com/kubernetes-client/csharp/blob/master/src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs#L22>

Sometime I wanna analyze the kube config by myself, it may be useful for me to load the default kube config file.

tg123:
PR welcomed

PR #635, opened 2021-05-21
Expose kube config default location

WeihanLi:
Expose `KubeConfigDefaultLocation` so that user could use directly
Relate issue: #634


brendandburns:
/lgtm
/approve

Thanks!

Issue #636, opened 2021-05-23
InClusterConfig doesn't load namespace


lucastheisen:
The go client [loads the namespace](https://github.com/kubernetes/client-go/blob/f6ce18ae578c8cca64d14ab9687824d9e1305a67/tools/clientcmd/client_config.go#L579-L594)

```golang
func (config *inClusterClientConfig) Namespace() (string, bool, error) {
	// This way assumes you've set the POD_NAMESPACE environment variable using the downward API.
	// This check has to be done first for backwards compatibility with the way InClusterConfig was originally set up
	if ns := os.Getenv("POD_NAMESPACE"); ns != "" {
		return ns, false, nil
	}

	// Fall back to the namespace associated with the service account token, if available
	if data, err := ioutil.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/namespace"); err == nil {
		if ns := strings.TrimSpace(string(data)); len(ns) > 0 {
			return ns, false, nil
		}
	}

	return "default", false, nil
}
```

But when i ask the config loaded by [`InClusterConfig()`](https://github.com/kubernetes-client/csharp/blob/master/src/KubernetesClient/KubernetesClientConfiguration.InCluster.cs#L38-L56) for its `.Namespace` i get `null`.  It seems like this code should add the Namespace.  Something like:

```csharp
        public static KubernetesClientConfiguration InClusterConfig()
        {
            if (!IsInCluster())
            {
                throw new KubeConfigException(
                    "unable to load in-cluster configuration, KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT must be defined");
            }

            var rootCAFile = Path.Combine(serviceAccountPath, ServiceAccountRootCAKeyFileName);
            var host = Environment.GetEnvironmentVariable("KUBERNETES_SERVICE_HOST");
            var port = Environment.GetEnvironmentVariable("KUBERNETES_SERVICE_PORT");

            var namespace = Environment.GetEnvironmentVariable("POD_NAMESPACE")
                ?? File.ReadAllText(Path.Combine(serviceAccountPath, NamespaceFileName));

            return new KubernetesClientConfiguration
            {
                Host = new UriBuilder("https", host, Convert.ToInt32(port)).ToString(),
                Namespace = namespace,
                TokenProvider = new TokenFileAuth(Path.Combine(serviceAccountPath, ServiceAccountTokenKeyFileName)),
[...]

PR #640, opened 2021-05-28
Add support for loading namespaces from pod configuration.

brendandburns:
See #636

### DIFF
--- a/src/KubernetesClient/FileUtils.cs
+++ b/src/KubernetesClient/FileUtils.cs
@@ -1,0 +1,34 @@
+using System;
+using System.IO.Abstractions;
+
+namespace k8s
+{
+    public static class FileUtils
+    {
+        private static IFileSystem realFileSystem = new FileSystem();
+        private static IFileSystem currentFileSystem = null;
+
+        public static void InjectFilesystem(IFileSystem fs)
+        {
+            currentFileSystem = fs;
+        }
+
+        public static IFileSystem FileSystem()
+        {
+            return currentFileSystem != null ? currentFileSystem : realFileSystem;
+        }
+
+        public sealed class InjectedFileSystem : IDisposable
+        {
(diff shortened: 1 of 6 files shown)


src/KubernetesClient/FileUtils.cs
```diff
@@ -0,0 +1,34 @@
+using System;
+using System.IO.Abstractions;
+
+namespace k8s
+{
+    public static class FileUtils
```
tg123, line 6:
```suggestion
    internal static class FileUtils
```

tg123: or I think make `IFileSystem` a field of config object which would be better than global static.
but for test purpose, it is good.

tg123 (approved):

tg123:
some nits then /LGTM

brendandburns:
Committed your suggestion. Thanks!

brendandburns:
@tg123 any ideas on the test failure here? This e2e test passes on my windows machine.

brendandburns:
Hrm, now if failed on ubuntu. I will try to reproduce the issue.

tg123:
flaky  test
re-run passed

brendandburns:
Thanks we should de-flake that test.

tg123:
> Thanks we should de-flake that test.

sure i will take care of the flake tests.
The test is now counting how many heartbeats in a fixed time period, would fail if testing machine is slow and did not emit enough heartbeat.

brendandburns:
I'm happy to take a look at the test if you prefer, let me know.

tg123:
> I'm happy to take a look at the test if you prefer, let me know.

sure appreciate it
tests in 
LeaderElectionTests.cs

brendandburns:
Ok, I will look to deflake those tests.

PR #653, opened 2021-06-27
fix leader election lease duration

mikenorgate:
Fix for the leader election lease duration not using total seconds


tg123:
thanks @mikenorgate, please fix the build and i will merge it

mikenorgate:
@tg123 build fixed

PR #655, opened 2021-07-01
Makes `realFileSystem` readonly

WeihanLi:
Makes `realFileSystem` read-only since it would not be updated


WeihanLi:
Seemed the failed test case is not related to this change, should I fix the test case firstly?

brendandburns:
/approve

Thanks for the PR!

PR #666, opened 2021-07-20
Add SaveKubeFile

### LINKED ISSUE #664: Save Configuration
Is there a method to save the configuration back to file? I'm looking for the equivalent of `kubectl config use-context`. I don't just want to change it in config object in memory, but persist it.

cdhunt:
Fixes #664

### DIFF
--- a/src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
+++ b/src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
@@ -690,6 +690,16 @@ namespace k8s
         }
 
         /// <summary>
+        ///     Saves Kube Config
+        /// </summary>
+        /// <param name="kubeconfig">Kube config file contents</param>
+        /// <param name="kubeconfigPath">Path to save the kubeconfig file</param>
+        public static void SaveKubeConfig(K8SConfiguration kubeconfig, string kubeconfigPath = null)
+        {
+            File.WriteAllTextAsync(kubeconfigPath ?? kubeconfig.FileName, Yaml.SaveToString(kubeconfig)).GetAwaiter().GetResult();
+        }
+
+        /// <summary>
         ///     Loads Kube Config
         /// </summary>
         /// <param name="kubeConfigs">List of kube config file contents</param>
--- a/tests/KubernetesClient.Tests/KubernetesClientConfigurationTests.cs
+++ b/tests/KubernetesClient.Tests/KubernetesClientConfigurationTests.cs
@@ -743,5 +743,27 @@ namespace k8s.Tests
                 Assert.Equal("some namespace", config.Namespace);
             }
         }
+
+        [Fact]
+        public void ChangeCurrentContext()
+        {
+
+            var k8sConfig =
+                KubernetesClientConfiguration.LoadKubeConfig(
+                    new FileInfo("assets/kubeconfig.yml"),
+                    false);
+            var cfg = KubernetesClientConfiguration.BuildConfigFromConfigObject(k8sConfig);
+            k8sConfig.CurrentContext = "horse-cluster";
+
+            var fileSystem = new MockFileSystem();
+
+            using (new FileUtils.InjectedFileSystem(fileSystem))
+            {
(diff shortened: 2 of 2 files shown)


cdhunt:
CLA signed

cdhunt:
So, `Yaml.SaveToString(kubeconfig)` works, but comments are lost and property order is all shuffled. 

Looks like porting the kubectl [modifyConfig](https://github.com/kubernetes/kubectl/blob/0f88fc6b598b7e883a391a477215afb080ec7733/pkg/cmd/config/set.go#L144) is the next logical approach.

brendandburns:
Any comments are lost on load, so yeah, I think modifying the config file directly is the only thing that will work for that use case.

src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
```diff
@@ -689,6 +689,23 @@ public static K8SConfiguration LoadKubeConfig(Stream kubeconfigStream)
             return LoadKubeConfigAsync(kubeconfigStream).GetAwaiter().GetResult();
         }
 
+        /// <summary>
+        ///     Saves Kube Config
+        /// </summary>
+        /// <param name="kubeconfig">Instance of the <see cref="K8SConfiguration"/> class to save</param>
+        /// <param name="kubeconfigPath">Path to save the kubeconfig file or <see cref="KubeConfigDefaultLocation"/> if not provided</param>
+        public static void SaveKubeConfig(K8SConfiguration kubeconfig, string kubeconfigPath = null)
```
brendandburns, line 697: Suggest just `Save(...)` since `KubernetesClientConfiguration.ConfigFile.SaveKubeConfig(...)` is kind of redundant.

src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
```diff
@@ -689,6 +689,23 @@ public static K8SConfiguration LoadKubeConfig(Stream kubeconfigStream)
...
+        {
+            if (kubeconfig == null)
+            {
+                throw new NullReferenceException(nameof(kubeconfig));
+            }
+
+            kubeconfigPath ??= kubeconfig.FileName ?? KubeConfigDefaultLocation;
+
+            File.WriteAllTextAsync(kubeconfigPath, Yaml.SaveToString(kubeconfig)).GetAwaiter().GetResult();
```
brendandburns, line 706: I would suggest adding both `Save` and `SaveAsync` and then just calling `SaveAsync` from `Save` with an await.

tests/KubernetesClient.Tests/KubernetesClientConfigurationTests.cs
```diff
@@ -743,5 +743,29 @@ public void LoadInClusterNamespace()
...
+            var fileSystem = new MockFileSystem();
+
+            using (new FileUtils.InjectedFileSystem(fileSystem))
+            {
+                KubernetesClientConfiguration.SaveKubeConfig(k8sConfig, "assets/kubeconfig-Save.yml");
+                var newConf = File.ReadAllText("assets/kubeconfig-Save.yml");
+
+                Assert.True(File.Exists("assets/kubeconfig-Save.yml"));
+                Assert.Matches("current-context: federal-context", newConf);
+                Assert.Matches(@"cluster: horse-cluster[\n\r]{1,2}\W+user: green-user[\n\r]{1,2}\W+namespace: chisel-ns[\n\r]{1,2}\W+name: federal-context", newConf);
```
brendandburns, line 767: This is brittle, I would suggest re-loading the KubeConfig file into a different configuration object and comparing the individual fields.

brendandburns:
Thanks for the PR, some initial comments.

tests/KubernetesClient.Tests/KubernetesClientConfigurationTests.cs
cdhunt: Agreed. However, my end goal is to produce an equivalent file with only the expected changes, not just an object. So, I ultimately want to validate the text. My current logic would be valid if only comparing object properties, but the file is very different.

Issue #563, opened 2021-02-17
Generating the API fails


qmfrederik:
I'm trying to re-generate the API, so that this client includes kubernetes-client/gen#180

I have the git repositories checked out at ~/git/gen and ~/git/csharp.

Running this command:

```shell
vagrant@ubuntu-focal:~/git/csharp/src/KubernetesClient$ ../../../gen/openapi/csharp.sh generated/ ../../csharp.settings 
```

eventually fails like this:

```
[INFO] AutoRest code generation utility [cli version: 3.0.6339; node: v14.15.5, max-memory: 8192 MB]
[INFO] (C) 2018 Microsoft Corporation.
[INFO] https://aka.ms/autorest
[INFO] Network Enabled: true
[INFO] NOTE: AutoRest core version selected from configuration: ~3.0.6298.
[INFO] Starting @autorest/core from /tmp/.autorest/@autorest_core@3.0.6373
[INFO]    Loading AutoRest core      '/tmp/.autorest/@autorest_core@3.0.6373/node_modules/@autorest/core/dist' (3.0.6373)
[INFO] No configuration found at 'file:///'.
[INFO]    Including configuration file 'file:///tmp/.autorest/@autorest_core@3.0.6373/node_modules/@autorest/core/resources/inspect.md'
[INFO]    Including configuration file 'file:///tmp/.autorest/@autorest_core@3.0.6373/node_modules/@autorest/core/resources/directives.md'
[INFO]    Including configuration file 'file:///tmp/.autorest/@autorest_core@3.0.6373/node_modules/@autorest/core/resources/pipeline.md'
[INFO]    Including configuration file 'file:///tmp/.autorest/@autorest_core@3.0.6373/node_modules/@autorest/core/resources/loader-openapi.md'
[INFO]    Including configuration file 'file:///tmp/.autorest/@autorest_core@3.0.6373/node_modules/@autorest/core/resources/miscellaneous.md'
[INFO]    Including configuration file 'file:///tmp/.autorest/@autorest_core@3.0.6373/node_modules/@autorest/core/resources/patches.md'
[INFO]    Including configuration file 'file:///tmp/.autorest/@autorest_core@3.0.6373/node_modules/@autorest/core/resources/plugin-powershell.md'
[INFO]    Including configuration file 'file:///tmp/.autorest/@autorest_core@3.0.6373/node_modules/@autorest/core/resources/plugin-azureresourceschema.md'
[...]

PR #669, opened 2021-07-24
port autorest2 template to local generate tools

tg123:
Port <https://github.com/Azure/autorest.csharp/tree/master/src/vanilla/Templates/Rest> template to local generate tools

**Why** #563
autorest2 is no longer maintained and we need to find a new swagger generator for the sdk.
also, there are many bugs caused by generator, for example, throw when 201 returned from server, which are not easy to get fixed.

the candidates are:
 * openapi-generator https://github.com/OpenAPITools/openapi-generator
 * autorest3

openapi-generator is most reasonable one, but new generator will cause tons of breaking changes.

as a result, templates form autorest were ported into mustache template and filled with nswagger objects. this is already used in watch and automapper code generation.

**roadmap**
near future, fix bugs and missing features due to autorest2, for example, expose client for keepalive.
long run, maybe act as a bridge to open-generator in order to have a smooth migration.

**changes in this generator**
- 200 201 202 are all treated as success return code
- IDictionary<string, IList<>> is now generated correctly (not in autorest2)
- parameter `pretty` in watch was changed to `string` to align with kubernetes impl

**test**
all testcases passed and models field compared
generated interface/impl/extension only and run test to ensure generated code are compatible with old codes.


tg123:
gh test #670

PR #671, opened 2021-07-25
Copy to/from pod

### LINKED ISSUE #464: Is there a C-Sharp method/code-snippet to upload a file from my local machine to a pod?

hol430:
Implementation for copy from/to a pod.

Important note, to make a copy, it requires that the tar binary is present in your container image, like the kubectl cp command.

This should eventually resolve #464

- [ ] Copy file from a pod
- [ ] Copy directory from a pod
- [ ] Copy file to a pod
- [ ] Copy directory to a pod

### DIFF
--- a/src/KubernetesClient/IKubernetes.Copy.cs
+++ b/src/KubernetesClient/IKubernetes.Copy.cs
@@ -140,10 +140,10 @@ namespace k8s
         /// The container in which the source directory exist.
         /// </param>
         /// <param name="sourceFolderPath">
-        /// The source file path.
+        /// The source directory path.
         /// </param>
         /// <param name="destinationFolderPath">
-        /// The destination file path.
+        /// The destination directory path.
         /// </param>
         /// <param name="cancellationToken">
         /// A <see cref="CancellationToken"/> which can be used to cancel the asynchronous operation.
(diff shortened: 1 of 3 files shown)


brendandburns:
Thanks for the PR! (and sorry for the delay in responding) the CI/CD is failing can you look into that?

hol430:
Thanks - yep, this is far from finished, I just need to find some more time to look into this. Not surprised CI is failing, as the code as it stands doesn't work in all cases yet. If I recall, I was having problems which were related to #397, but I may have a workaround. Hoping to have some time freed up in the next week or two, at which point I'll have another go at this.

csurfleet:
Hi @hol430 are you still working on this? I'm happy to take a look at the failures as I will need this functionality pretty soon but don't want to step on your toes if you're in the middle of it

hol430:
Hi @csurfleet - no, I haven't thought about this in a few weeks unfortunately. I think I was having problems closing the stream connected to stdin of the base64 process running in the pod. May have been a socket problem - if memory serves, I had some similar (identical?) code somewhere which was streaming to/from a base64 process running on my local machine without any problems.

Anyway, the CI failures will be because the feature doesn't actually work yet (for files which aren't tiny). Not sure when I will get time to have another go at this - so feel free to close this and/or create a new PR. Happy to talk further about the code that's already there - if my memory serves :).

PR #674, opened 2021-07-29
Fix typos

Gregordy:
This enhances the main README by fixing typos and adding missing words.


tg123:
i think the maybe `an` is the only typo

brendandburns:
Some of this is style/personal choice, but there are a few more typos than just the `an` (e.g. `the preferred way` ...)

Seems fine to merge.

/lgtm
/approve

Issue #675, opened 2021-08-04
LeaderElectionTest is flaky


brendandburns:
```
A total of 1 test files matched the specified pattern.
[xUnit.net 00:00:11.12]     k8s.Tests.LeaderElection.LeaderElectionTests.LeaderElection [FAIL]
  Failed k8s.Tests.LeaderElection.LeaderElectionTests.LeaderElection [3 s]
  Error Message:
   Assert.True() Failure
Expected: True
Actual:   False
  Stack Trace:
     at k8s.Tests.LeaderElection.LeaderElectionTests.LeaderElection() in D:\a\csharp\csharp\tests\KubernetesClient.Tests\LeaderElection\LeaderElectionTests.cs:line 187
```

brendandburns:
Closing (for now) since we merged #676

PR #676, opened 2021-08-07
fix flaky leader election tests

tg123:
#675

reason the test is flaky is that the test env task scheduler is not stable

how fix works?
- force 32 threads to ensure tasks switch smoothly
- use countdown latch to sync in case of task switch too slow

verify
re-run manually, got 10 successful runs in a row

### DIFF
--- a/tests/KubernetesClient.Tests/LeaderElection/LeaderElectionTests.cs
+++ b/tests/KubernetesClient.Tests/LeaderElection/LeaderElectionTests.cs
@@ -7,16 +7,25 @@ using System.Threading.Tasks;
 using k8s.LeaderElection;
 using Moq;
 using Xunit;
+using Xunit.Abstractions;
 
 namespace k8s.Tests.LeaderElection
 {
     public class LeaderElectionTests
     {
-        public LeaderElectionTests()
+        private static ITestOutputHelper xoutput;
+
+        public LeaderElectionTests(ITestOutputHelper output)
+            : base()
         {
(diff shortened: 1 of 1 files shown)


brendandburns:
/lgtm
/approve

Thanks!

tests/KubernetesClient.Tests/LeaderElection/LeaderElectionTests.cs
```diff
@@ -253,21 +276,15 @@ public void LeaderElectionWithRenewDeadline()
                 leaderElector.RunAsync().Wait();
             });
 
-            countdown.Wait(TimeSpan.FromSeconds(10));
-
-            // TODO flasky
-            // Assert.Equal(9, electionHistory.Count);
+            countdown.Wait(TimeSpan.FromSeconds(15));
```
WeihanLi, line 279: 15 seconds is a little magic I think, wondering that maybe we could just wait and specific a longer test case timeout, and we could remove `ThreadPool.SetMinThreads`/`ThreadPool.SetMaxThreads`

<https://stackoverflow.com/a/20283707/5519747>

tg123: the reason is await task.delay may sleep much longer than expected 
here the 15s is to ensure all tasks have enough time to run

running on my local machine, the task.delay sleeps 0-5ms more than expected. however on gh testing nodes, it would be more than 1s sometimes, which will cause the test fail

PR #679, opened 2021-08-09
General Updates

Christian-Oleson:
I'm reading through the code to get a better sense of what is covered and where we need to add features to be in parity with Kubernetes' abilities. As I go through this, I'm finding some issues in the code and correcting where possible.

### DIFF
--- a/src/KubernetesClient/Enums/WatchEventType.cs
+++ b/src/KubernetesClient/Enums/WatchEventType.cs
@@ -1,0 +1,32 @@
+﻿using System.Runtime.Serialization;
+
+namespace k8s.Enums
+{
+    /// <summary>Describes the type of a watch event.</summary>
+    public enum WatchEventType
+    {
+        /// <summary>Emitted when an object is created, modified to match a watch's filter, or when a watch is first opened.</summary>
+        [EnumMember(Value = "ADDED")]
+        Added,
+
+        /// <summary>Emitted when an object is modified.</summary>
+        [EnumMember(Value = "MODIFIED")]
+        Modified,
+
+        /// <summary>Emitted when an object is deleted or modified to no longer match a watch's filter.</summary>
+        [EnumMember(Value = "DELETED")]
+        Deleted,
+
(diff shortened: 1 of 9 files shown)


Christian-Oleson:
I signed the CLA.

src/KubernetesClient/Enums/WatchEventType.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System.Runtime.Serialization;
+
+namespace k8s.Enums
+{
+    /// <summary>Describes the type of a watch event.</summary>
+    public enum WatchEventType
```
tg123, line 6: why do you want to move ns?
this is a breaking change

src/KubernetesClient/generated/IKubernetes.Watch.cs
```diff
@@ -7,7 +7,8 @@
 using System; 
 using System.Collections.Generic; 
 using System.Threading; 
-using System.Threading.Tasks; 
+using System.Threading.Tasks;
```
tg123, line 10: generated code should be modified by code generator only

tg123 (commented):
not quite get the idea of this PR?
seems nothing changed but some code format?

src/KubernetesClient/Enums/WatchEventType.cs
Christian-Oleson: I can not move namespaces, but it seems (and perhaps I'm wrong here) that we should break up classes so they don't all live in the root of the source?

src/KubernetesClient/generated/IKubernetes.Watch.cs
Christian-Oleson: > not quite get the idea of this PR?
> seems nothing changed but some code format?
Yes, more or less. This is just an initial PR to take a look at the code, clean up some issues I found while reading along the way, and get a first PR into the repo. I wasn't aware that some of this was code generated, so I'll take a look further and go from there. 😄

Christian-Oleson: Ah, I see now the error. This is generated code. Closing the pr...

PR #700, opened 2021-09-16
Upgrade Microsoft.Rest.ClientRuntime from 2.3.10 to 2.3.23

### LINKED ISSUE #699: Upgrade Microsoft.Rest.ClientRuntime dependency from 2.3.10 to 2.3.23 to fix security vulnerability CVE-2019-0820
The Microsoft.Rest.ClientRuntime dependency is currently on 2.3.10 only, which causes a security alert when scanning the package using the [Black Duck dependency analyser](https://www.blackducksoftware.com/).

Microsoft.Rest.ClientRuntime 2.3.10 is indeed referencing NETStandard.Library 1.6.1 which itself references System.Text.RegularExpressions 4.3.0 which is known for having the security vulnerability CVE-2019-0820.

blackduck

One client of my company requires that our software and all its dependencies don't raise any security alert on Black Duck, that's how we discovered this dependency security issue and so we need a fix for it.

Simply upgrading Microsoft.Rest.ClientRuntime to 2.3.23 will be enough to fix the issue, with no other code change.

I'll be happy to do a PR myself with the change if nobody disagree with it.

RemiGaudin:
Upgrade Microsoft.Rest.ClientRuntime from 2.3.10 to 2.3.23 to fix security vulnerability CVE-2019-0820.

Fixes #699

### DIFF
--- a/src/KubernetesClient/KubernetesClient.csproj
+++ b/src/KubernetesClient/KubernetesClient.csproj
@@ -34,7 +34,7 @@
     <PackageReference Include="Fractions" Version="4.0.1" />
     <PackageReference Include="Nerdbank.GitVersioning" Version="3.3.37" PrivateAssets="all" />
     <PackageReference Include="Portable.BouncyCastle" Version="1.8.1.3" />
-    <PackageReference Include="Microsoft.Rest.ClientRuntime" Version="2.3.10" />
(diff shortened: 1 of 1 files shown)


RemiGaudin:
Close this PR as I commited with the wrong email address (sorry I'm new to github PR...)

PR #702, opened 2021-09-16
Upgrade Microsoft.Rest.ClientRuntime from 2.3.10 to 2.3.23

### LINKED ISSUE #699: Upgrade Microsoft.Rest.ClientRuntime dependency from 2.3.10 to 2.3.23 to fix security vulnerability CVE-2019-0820

RemiGaudin:
Upgrade Microsoft.Rest.ClientRuntime from 2.3.10 to 2.3.23 to fix security vulnerability CVE-2019-0820.

Fixes #699

### DIFF
--- a/src/KubernetesClient/KubernetesClient.csproj
+++ b/src/KubernetesClient/KubernetesClient.csproj
@@ -34,7 +34,7 @@
     <PackageReference Include="Fractions" Version="4.0.1" />
     <PackageReference Include="Nerdbank.GitVersioning" Version="3.3.37" PrivateAssets="all" />
     <PackageReference Include="Portable.BouncyCastle" Version="1.8.1.3" />
-    <PackageReference Include="Microsoft.Rest.ClientRuntime" Version="2.3.10" />
+    <PackageReference Include="Microsoft.Rest.ClientRuntime" Version="2.3.23" />
     <PackageReference Include="prometheus-net" Version="4.1.1" />
     <PackageReference Include="System.IdentityModel.Tokens.Jwt" Version="6.11.1" />
     <PackageReference Include="System.IO.Abstractions" Version="13.2.33" />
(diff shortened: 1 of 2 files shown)


RemiGaudin:
@brendandburns I'm sorry about the failing tests... The 5 tests that are failing are the `NamespacedPodExecAsyncXXX` ones in the `PodExecTests` serie. After digging a little bit into the code, it seems that all five tests are failing when using `kubernetesMock.Object`, which itself is calling a `k8s.Kubernetes` constructor that then raises a System.ArgumentNullException.

I'm afraid I will need some helps or guidances here as I feel pretty lost in this test code... Is it the `handlers` argument for the constructor which is null here? I followed up the call stack and it seems this constructor is called from the other constructor `Kubernetes(ServiceClientCredentials credentials, params DelegatingHandler[] handlers)`, but what is calling this other constructor? I don't manage to do the link between the `kubernetesMock.Object` using and this constructor call...

If you can take time to assist me on this issue it would be very much appreciated, thank you!

tg123:
I remember I gave updating after found there is breaking changes
but it is a must if there is a security issue

RemiGaudin:
@brendandburns / @tg123 Actually simply setting kubernetesMock.CallBase = true is enough to fix the 5 tests that were failing. I'm not sure about why this setting is now required though...

tg123:
thanks 

/LTGM

RemiGaudin:
@brendandburns / @krabhishek8260 / @tg123 Is there anybody able to lgtm & approved this PR? I'm sorry to put some pressure, but actually this issue is a blocker for us as our client keep refusing to install our software as long as this dependency has the vulnerability issue. Many thanks!

PR #731, opened 2021-10-19
Made a change to avoid calling _httpResponse.Content.ReadAsStringAsync() and deserialize from the content stream.

VladimirKhvostov:
I was investigating high memory usage and noticed that KubernetesClient uses a lot of memory.

Here is an excerpt from the ListSecretForAllNamespacesWithHttpMessagesAsync:

```
                _responseContent = await _httpResponse.Content.ReadAsStringAsync().ConfigureAwait(false);
                try
                {
                    _result.Body = SafeJsonConvert.DeserializeObject<V1SecretList>(_responseContent, DeserializationSettings);
                }
```
_httpResponse.Content.ReadAsStringAsync() creates a huge strings that are allocated on LOH, which is not ideal, since objects on LOH a collected only by Gen 2 GC.

In this PR I made a change to deserialize directly from the response content stream.

I wrote a little test program that calls ListSecretForAllNamespacesWithHttpMessagesAsync(). 
Before this changes, I needed about 115 MB of heap to load the secrets, after the change: 65 MB.


gen/KubernetesGenerator/ParamHelper.cs
```diff
@@ -19,11 +19,12 @@ public ParamHelper(GeneralNameHelper generalNameHelper, TypeHelper typeHelper)
 
         public void RegisterHelper()
         {
-            Helpers.Register(nameof(IfParamCotains), IfParamCotains);
```
VladimirKhvostov, line 22: Fixed typo

gen/KubernetesGenerator/templates/Kubernetes.cs.template
```diff
@@ -446,40 +475,26 @@ namespace k8s
                 throw ex;
             }
             // Create Result
-            var _result = new HttpOperationResponse{{GetReturnType operation "<>"}}();
-            _result.Request = _httpRequest;
-            _result.Response = _httpResponse;
-            {{#IfReturnType operation "any"}}
-            // Deserialize Response
-            if ((int)_statusCode == 200 || (int)_statusCode == 201 || (int)_statusCode == 202)
```
VladimirKhvostov, line 454: There is no need to check that _statusCode is 200 - 202. If this was not the case, HttpOperationException would have been thrown on line 475.

gen/KubernetesGenerator/templates/Kubernetes.cs.template
```diff
@@ -310,6 +309,39 @@ namespace k8s
             CustomInitialize();
         }
 
+        private async Task<HttpOperationResponse<T>> CreateResultAsync<T>(HttpRequestMessage httpRequest, HttpResponseMessage httpResponse, bool? watch, CancellationToken cancellationToken)
```
VladimirKhvostov, line 312: Creating this method reduced the size of KubernetesClient.dll by about 50K (1 percent)

tg123 (commented):
Thanks the PR, just some nits then LGTM

please sign cla
please make generate code a separate commit (squash though)


some background of this template:
the template is copied from autorest2 recently, as a result, you may see so many weird and duplicated code. it DOES need a clean up and optimization.
the reason it is still using mustache even nustable is not that powerful (no else ...) is because some day the template may merge to `openapi generator`

some planned changes to share:
 * newton.json -> system.json
 * remove `ServiceClient`, to use our own http client

brendandburns:
/approve

Thanks for the PR!

PR #744, opened 2021-11-12
add generic create

tg123:
- add create to generic client
- add tests to prevent regressions in #741


brendandburns:
/lgtm
/approve

(assuming the tests pass :)

PR #747, opened 2021-11-23
Fix x-kubernetes-int-or-string in CRD can not be desterilized correctly.

Binyang2014:
If CRD contains:
```yaml
  versions:
  - name: v1alpha1
    schema:
      openAPIV3Schema:
        description: Middleware is a specification for a Middleware resource.
        x-kubernetes-int-or-string: true
        required:
        - metadata
        - spec
        type: object
```
The `x-kubernetes-int-or-string` field can not be desterilized  to `XKubernetesIntOrString ` in `V1JSONSchemaProps` correctly. This because we use `CamelCaseNamingConvention` .This will match the field name with `CamelCase` style not exact as json Property name.

Disable naming conventions for this case

### DIFF
--- a/src/KubernetesClient/Yaml.cs
+++ b/src/KubernetesClient/Yaml.cs
@@ -225,7 +225,7 @@ namespace k8s
                         continue;
                     }
 
-                    var yamlAttribute = new YamlMemberAttribute { Alias = jsonAttribute.PropertyName };
+                    var yamlAttribute = new YamlMemberAttribute { Alias = jsonAttribute.PropertyName, ApplyNamingConventions = false };
                     builder.WithAttributeOverride(type, property.Name, yamlAttribute);
                 }
             }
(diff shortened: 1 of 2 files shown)


tg123:
could you please add some testcases for it
i remember this is due to property type info missing

Binyang2014:
> could you please add some testcases for it i remember this is due to property type info missing

OK, I will add some UTs.
It does not related to property type info missing, it's another issue and we also encountered.

For `x-kubernetes-int-or-string`, we already define the type of filed at
https://github.com/kubernetes-client/csharp/blob/e45fa773bade29c37f00198c701543b6c2c122c7/src/KubernetesClient/generated/Models/V1JSONSchemaProps.cs#L585-L586

But we still can not desterilize the field correctly due to naming convention.

Binyang2014:
@tg123 PTAL

Binyang2014:
@tg123 Can you help to take a look? We need this change in production environment.

tg123:
/LGTM

@Binyang2014 sorry, did not notice PR updated

PR #755, opened 2021-12-21
remove the dependency of unprocessed swagger file

tg123:
code generator no long requires `swagger.json.unprocessed`
swagger.json from `kubernetes-client/gen` only


tg123:
generated code is exactly the same as before

PR #766, opened 2022-01-13
Add environment variable checking for in cluster configs.

### LINKED ISSUE #765: IsInCluster does not check EnvironmentVariables
The go client checks for the presence of the "KUBERNETES_SERVICE_HOST" and "KUBERNETES_SERVICE_PORT" environment variables to determine if the client is running inside of a k8s cluster. (https://github.com/kubernetes/client-go/blob/6d69eb8ad66c8962b6ce2f610d46fa3ab7d23afb/tools/clientcmd/client_config.go#L603)

It looks like it used to check in the same way in the csharp client but was removed in this commit (https://github.com/kubernetes-client/csharp/commit/762c8af640e41d38971e09578174d20ef1fd1735#diff-f3d84ec98a6f0ec4d4a1f0e2f08d352506f3059c231dc3d3413bd3e08c1113a2)

Was this check accidently removed? Since line 24/25
https://github.com/kubernetes-client/csharp/blob/0e7ddae91706f34e54658f4617f7a9ff9f6a165a/src/KubernetesClient/KubernetesClientConfiguration.InCluster.cs#L24
currently are unused and are dead code.

brendandburns:
Closes #765

### DIFF
--- a/src/KubernetesClient/KubernetesClientConfiguration.InCluster.cs
+++ b/src/KubernetesClient/KubernetesClientConfiguration.InCluster.cs
@@ -24,6 +24,11 @@ namespace k8s
             var host = Environment.GetEnvironmentVariable("KUBERNETES_SERVICE_HOST");
             var port = Environment.GetEnvironmentVariable("KUBERNETES_SERVICE_PORT");
 
+            if (String.IsNullOrEmpty(host) || String.IsNullOrEmpty(port))
+            {
+                return false;
+            }
+
             var tokenPath = Path.Combine(ServiceAccountPath, ServiceAccountTokenKeyFileName);
             if (!FileUtils.FileSystem().File.Exists(tokenPath))
             {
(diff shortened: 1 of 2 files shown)


brendandburns:
/hold

looks like I need to update the tests.

src/KubernetesClient/KubernetesClientConfiguration.InCluster.cs
```diff
@@ -24,6 +24,11 @@ public static bool IsInCluster()
             var host = Environment.GetEnvironmentVariable("KUBERNETES_SERVICE_HOST");
             var port = Environment.GetEnvironmentVariable("KUBERNETES_SERVICE_PORT");
 
+            if (String.IsNullOrEmpty(host) || String.IsNullOrEmpty(port))
```
tg123, line 27:
```suggestion
            if (string.IsNullOrEmpty(host) || string.IsNullOrEmpty(port))
```

happier lint

tg123 (commented):
/LGTM
nits

PR #771, opened 2022-01-30
AddRange extension enhancement

WeihanLi:
Small enhancement for AddRange enhancement


tg123:
/LGTM

btw you can rebase to avoid `There is no currently active test.` error

PR #776, opened 2022-02-08
Support .net 6 for tcp keep alive

zhiweiv:
The tcp keep alive only work for .net 5 now, it should work for .net 6 too.


zhiweiv:
Close this to for cla, will open another one

PR #777, opened 2022-02-08
Support .net 6 for tcp keep alive

zhiweiv:
The tcp keep alive only work for .net 5 now, it should work for .net 6 too.


tg123:
/LGTM

omg i remember i `s/NET5_0/NET5_0_OR_GREATER` when upgrading to net6

PR #782, opened 2022-02-18
fix flaky logger

tg123:


tg123:
failed 3/20 due to github action env

PR #787, opened 2022-02-25
fix Unable to process file command 'env' successfully.

tg123:
fix from https://github.com/dotnet/Nerdbank.GitVersioning/issues/683


tg123:
10 succ in a row

PR #791, opened 2022-02-27
Update System.Text.Json dependency

WeihanLi:
- Fix the condition
- Remove `System.Text.Json` dependency for .NET 6,   still required for .NET 5 because of new json features introduced in `6.0.0`

### DIFF
--- a/src/KubernetesClient/KubernetesClient.csproj
+++ b/src/KubernetesClient/KubernetesClient.csproj
@@ -47,7 +47,7 @@
     <PackageReference Include="IdentityModel.OidcClient" Version="4.0.0" />
 
     <PackageReference Include="Portable.BouncyCastle" Version="1.8.10" Condition="'$(TargetFramework)' == 'netstandard2.1'" />
-    <PackageReference Include="System.Text.Json" Version="6.0.0" Coondition="'$(TargetFramework)' == 'netstandard2.1'" />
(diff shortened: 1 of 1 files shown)


tg123:
The library is built-in as part of the shared framework for .NET Core 3.0 and later versions
https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-overview?pivots=dotnet-5-0#how-to-get-the-library

WeihanLi:
> The library is built-in as part of the shared framework for .NET Core 3.0 and later versions
> https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-overview?pivots=dotnet-5-0#how-to-get-the-library

The build would fail if remove the dependency for .NET 5, the built-in version seemed is not latest version

tg123:
tests proj targets net5 as well, could you please elaborate yours

WeihanLi:
The condition did not work before because of the condition typo
``` diff
- Coondition
+ Condition
```

tg123:
lol this is very interesting.
the `coondition` here was intent to make net5+ use their own builtin json library instead of a dependency
it equals to no condition today

i tested fixing the typo and found serval build errors and understand the issue eventually.

the correct way, i believe, to fix it is to fix the typo and workaround those features in .6
OR
change to condition to `==netstandard2.1 or ==net5` in case of adopting net7

seems the second one is easier as netcore3.1 and net5 are approaching the eol

tg123:
thanks for the PR 

/LGTM

Issue #743, opened 2021-11-10
Framework no longer supported


KnicKnic:
In PR https://github.com/kubernetes-client/csharp/pull/603 the code was switched from emitting .netstandard 2.0 to .netstandard 2.1 . Can we change this back to .netstandard 2.0 Otherwise cannot use this package with .net Framework. 
@tg123

tg123:
lol going to drop netstandard2.1 as well for crypot and json api support
May I know why do you'd like netstandard2?
it is very easy to upgrade to net5/6

brendandburns:
Since dotnet 2.1 is EOL'd in favor of 3.1 or 5/6 I don't think we really can or should support them, users should upgrade.

I'm not sure that the work needed to support a deprecated platform is worth it for this project given the limited resources that we currently have.

KnicKnic:
@brendandburns  please see table https://docs.microsoft.com/en-us/dotnet/standard/net-standard#net-implementation-support notice .netstandard vs .net vs .net core vs .netframework.
* .netstandard is the compatibility layer
* .net (can be though of as .netcore's new name, this is a gross simplification)


.Net CORE 2.1 is EOL this is not .net standard

tg123:
netcoreapp3.0 which impl the one netstandard2.1 is also eol
https://devblogs.microsoft.com/dotnet/the-future-of-net-standard/ 
net5+ is new netstandard

However, I agree to provide support to netframework4.8 since it is still in support. (could be via netstandard2.0)
One solution to this might be splitting the nuget to multiple parts, making websocket related to a separated nuget.
this is also being discussed here https://github.com/kubernetes-client/csharp/pull/665#issuecomment-883983454

brendandburns:
@KnicKnic thanks for the explanation (all of the .NET versioning stuff is pretty opaque to me)

I can see the value in supporting .NET framework 4.8 to handle legacy apps. However, I'm not certain that we have the resources to do it. We would be open to proposals as to how to best accomplish this support and also to support the work to actually implement the proposal once we design it.

Perhaps the best approach would be to send a PR with a Markdown design document and we can comment/iterate on that until we are happy with the approach?

PR #793, opened 2022-03-03
netstandard2 extendable lib model, basic rest

tg123:
# The problem
 * old runtime support
    the kubernetes .net sdk is dropping support of old .net runtimes because of some `soft-required` features missing. for exmaple, websocket apis are missing in netstandard2 and certificates loading api are only in net5 or later. #743 

   however, it is difficult to upgrade .net runtime for many tricky reasons. it is time to welcome those users in future sdk versions.

  * generated model library
    feature request from many internal ms team or external users. mostly because a slim client is enough for their case and the fancy features introduces many unwanted dependencies.

# The proposal 

 the idea is to break the sdk into multiple smaller library:
- **KubernetesClient.Models (netstandard2):** generated models
- KubernetesClient.Models.Versioning (netstandard2): generated models version converter with injectable 
- **KubernetesClient.Basic** (netstandard2): generated rest stub, this is an _abstract_ class, http call related must be implemented in runtime specific libraries.
- **KubernetesClient** (net6+): full featured package for most users
- KubernetesClient.Net48 (net48): to support LTS windows server users until net48 eol
- KubernetesClient.Auth.XXX (netstandard2, net6?): cloud specific token generator
- KubernetesClient.Kubectl (net6): kubectl like api
- ....

**bold** items are current in this PR, others are planned only
users can build highly customized library with basic generated model and vanillarest libraries.


tg123:
#743

src/KubernetesClient.Models/ResourceQuantity.cs
```diff
@@ -57,7 +54,7 @@ namespace k8s.Models
     ///     cause implementors to also use a fixed point implementation.
     /// </summary>
     [JsonConverter(typeof(QuantityConverter))]
-    public partial class ResourceQuantity : IYamlConvertible
```
brendandburns, line 60: I'm worried about dropping YAML support here? Is that actually what is happening?

src/KubernetesClient.VanillaRest/VanillaKubernetes.cs
```diff
@@ -0,0 +1,74 @@
...
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+using k8s.Autorest;
+using System.Net.Http.Headers;
+
+
+namespace k8s
+{
+    public abstract partial class VanillaKubernetes
```
brendandburns, line 12: I'd prefer a more descriptive name than 'Vanilla' so that it is understandable to someone browsing the code.

src/KubernetesClient/Yaml.cs
```diff
@@ -30,6 +31,7 @@ public static class Yaml
                 .WithNamingConvention(CamelCaseNamingConvention.Instance)
                 .WithTypeConverter(new IntOrStringYamlConverter())
                 .WithTypeConverter(new ByteArrayStringYamlConverter())
+                .WithTypeConverter(new ResourceQuantityYamlConverter())
```
brendandburns, line 34: Ah, I see this now, nevermind my earlier comment.

src/KubernetesClient.Models/ResourceQuantity.cs
brendandburns: Nevermind, I see we added a converter below.

src/KubernetesClient.VanillaRest/VanillaKubernetes.cs
tg123: any ideas? I update the PR with details proposal, did not have a better name in mind.

tg123: ping @brendandburns

brendandburns: Apologies for the delay, perhaps `AbstractKubernetes` or `BasicKubernetes`?

Directory.Build.props
```diff
@@ -3,4 +3,30 @@
...
   </PropertyGroup>
+  
+  <PropertyGroup>
+    <Authors>The Kubernetes Project Authors</Authors>
+    <Copyright>2017 The Kubernetes Project Authors</Copyright>
+    <Description>Client library for the Kubernetes open source container orchestrator.</Description>
+
+    <PackageLicenseExpression>Apache-2.0</PackageLicenseExpression>
+    <PackageProjectUrl>https://github.com/kubernetes-client/csharp</PackageProjectUrl>
+    <PackageIconUrl>https://raw.githubusercontent.com/kubernetes/kubernetes/master/logo/logo.png</PackageIconUrl>
```
WeihanLi, line 14: Maybe we could use the package icon file since the URL is deprecated

https://docs.microsoft.com/en-us/nuget/reference/nuspec#iconurl
https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets#packing-an-icon-image-file

tg123: thanks, added #797 and lets fix it in a separate PR

PR #794, opened 2022-03-08
Code Enhancements

WeihanLi:
- Keep private members readonly in `StringTokenProvider`
- Use `long.TryParse` instead of `int.TryParse`
- Sealed internal types(maybe we could enforce this with some analyzer)

### DIFF
--- a/src/KubernetesClient/IntOrStringConverter.cs
+++ b/src/KubernetesClient/IntOrStringConverter.cs
@@ -21,7 +21,7 @@ namespace k8s.Models
         {
             var s = value?.Value;
 
-            if (int.TryParse(s, out var intv))
+            if (long.TryParse(s, out var intv))
             {
                 writer.WriteNumberValue(intv);
                 return;
--- a/src/KubernetesClient/IntOrStringConverter.cs
+++ b/src/KubernetesClient/IntOrStringConverter.cs
@@ -1,6 +1,6 @@
 namespace k8s.Models
 {
-    internal class IntOrStringConverter : JsonConverter<IntstrIntOrString>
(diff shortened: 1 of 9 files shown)


src/KubernetesClient/IntOrStringConverter.cs
```diff
@@ -21,7 +21,7 @@ public override void Write(Utf8JsonWriter writer, IntstrIntOrString value, JsonS
         {
             var s = value?.Value;
 
-            if (int.TryParse(s, out var intv))
+            if (long.TryParse(s, out var intv))
```
WeihanLi, line 24: From the code<https://github.com/kubernetes-client/csharp/blob/01b2aa832cfdc6f2c89153fe0123edc0a13b1bbf/src/KubernetesClient/IntstrIntOrString.cs#L11-L14>, we could see `long` is acceptable, so maybe we should use `long.TryParse`

tg123, line 24: int str?

src/KubernetesClient/LineSeparatedHttpContent.cs
```diff
@@ -6,7 +6,7 @@
 
 namespace k8s
 {
-    internal class LineSeparatedHttpContent : HttpContent
+    internal sealed class LineSeparatedHttpContent : HttpContent
```
tg123, line 9: any reason to make internal class sealed?

WeihanLi: There's a ticket with some introduction https://github.com/dotnet/runtime/issues/49944

tg123 (approved):
/LGTM

PR #800, opened 2022-03-21
Support for package icon file

### LINKED ISSUE #797: warning NU5048: The 'PackageIconUrl'/'iconUrl' element is deprecated.
Maybe we could use the package icon file since the URL is deprecated

https://docs.microsoft.com/en-us/nuget/reference/nuspec#iconurl
https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets#packing-an-icon-image-file

_Originally posted by @WeihanLi in https://github.com/kubernetes-client/csharp/pull/793#discussion_r830584173_

WeihanLi:
Fixes #797 

Currently, I'm using the logo from the icon URL:
 <https://raw.githubusercontent.com/kubernetes/kubernetes/master/logo/logo.png>

There's a concern, I copied the logo to the repository, it may need to be updated when the source logo was updated.


src/KubernetesClient/KubernetesClient.csproj
```diff
@@ -8,6 +8,7 @@
     <PackageLicenseExpression>Apache-2.0</PackageLicenseExpression>
     <PackageProjectUrl>https://github.com/kubernetes-client/csharp</PackageProjectUrl>
     <PackageIconUrl>https://raw.githubusercontent.com/kubernetes/kubernetes/master/logo/logo.png</PackageIconUrl>
```
tg123, line 10: remove?

WeihanLi: From the document, we could keep this for better compatibility. However, it may not make much sense. I would remove it if you want.

https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets#packageicon

> To maintain backward compatibility with clients and sources that don't yet support PackageIcon, specify both PackageIcon and PackageIconUrl

brendandburns:
/lgtm
/approve

Let's keep the URL for now.

Thanks for the PR!

PR #806, opened 2022-03-24
move kubectl config and yaml related to model module

tg123:
expose 
 * KubernetesYaml
 * KubernetesJson

both them expose API for users to deal with Kubernetes favored data objects

- public static TValue Deserialize<TValue>(string)
- public static TValue Deserialize<TValue>(Stream)
- public static string Serialize(object value)


brendandburns:
This looks fine to me, but it is a breaking change right? Should we be incrementing at least the minor version?

tg123:
@brendandburns no worries, 7.2.x is not published yet, my plan is to have `net4.8` support in 7.2.x

PR #811, opened 2022-03-28
Reuse const strings and expose PluralName

stan-sz:
Both in `ModelExtensions.cs`


src/LibKubernetesGenerator/templates/ModelExtensions.cs.template
```diff
@@ -1,17 +1,17 @@
...
 // Code generated by https://github.com/kubernetes-client/csharp/tree/master/gen/KubernetesGenerator
-// Changes may cause incorrect behavior and will be lost if the code is
-// regenerated.
+// Changes may cause incorrect behavior and will be lost if the code is regenerated.
 // </auto-generated>
 namespace k8s.Models
 {
 {{#.}}
-    [KubernetesEntity(Group="{{GetGroup . }}", Kind="{{GetKind . }}", ApiVersion="{{GetApiVersion . }}", PluralName={{GetPlural .}})]
+    [KubernetesEntity(Group=KubeGroup, Kind=KubeKind, ApiVersion=KubeApiVersion, PluralName=KubePluralName)]
```
stan-sz, line 8: With this change, less hardcoded strings are embedded, rather a const string is reused.

PR #819, opened 2022-04-03
Restart sts daemonset deployment example

kibernetik542:


examples/restart/Program.cs
```diff
@@ -0,0 +1,89 @@
...
+{
+    var origin = new DateTime(1970, 1, 1, 0, 0, 0, 0, DateTimeKind.Utc);
+    var diff = date.ToUniversalTime() - origin;
+    return Math.Floor(diff.TotalSeconds);
+}
+
+async Task RestartDamonsetAsync(string name,string @namespace)
+{
+    var config = KubernetesClientConfiguration.BuildConfigFromConfigFile();
+    IKubernetes client = new Kubernetes(config);
```
brendandburns, line 17: why not `var client`?

examples/restart/Program.cs
```diff
@@ -0,0 +1,89 @@
...
+
+    var expected = JsonSerializer.SerializeToDocument(daemonSet);
+
+    var patch = old.CreatePatch(expected);
+    await client.PatchNamespacedDaemonSetAsync(new V1Patch(patch, V1Patch.PatchType.JsonPatch), name, @namespace);
+}
+
+async Task RestartDeploymentAsync(string name,string @namespace)
+{
+    var config = KubernetesClientConfiguration.BuildConfigFromConfigFile();
```
brendandburns, line 41: This code is duplicated in a bunch of places, perhaps it makes more sense to pass the client to the `RestartDeploymentAsync` function?

examples/restart/Program.cs
```diff
@@ -0,0 +1,89 @@
...
+using k8s.Models;
+
+double ConvertToUnixTimestamp(DateTime date)
+{
+    var origin = new DateTime(1970, 1, 1, 0, 0, 0, 0, DateTimeKind.Utc);
+    var diff = date.ToUniversalTime() - origin;
+    return Math.Floor(diff.TotalSeconds);
+}
+
+async Task RestartDamonsetAsync(string name,string @namespace)
```
brendandburns, line 14: nit space between the command and the `string` declaration (here and everywhere)

brendandburns: Also, please add explicit disables so that this doesn't trigger a style warning (or fix the style warning). Here and below.

brendandburns:
Thanks for the PR, a few cleanups and then I think it is ready to merge.

kibernetik542:
> Thanks for the PR, a few cleanups and then I think it is ready to merge.

I fixed all suggestions and I think now it should be fine

brendandburns:
/lgtm
/approve

Thanks!

PR #826, opened 2022-04-04
add tg123 to owners

tg123:


OWNERS
```diff
@@ -3,7 +3,9 @@
 approvers:
 - brendandburns 
 - krabhishek8260
```
brendandburns, line 5: May as well move krabhishek8260 to emeritus (here and below)

brendandburns:
One minor comment cc @krabhishek8260

tg123:
close as cannot edit branch

PR #830, opened 2022-04-14
init kubectl

tg123:
this is the init proj struct and first api `version`
https://github.com/kubernetes-client/csharp/projects/2

https://github.com/kubernetes-client/csharp/projects/2#card-80285137

the rest api is somewhat hard to use and need time to figure what is the correct api I should call. 
kubectl will provide users a kubectl style api and do not have to know how to do the right things e.g. `how to kubectl apply using c# client`

the PR is for public review the API style and introduce a place for others to contribute.
the namespace will be `k8s.kubectl.beta` for a while until most of apis are stable


tg123:
#831

brendandburns:
/lgtm
/approve

There is a Java equivalent here: https://github.com/kubernetes-client/java/tree/master/extended/src/main/java/io/kubernetes/client/extended/kubectl

PR #808, opened 2022-03-24
Kubernetes.Classic: add support for netstandard2.0 and net48

tg123:
fix #743

support basic API call for older .net runtime, watcher, exec etc. are not supported at the moment

PR #833, opened 2022-04-19
Set http 2.0 for .net 5.0+

zhiweiv:
In https://github.com/kubernetes-client/csharp/pull/808, http 2 was omitted on .net 5.0+, this causes watch hang because httpclient keepalive only work on http 2.

### DIFF
--- a/src/KubernetesClient/Kubernetes.cs
+++ b/src/KubernetesClient/Kubernetes.cs
@@ -104,7 +104,7 @@ namespace k8s
             var httpRequest = new HttpRequestMessage();
             httpRequest.Method = new HttpMethod(method);
             httpRequest.RequestUri = new Uri(BaseUri, relativeUri);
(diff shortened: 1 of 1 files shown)


tg123 (approved):
/LGTM

zhiweiv:
@tg123 
I am wondering, should all `NETSTANDARD2_1_OR_GREATER `be replaced to `NETSTANDARD2_1_OR_GREATER || NET5_0_OR_GREATER`?

There are another three in https://github.com/kubernetes-client/csharp/blob/master/src/KubernetesClient/Authentication/TokenFileAuth.cs.

tg123:
@zhiweiv you are right, lets fix them as well
I am thinking adding a test to enforce http2

zhiweiv:
I will submit another PR to update remains, this is only for watch hang issue.

PR #839, opened 2022-04-22
fix obsolete warning about KubernetesYaml

zhiweiv:
Fix obsolete warning about KubernetesYaml which appears in build and PR review page, replace with new method.


brendandburns:
/lgtm
/approve

Thanks for the PR!

PR #848, opened 2022-05-06
Update KubernetesClientConfiguration.ConfigFile.cs

RyanGrange:
Handle unset environment variables returning null values to Path.Combine


RyanGrange:
AWS Lambdas don't set a HOME environment variable, leading to a null argument for Path.Combine

"cause": {
    "errorType": "ArgumentNullException",
    "errorMessage": "Value cannot be null. (Parameter 'path1')",
    "stackTrace": [
      "at System.IO.Path.Combine(String path1, String path2)",
      "at k8s.KubernetesClientConfiguration..cctor()"
    ]
  }

This patch simply fixes the null to a valid directory character for the different OS environments.

tg123:
/LGTM

sad to know aws lambda is HOMELESS

tg123 (approved):

PR #850, opened 2022-05-07
8.0: API Group + v1.24

tg123:
- update api model to v1.24
- support api group

#### why this change
before api group, apis with same operationid, for example `getAPIResources`, were suffixed with a number which does not make any sense.
this change is also aligning api style with other sdk, for example, java

#### new api exmaple

before
```
client.ListNamespacedPod("default");
```

after
```
client.CoreV1.ListNamespacedPod("default");
```

something should follow up 
`client.CoreV1` creates a new object each time which can be optimized and cached


brendandburns:
/lgtm
/approve

Thanks!

PR #853, opened 2022-05-08
use explicitly impl instead of new

tg123:
follow up to #850 
the `client` object impled all operation interfaces
no `new` needed, performance improved

extra benefit: Kubernetes obj does not have to use api group to call api, easier for users to migrate

code below is valid again
```
var client = new Kubernetes(config);
client.ListNamespacedPod("default");
```


src/KubernetesClient.Basic/AbstractKubernetes.cs
```diff
@@ -5,11 +5,71 @@ namespace k8s;
...
+                        break;
+                }
+            }
+        }
+
+        public override string ToString()
+        {
+            if (parameters.Count > 0)
+            {
+                return "?" + string.Join("&", parameters);
```
brendandburns, line 50: nit: consider string interpolation `$"?{string.Join("&", parameters)}"`

brendandburns:
One tiny nit, feel free to leave unaddressed.

src/KubernetesClient.Basic/AbstractKubernetes.cs
tg123, line 50:
```suggestion
                return $"?{string.Join("&", parameters)}"
```

Issue #859, opened 2022-05-20
SSL Error when connecting to server with own CA Authority Chain


KLazarov:
### Description:
The client fails to connect to the Kubernetes cluser if the Kubernetes cluser is using a self signed CA chain. A potential fix is presented for the HTTP Client.

### Problem:
If the Kubernetes cluster has it's own self signed CA chain, then the Kubernetes client is unable to connect to it as it trows a SSL error. As seen here:
![bug report ssl error](https://user-images.githubusercontent.com/74538639/169517885-86d0ce78-fc4d-4385-9f79-09e2c472c71f.png)

The kube.conf provides a self signed CA Authority chain that proves that the SSL connection is actually secure. The CA chain has been tested in both kubectl, in Postman, and in the solution and all of them work flawlessly.

### The suggested solution:
The issue lies in the Microsoft provided `HttpClient`. Even if you add the CA Authority with a `HttpClientHandler`, it still produced the same SSL error. The solution that I found and propose here is based on: [Using custom root CA with HttpClient](https://stackoverflow.com/a/71788784). By providing a `ServerCertificateCustomValidationCallback` to the `HttpClient` it's now able to connect to Kubernetes clusters with their own CA Authority chain, as long as the Certificates are provided to the clients, without having to use `insecure-skip-tls-verify: true`.

The full code for the `HttpClient` is a follows:

```
public async Task<IActionResult> GetKubernetes()
{
    var handler = new HttpClientHandler();
    handler.ClientCertificateOptions = ClientCertificateOption.Manual;
    handler.SslProtocols = SslProtocols.Tls13;
    handler.ClientCertificates.Add(new X509Certificate2(<name_of_ca.pem_file>));
    handler.ServerCertificateCustomValidationCallback = ServerCertificateCustomValidation;
    var client = new HttpClient(handler);
    client.DefaultRequestHeaders.Authorization = new AuthenticationHeaderValue("Bearer", <bearer_token>);
    var result = client.GetAsync(<url_to_kubernetes_cluster>).GetAwaiter().GetResult();
    return Ok(result);
}
private static bool ServerCertificateCustomValidation(
    HttpRequestMessage requestMessage,
    X509Certificate2? certificate,
    X509Chain? chain,
    SslPolicyErrors sslErrors)
{
    if (certificate == null) return false;
    if (chain == null) return false;
    chain.ChainPolicy.TrustMode = X509ChainTrustMode.CustomRootTrust;
[...]

PR #860, opened 2022-05-20
Updated CertificateValidationCallBack to build the CaCert bundle with respect to the rootChain certs to verify that they are correct.

KLazarov:
Implemented fix proposed at: https://github.com/kubernetes-client/csharp/issues/859

### Problem:

Currently the cert verification works only if there is no multi level CA chain that it needs to verify on. If the chain is too long, it will mark `isTrusted` as `false`.

### Suggestion

Instead of trying to directly compare the certs, we actually  build the cert with the help of the chain and verify that it's correct. If the cert or the chain are broken, then `isTrusted` will be `false`. If the chain and cert is correct, then `isTrusted` will be `true`.


tg123:
could please sign cla and add testcases?

KLazarov:
I have now added the tests for the issue that my code fixes. They are at the bottom of CertificateValidationTests.cs.

I'm still waiting from my company to completly fill out the EasyCLA as we got a bit stuck on it during the weekend.

KLazarov:
The EasyCLA has now been accepted.

tg123:
I looked into the code and seems your #859 code
```
chain.ChainPolicy.CustomTrustStore.Add(ca)
```
makes more sense

here the code is verifying if each ca cert on the cert's chain
should that be verifying cert's chain on a trusted ca?


what is your idea?

KLazarov:
The issue that I stumbled upon and I'm trying to fix here is if there is a chain that is multiple steps long in a single file provided by the Service Account.

As we are running a managed Kubernetes cluster, we get our Certificates generated by the provider, and in our case there are multiple certificates in the same file, and they describe the full chain. By running `chain.Build(cert)` it verifies that the cert that it tries to connect with is actually correctly generated from the Chain.

From what I see
```chain.ChainPolicy.CustomTrustStore.Add(ca)```
will not work, as it still needs to verify that the certificate, and it's easiest to do that using `chain.Build(cert)`.

tg123:
ok could you please also remove `var rootCert = chain.ChainElements[chain.ChainElements.Count - 1].Certificate;` since it is no longer used.

then /LTGM

KLazarov:
Done.

/LTGM

KLazarov:
Btw just read in the guide lines about squashing and commit message length: https://github.com/kubernetes/community/blob/master/contributors/guide/pull-requests.md#squashing

I will squash it into one commit tomorrow.

tg123 (commented):
/LTGM

tg123:
/

> Btw just read in the guide lines about squashing and commit message length: https://github.com/kubernetes/community/blob/master/contributors/guide/pull-requests.md#squashing
> 
> I will squash it into one commit tomorrow.

you do not have to, it is automatic

tg123 (approved):

PR #883, opened 2022-06-03
shortcut for iitem

### LINKED ISSUE #861: Make `await ListNamespacedServiceAsync` return enumerable
Currently, I have to 
```csharp
(await ListNamespacedServiceAsync(..)).Items.Where(..)
```
so I it would be nice 
```csharp
(await ListNamespacedServiceAsync(..)).Where(..)``
```

When I see `List` in method name, I always mistype `Where` as list kind of enumerable.

Thank you Sir

tg123:
fix #861 
thank @WeihanLi for the idea

### DIFF
--- a/src/KubernetesClient.Models/IItems.cs
+++ b/src/KubernetesClient.Models/IItems.cs
@@ -12,4 +12,12 @@ namespace k8s
         /// </summary>
         IList<T> Items { get; set; }
     }
+
+    public static class ItemsExt
+    {
+        public static IEnumerator<T> GetEnumerator<T>(this IItems<T> items)
+        {
+            return items.Items.GetEnumerator();
+        }
+    }
 }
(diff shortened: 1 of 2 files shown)


tg123:
linq is not available anyway :(

src/KubernetesClient.Models/IItems.cs
```diff
@@ -12,4 +12,12 @@ public interface IItems<T>
         /// </summary>
         IList<T> Items { get; set; }
     }
+
+    public static class ItemsExt
+    {
+        public static IEnumerator<T> GetEnumerator<T>(this IItems<T> items)
+        {
+            return items.Items.GetEnumerator();
```
xiaomi7732, line 20: Just curious, is it possible for `items.Items` to be null in any case? What is the expected behavior? Let it throw or gracefully quit the enumeration?

tg123: yup you are right, would throw npe
maybe make it empty would be better

the issue now is that linq is not compatible with this impl

WeihanLi:
> linq is not available anyway :(

It's ok from my side
image

tg123:
@brendanburns I believe this is ready for merge
any comments?

brendandburns:
/lgtm
/approve

Thanks

PR #944, opened 2022-07-11
 Console example of OpenTelemetry tracing

m3nax:
Minimal example of integration between the Opentelemetry (https://www.nuget.org/packages/OpenTelemetry) library and the httpclient used in the Kubernetes class

As suggested in the comment: https://github.com/kubernetes-client/csharp/discussions/898#discussioncomment-3013581


brendandburns:
/lgtm
/approve

Thanks for the PR!

PR #957, opened 2022-07-19
fix(StringQuotingEmitter): add missing patterns for double quoting

### LINKED ISSUE #953: Yaml Serialization is missing some tokens that need to be double quoted
Hello,

I found out that there are some special characters that are not being considered to be double quoted when serializing a Kubernetes object to yaml.
In the **StringQuotingEmitter** class, there is this regex:

`private static readonly Regex QuotedRegex =
            new Regex(@"^(\~|null|true|false|-?(0|[0-9]*)(\.[0-9]*)?([eE][-+]?[0-9]+)?)?$");`

This is not taking into consideration, at least, the following values "True", "TRUE", "False", "FALSE", "Y", "N", "y", "n", "Null" and "NULL".

vitorpmoreira:
Fixes #953  
**Added the missing patterns to the StringQuotingEmmiter** class to ensure that these values are double quoted when serializing to yaml.
Added a complementary Unit Test.

### DIFF
--- a/src/KubernetesClient.Models/StringQuotingEmitter.cs
+++ b/src/KubernetesClient.Models/StringQuotingEmitter.cs
@@ -8,7 +8,7 @@ namespace k8s
     // adapted from https://github.com/cloudbase/powershell-yaml/blob/master/powershell-yaml.psm1
     public class StringQuotingEmitter : ChainedEventEmitter
     {
-        // Patterns from https://yaml.org/spec/1.2/spec.html#id2804356
+        // Patterns from https://yaml.org/spec/1.2/spec.html#id2804356 and https://yaml.org/type/bool.html (spec v1.1)
         private static readonly Regex QuotedRegex =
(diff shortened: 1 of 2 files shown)


tg123:
should be https://yaml.org/type/bool.html right

vitorpmoreira:
> 

yes, nice one! And the null is also missing the "~" as in https://yaml.org/type/null.html

src/KubernetesClient.Models/StringQuotingEmitter.cs
```diff
@@ -10,7 +10,7 @@ public class StringQuotingEmitter : ChainedEventEmitter
     {
         // Patterns from https://yaml.org/spec/1.2/spec.html#id2804356
```
tg123, line 11: could you please also add `https://yaml.org/type/bool.html`

seems no `on/off` in spec 1.2

vitorpmoreira: sure! by any chance, is there any documentation regarding the spec versions that are supported by the k8s yaml parser? shouldn't they be the same for all client libraries + kubectl itself?

tg123 (approved):
/LGTM

Issue #949, opened 2022-07-13
`tar` process hangs when reading `stdin` in `NamespacedPodExecAsync`


lazarkrstic:
I am trying to implement `cp` command by starting tar process - the same principe as in `kubectl`.

Here is th code. 
````C#
        public async Task<int> CopyFileToPodAsync(string name, string @namespace, string container, string sourceFilePath, string destinationFilePath, CancellationToken cancellationToken = default(CancellationToken))
        {
            // All other parameters are being validated by MuxedStreamNamespacedPodExecAsync called by NamespacedPodExecAsync
            ValidatePathParameters(sourceFilePath, destinationFilePath);

            // The callback which processes the standard input, standard output and standard error of exec method
            var handler = new ExecAsyncCallback(async (stdIn, stdOut, stdError) =>
            {
                var fileInfo = new FileInfo(destinationFilePath);
                try
                {
                    using (var outputStream = new MemoryStream())
                    {
                        using (var inputFileStream = File.OpenRead(sourceFilePath))
                        using (var gZipOutputStream = new GZipOutputStream(outputStream))
                        using (var tarOutputStream = new TarOutputStream(gZipOutputStream))
                        {
                            // To avoid gZipOutputStream to close the memoryStream
                            gZipOutputStream.IsStreamOwner = false;

                            var fileSize = inputFileStream.Length;
                            var entry = TarEntry.CreateTarEntry(fileInfo.Name);
                            entry.Size = fileSize;

                            tarOutputStream.PutNextEntry(entry);
                            inputFileStream.CopyTo(tarOutputStream);
                            tarOutputStream.CloseEntry();
                        }
                        outputStream.Position = 0;

                        await outputStream.CopyToAsync(stdIn);
                        await outputStream.FlushAsync();
                    }
                }
                catch (Exception ex)
                {
                    throw new IOException($"Copy command failed: {ex.Message}");
                }

                using (System.IO.StreamReader streamReader = new System.IO.StreamReader(stdError))
                {
                    while (streamReader.EndOfStream == false)
                    {
                        string error = await streamReader.ReadToEndAsync();
                        throw new IOException($"Copy command failed: {error}");

                    }
                }


            });

            var destinationFolder = GetFolderName(destinationFilePath);

            return await Kubernetes.NamespacedPodExecAsync(
                name,
                @namespace,
                container,
                new string[] { "sh", "-c", $"tar -xzmf - -C {destinationFolder}" },
                false,
                handler,
[...]

PR #961, opened 2022-07-20
Add copy API

WeihanLi:
Trying to implement the copy file API

tested and works on my local environment, while the copy file to pod implement may be bad


WeihanLi:
> #949 
> 
> I prefer to add it to Kubectl API

That would be great

PR #962, opened 2022-07-21
Copy local file to pod using tar

lazarkrstic:


examples/cp/demo.txt
```diff
@@ -0,0 +1,4 @@
+Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.
```
tg123, line 1: could you please remove this file?

tg123 (commented):
thanks

I will make inner websocket chunked

tg123 (approved):
/LGTM

lazarkrstic:
Can you remove label `invalid-commit-message` , I amended my privious git message

tg123:
do not put `#` in message, just remove it and push -f

tg123 (approved):
/LGTM

cody-malcolm:
@lazarkrstic Thanks for doing this. When I use this example, the file is copied to the pod successfully, but then the thread blocks on the error checking starting at line 96...specifically being blocked by `streamReader.EndOfStream`. I have tried various other ways to perform the same check (such as ReadLine, Read, or stdError.Length) and they all block. If I remove the error check entirely, it will block in Kubernetes.Exec.cs at `await errorReader.ReadToEndAsync().ConfigureAwait(false);`. Do you know why this might be happening or what can be done to fix it? I'm out of ideas to fix it. Thanks in advance.

tg123:
@cody-malcolm what is your base image?

only ubuntu' tar passed the test


see comment here
https://github.com/kubernetes-client/csharp/blob/fd3931c994f6ce6fab60253acf72edb7d750ff30/tests/E2E.Tests/MinikubeTests.cs#L654

cody-malcolm:
@tg123 We are using node:16.15.0-alpine base image. Based on Alpine Linux.

PR #990, opened 2022-08-30
Updated lib dependency and bump version of System.Diagnostics.Diagnos…

m3nax:
…ticSource to 6.0.0


brendandburns:
/approve

@tg123 for final lgtm to check this to make sure there's nothing unexpected caused by this change. (he knows dotnet build better than I do)

tg123 (approved):
/LGTM

PR #994, opened 2022-09-03
Enhance AbstractKubernetes SendRequest extensibility

wimoy:
We want to use a Polly to handle retries in the KubernetesClient send request code path. HttpRequestMessage and HttpContent objects can only be passed in once to an HttpClient.SendAsync call. On each retry, a new HttpRequestMessage needs to be constructed; the previous HttpRequestMessage cannot be reused. We want the send request extensibility point in KubernetesClient to provide the underlying data used to construct the HttpRequestMessage instead of providing the HttpRequestMessage itself.

This PR generalizes the existing send request extensibility without making the contract more complicated.

### DIFF
--- a/src/KubernetesClient.Basic/AbstractKubernetes.cs
+++ b/src/KubernetesClient.Basic/AbstractKubernetes.cs
@@ -54,19 +54,6 @@ public abstract partial class AbstractKubernetes
         }
     }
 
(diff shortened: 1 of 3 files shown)


tg123:
btw could you please add an example or document how to mock/test Kubernetes client?

better to illustrate why the change is needed
and 
help others in similar situations

brendandburns:
/approve

Sorry I missed this PR...

PR #1002, opened 2022-09-12
fix: reading JsonPropertyAttributes for external types

IvanJosipovic:
This PR fixes mapping JsonPropertyAttributes to YamlDotNet Property Names for external types.

Currently, "WithOverridesFromJsonPropertyAttributes" only maps JsonPropertyAttributes in the KubernetesClient.Models assembly.

PR #1003, opened 2022-09-12
System.Text.Json support for YamlDotNet

IvanJosipovic:
This PR integrates a new library [YamlDotNet.System.Text.Json](https://github.com/IvanJosipovic/YamlDotNet.System.Text.Json) which allows YamlDotNet to de/serialize System.Text.Json objects and handles its attributes.

- fix: reading JsonPropertyAttributes for external types
  - Currently KubernetesYaml only reads JsonProperyAttributes for types in the [KubernetesClient.Models Assembly](https://github.com/kubernetes-client/csharp/blob/13cc644293afd587556c7b0f07828c940f202fbb/src/KubernetesClient.Models/KubernetesYaml.cs#L256)
- feat: support yaml de/serialization of JsonElement/JsonDocument/JsonObject etc
  - Allows the client to support Generic objects which are needed to support CRDs with XKubernetesPreserveUnknownFields
  ```csharp
  public class GenericObject : IKubernetesObject<V1ObjectMeta>
  {
      public string ApiVersion { get; set; }
      public string Kind { get; set; }
      public V1ObjectMeta Metadata { get; set; }
      public JsonObject Spec { get; set; }
  }
  ```

### DIFF
--- a/src/KubernetesClient.Models/KubernetesClient.Models.csproj
+++ b/src/KubernetesClient.Models/KubernetesClient.Models.csproj
@@ -15,5 +15,6 @@
     <PackageReference Include="AutoMapper" Version="10.1.1" />
     <PackageReference Include="Fractions" Version="7.1.0" />
     <PackageReference Include="YamlDotNet" Version="12.0.0" />
+    <PackageReference Include="YamlDotNet.System.Text.Json" Version="1.0.0" />
   </ItemGroup>
 </Project>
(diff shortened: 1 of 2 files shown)


tg123:
personally prefer #1002
the lib is cool, but wish to keep the dependencies as less as it could

IvanJosipovic:
Thanks @tg123, I assumed as much :)

As a workaround, would you be open to me creating a PR that allow us to customize the Json/Yaml serializer/deserializer settings? This would allow me to inject my own code into the serializers without having to duplicate code or use reflection.

PR #1010, opened 2022-09-14
Fix some compilation warnings

m3nax:
Goal:
- Fix some of annoying warnings generated during CI jobs

Side goal:
- Simplify the debugging of the CI job
- Reduce CI logs

Example: https://github.com/kubernetes-client/csharp/actions/runs/3055244950/jobs/4928083695#step:7:52


m3nax:
@tg123 More than half or the warning has disappeared, what do you think of the remaining ones? can we solve them?

brendandburns:
/lgtm
/approve

Thank you for taking on this cleanup work!!

Issue #1011, opened 2022-09-14
[Code style] Force code writing style verification during pull request checks


m3nax:
Force code writing style verification during pull request checks.

Objective: 
- increase the uniformity of the writing style of the code

tg123:
`kubernetes-client.ruleset`

m3nax:
Can we make more restrictive rule?
Making your code smoother and more consistent makes editing easier and making it easier to contribute....

m3nax:
Witch Action check code style?

brendandburns:
@m3nax we used to have it, but it looks like @tg123 may have accidentally removed it in https://github.com/kubernetes-client/csharp/commit/250a45741e04662a408af508ec87dfb66eed1bdb

@tg123 was this on purpose? (and my mistake for missing it at the time when I reviewed it...)

tg123:
not a mistake

the `dotnet fmt` conflicts with[ `stylecop` ](https://github.com/kubernetes-client/csharp/blob/master/kubernetes-client.ruleset) check

cannot pass them together, so I disable the `dotnet fmt`

m3nax:
ruleset is deprecated: https://learn.microsoft.com/en-us/visualstudio/code-quality/use-roslyn-analyzers?view=vs-2019#convert-an-existing-ruleset-file-to-editorconfig-file

Can i covert it in a .editorconfig?

PR #1023, opened 2022-09-20
Fix duplication errors in ruleset

m3nax:
Due to the duplication of the SA1500 key, the ruleset cannot be opened from visual studio.

This pull fix the problem.

Related https://github.com/kubernetes-client/csharp/issues/1011


tg123 (approved):
LGTM

Issue #846, opened 2022-04-30
merge k8s.Util.Common.Generic and GenericClient


tg123:
there are 2 impls and should merge them into 1

koepalex:
I comment on one pull request, but maybe I should have better posted here: https://github.com/kubernetes-client/csharp/pull/1042#issuecomment-1263282400

PR #1042, opened 2022-09-30
remove unused util code

tg123:
- remove duplicated generic code which impled 
- remove unfinished cache/informer code (still missing informer reflector ...)


tg123:
#846

koepalex:
We used in our code the `GenericKubernetesApi` to watch kubernetes resources (namespaces scoped and global), this class was removed from nuget `9.0.x`

```csharp
// constructor
_genericClient = new GenericClient(_kubernetesClient, GROUP, VERSION, PLURAL);
_genericKubernetesApi = new GenericKubernetesApi(GROUP, VERSION, PLURAL, _kubernetesClient);
// later
var watcher = _genericKubernetesApi.Watch(NAMESPACE, onEvent, onError, onClosed, cancellationToken);
```

We are using this approach to be able to reconnect when the connection is lost (onClosed) and for the error logging.
With removing of the API, I think we need to change to, the way shown in [Watch Example](https://github.com/kubernetes-client/csharp/blob/master/examples/watch/Program.cs). 

But how can we deal with _reconnect_ and _error logging_ in that case?

tg123:
thanks for reporting, will add `watch` as requested

brendandburns:
@tg123 will watch be added to this PR? Or in a follow up?

tg123:
> @tg123 will watch be added to this PR? Or in a follow up?

lets do it in followup

PR #1051, opened 2022-10-07
Use consistent net6.0 TargetFramework

stan-sz:
This is the only project that outstands from the rest in this repo.


tg123 (approved):
/LGTM

PR #1061, opened 2022-10-21
Fix strict yaml parsing

stan-sz:
Strict parsing was missing a positive test case that would ensure it's working as expected on a well-formed yaml file.


tg123 (approved):
/LGTM

PR #1085, opened 2022-11-09
add net7 remove net5

tg123:
net7 released today
drop net5 support
bump ver to 9.1


stan-sz:
Please update the version compatibility matrix at https://github.com/kubernetes-client/csharp/blob/master/README.md#L148

Issue #1136, opened 2022-12-14
Update System.IO.Abstraction to version >= 18.0.0


anhaehne:
**Describe the bug**
After 18.0.0 all types of `System.IO.Abstraction` were moved to the package `Testable.System.IO.Abstractions` which breaks all dependencies on prior versions.
To be able to update `System.IO.Abstraction` in my project to a version >= 18.0.0 this package has update first.

It is recommended to switch to the packages `Testable.System.IO.Abstraction` and `Testable.System.IO.Abstraction.Wrappers`.

Since this is a library issue not related to kubernetes, i will leave out the rest of the template.

For all having the same issue there exists a workaround by creating an aliased package reference:

csproj:
``` xml
<ItemGroup>
   <PackageReference Include="System.IO.Abstractions" Version="17.2.3">
       <Aliases>LegacySystemIOAbstractions</Aliases>
   </PackageReference>
</ItemGroup>
```
startup.cs:
``` csharp
extern alias LegacySystemIOAbstractions;
using System.IO.Abstractions;

public class Startup : RESTworld.AspNetCore.StartupBase
{
    public void ConfigureServices(IServiceCollection services)
    {
        services.AddScoped<IFileSystem, FileSystem>();
        services.AddScoped<LegacySystemIOAbstractions.System.IO.Abstractions.IFileSystem, LegacySystemIOAbstractions.System.IO.Abstractions.FileSystem>();
    }
}
```

tg123:
i was thinking removing as many dependencies as we could.
this is in the backlog but i failed to do so many times.

allow me some time to remove it

PR #1157, opened 2022-12-28
remove system io abstraction

tg123:
`System.IO.Abstraction` is for test only, but it introduces new dependency to the sdk users.

make it test only

#1136


brendandburns:
This looks good to me, but there is a merge conflict.

PR #1191, opened 2023-02-01
Dispose certificates in `Kubernetes.Dispose()`

### LINKED ISSUE #1189: Got "There is not enough space on the disk" error when testing the Kubernetes client in a Windows based CI pipeline
**Describe the bug**
We have a Windows based CI pipeline to test deploying K8s resource to AKS using the Kubernetes client. The pipeline is invoked many times a day and has been running for months but now we are getting "There is not enough space on the disk" error. Below is the call stack:

```
   at Internal.Cryptography.Pal.CertificatePal.FilterPFXStore(ReadOnlySpan`1 rawData, SafePasswordHandle password, PfxCertStoreFlags pfxCertStoreFlags)
   at Internal.Cryptography.Pal.CertificatePal.FromBlobOrFile(ReadOnlySpan`1 rawData, String fileName, SafePasswordHandle password, X509KeyStorageFlags keyStorageFlags)
   at System.Security.Cryptography.X509Certificates.X509Certificate..ctor(ReadOnlySpan`1 data)
   at System.Security.Cryptography.X509Certificates.X509Certificate2..ctor(Byte[] rawData)
   at k8s.CertUtils.GeneratePfx(KubernetesClientConfiguration config) in /_/src/KubernetesClient/CertUtils.cs:line 105
   at k8s.CertUtils.GetClientCert(KubernetesClientConfiguration config) in /_/src/KubernetesClient/CertUtils.cs:line 203
   at k8s.Kubernetes.InitializeFromConfig(KubernetesClientConfiguration config) in /_/src/KubernetesClient/Kubernetes.ConfigInit.cs:line 90
   at k8s.Kubernetes..ctor(KubernetesClientConfiguration config, DelegatingHandler[] handlers) in /_/src/KubernetesClient/Kubernetes.ConfigInit.cs:line 28
```

After checking the code, it seems to me that the root cause is that each `Kubernetes()` constructor call creates a one-time use `X509Certificate2` file on the disk that is never removed:

https://github.com/kubernetes-client/csharp/blob/e6f5f0a9ff782725b3b62129551bd756bcced3e0/src/KubernetesClient/CertUtils.cs#L94-L107

Should line 94 above be `using var cert = X509Certificate2.CreateFromPem(certData, keyData);`?

e.g. `9.0.38`

**Server Kubernetes Version**
N/A

**Dotnet Runtime Version**
6.0

**To Reproduce**
Call `new Kubernetes(KubernetesClientConfiguration)` many times until the disk is full.

**Expected behavior**
The `X509Certificate2` files should be cleaned up.

**KubeConfig**
N/A

 - OS: Windows
 - Environment [e.g. container]
 - Cloud [e.g. Azure]

**Additional context**

shenglol:
The `Kubernetes` constructor creates client and SSL CA certs via `new X509Certificate2(...)` if the kubeconfig contains cert path/data. As a result, on Windows, some one-time use files may be created for each `new Kuberentes(...)` call (see [The most dangerous constructor in .NET](https://snede.net/the-most-dangerous-constructor-in-net/#:~:text=You%20should%20never%20instantiate%20a,files%2C%20that%20are%20never%20removed.)).

According to the [SO answer](https://stackoverflow.com/questions/56337434/should-i-dispose-of-x509certificate2), the files will be removed by GC and calling `Dispose` is not required. However, since GC is not guaranteed to invoke all finalizers, we cannot rely on it to do the cleanup. I ran a local test and found that the files are not always cleaned up automatically. Thus, calling `X509Certificate2.Dispose()` is a must. This is also mentioned in the [MS doc](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.x509certificate2?view=net-7.0#remarks).

The PR updates `Kubernetes.Dispose()` to dispose certificates.

Closes #1189.

### DIFF
--- a/src/KubernetesClient/Kubernetes.WebSocket.cs
+++ b/src/KubernetesClient/Kubernetes.WebSocket.cs
@@ -239,11 +239,6 @@ namespace k8s
             }
 
             // Set Credentials
-            if (this.ClientCert != null)
-            {
-                webSocketBuilder.AddClientCertificate(this.ClientCert);
-            }
-
             if (this.HttpClientHandler != null)
             {
 #if NET5_0_OR_GREATER
--- a/src/KubernetesClient/Kubernetes.ConfigInit.cs
+++ b/src/KubernetesClient/Kubernetes.ConfigInit.cs
@@ -90,27 +90,27 @@ namespace k8s
             // set credentails for the kubernetes client
             SetCredentials(config);
 
-            var clientCert = CertUtils.GetClientCert(config);
-            if (clientCert != null)
+            ClientCert = CertUtils.GetClientCert(config);
+            if (ClientCert != null)
             {
 #if NET5_0_OR_GREATER
-                HttpClientHandler.SslOptions.ClientCertificates.Add(clientCert);
+                HttpClientHandler.SslOptions.ClientCertificates.Add(ClientCert);
 
                 // TODO this is workaround for net7.0, remove it when the issue is fixed
                 // seems the client certificate is cached and cannot be updated
                 HttpClientHandler.SslOptions.LocalCertificateSelectionCallback = (sender, targetHost, localCertificates, remoteCertificate, acceptableIssuers) =>
                 {
-                    return clientCert;
+                    return ClientCert;
                 };
 #else
-                HttpClientHandler.ClientCertificates.Add(clientCert);
+                HttpClientHandler.ClientCertificates.Add(ClientCert);
 #endif
             }
         }
 
         private X509Certificate2Collection CaCerts { get; }
 
-        private X509Certificate2 ClientCert { get; }
+        private X509Certificate2 ClientCert { get; set; }
 
         private bool SkipTlsVerify { get; }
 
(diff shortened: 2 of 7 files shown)


src/KubernetesClient/Kubernetes.WebSocket.cs
```diff
@@ -239,11 +239,6 @@ protected async Task<WebSocket> StreamConnectAsync(Uri uri, string webSocketSubP
             }
 
             // Set Credentials
-            if (this.ClientCert != null)
```
brendandburns, line 242: Why are you deleting this?

brendandburns:
Generally looks good, one comment.

src/KubernetesClient/Kubernetes.WebSocket.cs
shenglol: This is dead code. `this.ClientCert` is never assigned and is always null before the change. I did a git blame and it seems it's a leftover from a previous refactoring. The logic is covered by the code blow.

tg123 (approved):
/LGTM

PR #1208, opened 2023-02-17
Use camelCase policy when serializing enums

luiscantero:
The KubernetesJson class currently uses CamelCase as PropertyNamingPolicy. For enums it uses JsonStringEnumConverter, however, without the CamelCase policy. This results in invalid serialization (e.g. PascalCase).

This PR fixes this issue.


tg123:
any example which model is using pascal case?

luiscantero:
I'm using `GenericClient.CreateNamespacedAsync` to pass an object. The object contains a property that is an enum specified using PascalCase, e.g.:
```
animal.Kind = Animal.Dog;
gc.CreateNamespacedAsync(animal);
```

This causes an exception in our application as it expects `dog` (camelCase) but gets `Dog`. We cannot use our own converter/attribute as the class KubernetesJson specifies its own JsonSerializerOptions, which overrides our custom converter settings.

tg123:
not sure if it would break something, please allow me some time to do some research.

luiscantero:
Will prepare new PR

PR #1211, opened 2023-02-20
Proposal: user defined model version explicit converter

tg123:
# The Problem

The k8s models now have `explicit converter` for converting model objects in different versions, e.g. V1 <-> V1beta1
Typical usage is `V1X v1obj = (V1X)v1beta1obj`

The converter is based on AutoMapper and introduces the dependency to AutoMapper which stopped supporting `netstandard2.0`. Removing the dependency would also make the sdk clean and more secure. 

In additional, seems the feature is barely used.

# New Design

- a new `KubernetesClient.ModelConverter` lib for those who want to use version converters and move AutoMapper to this lib.
- users should call `ModelVersionConverter.Converter = AutoMapperModelVersionConverter.Instance;` to set Converter or it will throw an error
- ModelVersionConverter.Converter is an interface `IModelVersionConverter`. users could impl it and set `ModelVersionConverter.Converter` to their own converter


brendandburns:
@tg123 this seems fine to me. Do you think this is ready to merge?

tg123:
I am good. Please also add a nuget lib `KubernetesClient.ModelConverter` before merging.

brendandburns:
/lgtm
/approve

I _think_ it will "just work" because the `KubernetesClient` package already exists. We'll see. If not I will upload the artifact.

brendandburns:
*sigh* looks like it didn't, I will build/update this nuget manually.

brendandburns:
Ok, this is now pushed: https://www.nuget.org/packages/KubernetesClient.ModelConverter/10.1.4

PR #1242, opened 2023-03-22
Exec set CreateNoWindow to true

IvanJosipovic:
Currently, when exec is ran for kubelogin etc, a new window is spawned. This PR should hide that window.


tg123 (approved):
/LGTM

Issue #1252, opened 2023-03-30
Strong named assemblies


jmuukka:
Is it possible to get NuGet packages that have signed assemblies? Our system prevents usage of unsigned assemblies.

I signed the packages myself, but since they depends on each other, compiler says that some type is in Models assembly which could not be found. There is something wrong with this procedure, either technical or logical, and it's error prone.

It would be preferred that we would not need this manual process at all and could just use the signed dlls. The signing the dlls is quite quick thing to do. We would prefer to get the signed versions soon.

I actually could do the work, and make a PR, if you could provide the .snk file.

jmuukka:
Now I have the branch ready. It just needs your .snk and then I can update the public key in "InternalsVisibleTo".

tg123:
strong name is no longer supported. see https://github.com/kubernetes-client/csharp/pull/535

PR #1254, opened 2023-03-30
feat: Sign assemblies (#1252)

jmuukka:
Signed the assemblies with a .snk, which was created using a sn.exe tool. It was necessary to update InternalsVisibleTo with a public key.

There are companies which require signed assemblies. I am a subcontractor for one who is needing signed assemblies.


jmuukka:
I don't know which to select in CLA Authorization. It's not my own code, so I should select a corporate, but which corporate to enter there?

tg123:
strong name is no longer supported. see #535

see
https://learn.microsoft.com/en-us/dotnet/standard/assembly/strong-named#why-strong-name-your-assemblies

tg123:
feel free to sign cla with the email that you use in git commit for later PR
close as no longer supported

PR #1257, opened 2023-04-03
Add AddJsonOptions to KubernetesClientConfiguration and KubernetesJson

luiscantero:
* Add AddJsonOptions so that JsonSerializerOptions can be customized
* Add unit test for JSON serialization


src/KubernetesClient/KubernetesClientConfiguration.cs
```diff
@@ -108,5 +108,11 @@ public partial class KubernetesClientConfiguration
         /// Do not use http2 even it is available
         /// </summary>
         public bool DisableHttp2 { get; set; } = false;
+
+        /// <inheritdoc cref="KubernetesJson.AddJsonOptions(Action{JsonSerializerOptions})"/>
+        public void AddJsonOptions(Action<JsonSerializerOptions> configure)
+        {
+            KubernetesJson.AddJsonOptions(configure);
```
tg123, line 115: this is a global var
what about introducing a per kuberclient level json inst? default to KubernetesJson
will use per inst json serializer if `addjsonopt` called

here is some the place where kuberetesjson is being called
https://github.com/kubernetes-client/csharp/blob/aa84b453dca3e4af96137ec60ada5f6fcd6ed694/src/KubernetesClient/Kubernetes.cs#L129

luiscantero: Done

luiscantero:
Pushed fix for CodeQL issues

tg123 (approved):
/LGTM

Issue #621, opened 2021-05-16
SslCaCerts appears to be require even though kubectl does not need it


lucastheisen:
If i understand correctly (from other comments I think i have seen on this project), if it works in `kubectl` it should work here.  However, my KUBECONFIG does not have an explicit `certificate-authority` nor `certificate-authority-data` and `kubectl` commands still work fine.  However, when i use this library i still get:

```console
Unhandled exception. k8s.Exceptions.KubeConfigException: A CA must be set when SkipTlsVerify === false
   at k8s.Kubernetes.InitializeFromConfig(KubernetesClientConfiguration config)
   at k8s.Kubernetes.CreateHttpClient(DelegatingHandler[] handlers, KubernetesClientConfiguration config)
   at k8s.Kubernetes..ctor(KubernetesClientConfiguration config, DelegatingHandler[] handlers)
   at k8s.GenericClient..ctor(KubernetesClientConfiguration config, String group, String version, String plural)
```

This appears to be part of the [configuration initialization](https://github.com/kubernetes-client/csharp/blob/09f08d76590b3da06ea29beaa053f3b95564a380/src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs#L293-L306):

```csharp
            if (uri.Scheme == "https")
            {
                if (!string.IsNullOrEmpty(clusterDetails.ClusterEndpoint.CertificateAuthorityData))
                {
                    var data = clusterDetails.ClusterEndpoint.CertificateAuthorityData;
                    SslCaCerts = new X509Certificate2Collection(new X509Certificate2(Convert.FromBase64String(data)));
                }
                else if (!string.IsNullOrEmpty(clusterDetails.ClusterEndpoint.CertificateAuthority))
                {
                    SslCaCerts = new X509Certificate2Collection(new X509Certificate2(GetFullPath(
                        k8SConfig,
                        clusterDetails.ClusterEndpoint.CertificateAuthority)));
                }
            }
```

It seems like there should be an allowed _else_ condition that falls back to the OS level CA.  I can work around this with something like:

```csharp
            var k8sConfig = KubernetesClientConfiguration
                    .BuildConfigFromConfigFile(this.configuration.KubeConfigFile);

            // point to OS level ca-certificates file
            k8sConfig.SslCaCerts = new X509Certificate2Collection(
                new X509Certificate2("/etc/ssl/certs/ca-certificates.crt"));

            var client = new Kubernetes(k8sConfig);
```

[...]

PR #1261, opened 2023-04-11
Use the system certificate store if no certificates are specified.

marcusbooyah:
This is a fix for #621.

### DIFF
--- a/src/KubernetesClient/Kubernetes.ConfigInit.cs
+++ b/src/KubernetesClient/Kubernetes.ConfigInit.cs
@@ -72,28 +72,12 @@ namespace k8s
                 }
                 else
                 {
-                    if (CaCerts == null)
+                    if (CaCerts != null)
                     {
-                        var store = new X509Store(
-                            StoreName.CertificateAuthority,
-                            StoreLocation.CurrentUser);
 #if NET5_0_OR_GREATER
                         HttpClientHandler.SslOptions.RemoteCertificateValidationCallback =
 #else
                         HttpClientHandler.ServerCertificateCustomValidationCallback =
-#endif
-                            (sender, certificate, chain, sslPolicyErrors) =>
-                            {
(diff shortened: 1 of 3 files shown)


tg123:
i would say this is not included in `kubectl`

tg123:
do you mean `SkipTlsVerify=false` and let sdk use cert store if cacert == null

tg123:
> > do you mean `SkipTlsVerify=false` and let sdk use cert store if cacert == null
> 
> That's what this PR does

why not put your letsencrypt ca into kubeconfig

jefflill:
This PR seems like a bit of a no-brainer.  Having KubernetesClient be consistent with kubectl is a good idea and it would be nice if future developers didn't need to debug and then workaround things like this.

tg123:
totally agree the sdk should have same behavior as `kubectl`,
let me double check what is the default behavior if no ca is set in kubectl

and btw I prefer
```
 to not set the ServerCertificateCustomValidationCallback at all.
```

tg123:
Thanks, I double confirmed you are right, should set callback to null in order to use system default.

here is how `kubectl`Rset rootCA
https://github.com/kubernetes/client-go/blob/704cac462081473b521f15547302f001f78ef2e4/transport/transport.go#L98

```
// RootCAs defines the set of root certificate authorities
	// that clients use when verifying server certificates.
	// If RootCAs is nil, TLS uses the host's root CA set.
	RootCAs *[x509](https://pkg.go.dev/crypto/x509).[CertPool](https://pkg.go.dev/crypto/x509#CertPool)
```

tg123:
/LGTM

cc @brendanburns 
do you think we need to upgrade the ver for this?

brendandburns:
/lgtm
/approve

No, I don't think this is a breaking change, I think it is more of a bug-fix. (honestly I doubt anyone will notice as few people use well-known CAs for the kube apiservers)

Issue #1092, opened 2022-11-17
Calls to configure Kubernetes client hang (i.e. BuildConfigFromConfigFile) due to Verbose logging in kube config for AWS External Authentication


siparsons:
Describe the bug
Activating verbose logging of the AWS console causes the following line of code to hang:

https://github.com/kubernetes-client/csharp/blob/48b5ea6de09de8aa9bd2a54256efd2a252e6e151/src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs#L559

Kubernetes C# SDK Client Version
8.0.68

Server Kubernetes Version
e.g. 1.24.2

Dotnet Runtime Version
.Net core

To Reproduce
I have the following local Kube config setting (see section below). Note that this is allowing the user token to be retrieved from an external authentication provider - AWS Client Console in this case. 

When using AWS console it is sometimes necessary to turn on "Verbose Logging" of the console application to diagnose connection issues. This is done by including the additional "--debug" command line argument seen below. 

```
users:
- name: arn:aws:eks:eu-west-xxxxxxxxxxxxxxx:cluster/xxxxxxxxxxxxxxx
  user:
    exec:
      apiVersion: client.authentication.k8s.io/v1beta1
      args:
      - --region
      - eu-west-2
      - eks
      - get-token
      - --cluster-name
      - xxxxxxxxxxxxxxx
      - --debug
      command: aws
      env:
      - name: AWS_PROFILE
        value: xxxxxxxxxxxxxxx
      interactiveMode: IfAvailable
      provideClusterInfo: false
``` 

The effect of this is twofold.

1. The console/process output is verbose (as one might expect)
2. The console/process output includes far more than just the final JSON response relied on by the kubernetes-client library

The response result has no paging information
Expect  call to KubernetesClientConfiguration.BuildConfigFromConfigFile() to return a valid client, rather than hang.

Suggested fix. 

The immediate "hanging" problem would appear to be due to the verbose nature of the stdout when starting the process.

This is because the process response is being called Synchronously. 

Instad use the following approach to retrieve the full output from the process:

```
process.Start();
process.BeginOutputReadLine();
process.BeginErrorReadLine();

var stdout = "";
var stderr = "";

process.OutputDataReceived += (sender, args) =>
{
    stdout += args.Data;
};
process.ErrorDataReceived += (sender, args) =>
{
    stderr += args.Data;
};

process.WaitForExit();

```

[...]

PR #1267, opened 2023-04-17
fix: Exec hangs

### LINKED ISSUE #1266: Expose exec Standard error
**Describe the bug**
When connecting to a new Azure AKS instance, kubelogin exposes a URL that we must open. I think we have to somehow expose this in the client so the user can interact with the URL as needed. Kubelogin exposes this output as Standard Error.
image

10.1.19

**Server Kubernetes Version**
1.25.4

**Dotnet Runtime Version**
net6

**To Reproduce**
Install Azure CLI

az aks get-credentials --resource-group $ResourceGroup --name $Name --subscription $Subscription 

Load Kubeconfig with the client and attempt to list Pods etc.

**Expected behavior**
Client should expose standard error for exec and keep the process open and allow us view the output.

 - OS: Windows
 - Cloud: Azure

IvanJosipovic:
Fixes #1266

### DIFF
--- a/src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
+++ b/src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
@@ -30,6 +30,11 @@ namespace k8s
         internal static string KubeConfigEnvironmentVariable { get; set; } = "KUBECONFIG";
 
         /// <summary>
+        /// Exec process timeout in seconds
+        /// </summary>
+        public static int ExecTimeout { get; set; } = 10;
+
+        /// <summary>
         ///     Initializes a new instance of the <see cref="KubernetesClientConfiguration" /> from default locations
         ///     If the KUBECONFIG environment variable is set, then that will be used.
         ///     Next, it looks for a config file at <see cref="KubeConfigDefaultLocation"/>.
@@ -563,7 +568,7 @@ namespace k8s
             var buffer = new char[4096];
             var stdErrorTask = process.StandardError.ReadAsync(buffer, 0, buffer.Length); // Assumes process will continue
 
-            var result = Task.WaitAny(new Task[] { stdErrorTask, stdOutTask }, TimeSpan.FromSeconds(10));
+            var result = Task.WaitAny(new Task[] { stdErrorTask, stdOutTask }, TimeSpan.FromSeconds(ExecTimeout));
 
             if (result == -1)
             {
--- a/src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
+++ b/src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
@@ -549,84 +549,61 @@ namespace k8s
 
             var process = CreateRunnableExternalProcess(config);
 
+            var stdout = "";
+            var stderr = "";
+
+            process.OutputDataReceived += (sender, args) => stdout += args.Data;
+            process.ErrorDataReceived += (sender, args) => stderr += args.Data;
+
             try
             {
                 process.Start();
+                process.BeginOutputReadLine();
+                process.BeginErrorReadLine();
             }
             catch (Exception ex)
             {
                 throw new KubeConfigException($"external exec failed due to: {ex.Message}");
             }
 
-            var sb = new StringBuilder();
-            var buffer = new char[1024];
-            while (true)
+            // Wait for a maximum of 5 seconds, if a response takes longer probably something went wrong...
+            if (process.WaitForExitAsync().Wait(TimeSpan.FromSeconds(5)))
             {
-                var readTask = process.StandardError.ReadAsync(buffer, 0, buffer.Length);
-
-                if (readTask.Wait(TimeSpan.FromSeconds(2)))
+                try
                 {
-                    var bytesRead = readTask.Result;
-                    if (bytesRead == 0)
+                    var responseObject = KubernetesJson.Deserialize<ExecCredentialResponse>(stdout);
+                    if (responseObject == null || responseObject.ApiVersion != config.ApiVersion)
                     {
-                        break; // end of stream reached
+                        throw new KubeConfigException(
+                            $"external exec failed because api version {responseObject.ApiVersion} does not match {config.ApiVersion}");
                     }
 
-                    sb.Append(buffer, 0, bytesRead);
+                    if (responseObject.Status.IsValid())
+                    {
+                        return responseObject;
+                    }
+                    else
+                    {
+                        throw new KubeConfigException($"external exec failed missing token or clientCertificateData field in plugin output");
+                    }
                 }
-                else
+                catch (JsonException ex)
                 {
-                    // timeout occurred
-                    break;
+                    throw new KubeConfigException($"external exec failed due to failed deserialization process: {ex}");
                 }
-            }
-
-            var stderr = sb.ToString();
-
-            if (!string.IsNullOrWhiteSpace(stderr))
-            {
-                throw new KubeConfigException($"external exec failed due to: {stderr}");
-            }
-
-            // Wait for a maximum of 5 seconds, if a response takes longer probably something went wrong...
-            var stdOutTask = process.StandardOutput.ReadToEndAsync();
-
-            string stdOut;
-
-            if (stdOutTask.Wait(TimeSpan.FromSeconds(5)))
-            {
-                stdOut = stdOutTask.Result;
+                catch (Exception ex)
+                {
+                    throw new KubeConfigException($"external exec failed due to uncaught exception: {ex}");
+                }
             }
             else
             {
+                if (!string.IsNullOrWhiteSpace(stderr))
+                {
+                    throw new KubeConfigException($"external exec failed due to: {stderr}");
+                }
+
                 throw new KubeConfigException("external exec failed due to timeout");
-            }
-
-            try
-            {
-                var responseObject = KubernetesJson.Deserialize<ExecCredentialResponse>(stdOut);
-                if (responseObject == null || responseObject.ApiVersion != config.ApiVersion)
-                {
-                    throw new KubeConfigException(
-                        $"external exec failed because api version {responseObject.ApiVersion} does not match {config.ApiVersion}");
-                }
-
-                if (responseObject.Status.IsValid())
-                {
-                    return responseObject;
-                }
-                else
-                {
-                    throw new KubeConfigException($"external exec failed missing token or clientCertificateData field in plugin output");
-                }
-            }
-            catch (JsonException ex)
-            {
-                throw new KubeConfigException($"external exec failed due to failed deserialization process: {ex}");
-            }
-            catch (Exception ex)
-            {
-                throw new KubeConfigException($"external exec failed due to uncaught exception: {ex}");
             }
         }
 
--- a/src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
+++ b/src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
@@ -6,6 +6,7 @@ using System.IO;
 using System.Net;
 using System.Runtime.InteropServices;
 using System.Security.Cryptography.X509Certificates;
+using System.Text;
 using System.Threading.Tasks;
 
 namespace k8s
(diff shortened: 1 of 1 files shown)


src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
```diff
@@ -557,13 +558,38 @@ public static ExecCredentialResponse ExecuteExternalCommand(ExternalExecution co
...
-            var stdout = process.StandardOutput.ReadToEnd();
-            var stderr = process.StandardError.ReadToEnd();
-            if (string.IsNullOrWhiteSpace(stderr) == false)
+            var sb = new StringBuilder();
+            var buffer = new char[1024];
+            while (true)
+            {
+                var readTask = process.StandardError.ReadAsync(buffer, 0, buffer.Length);
+
+                if (readTask.Wait(TimeSpan.FromSeconds(2)))
```
tg123, line 567: exec will still hang if the process is consistently sending output, right?

src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
```diff
@@ -557,13 +558,38 @@ public static ExecCredentialResponse ExecuteExternalCommand(ExternalExecution co
...
+
+            if (!string.IsNullOrWhiteSpace(stderr))
             {
                 throw new KubeConfigException($"external exec failed due to: {stderr}");
             }
 
+            var stdout = process.StandardOutput.ReadToEnd();
+
             // Wait for a maximum of 5 seconds, if a response takes longer probably something went wrong...
             process.WaitForExit(5);
```
tg123, line 594: maybe it is to move wait output to a task and put a overall timeout here

src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
IvanJosipovic: That is correct, if the external application keeps sending errors. This will continue looping until it stops for two seconds.

My only experience has been with Kubelogin. I'm not sure how other clients behave here.

tg123: link #1092

src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
IvanJosipovic: I added a 5s timeout to the StandardOutput.

src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
IvanJosipovic: @tg123, thanks for linking that issue. I used the Event based option from that Issue and updated the PR.

Now, the process will timeout after 5 seconds no matter what, then we check if stderror has a value, if it does we will throw the response as an exception otherwise we throw a timeout exception.

The only gap remaining is the process will stick around on Errors indefinitely. We might need to add some more code to store the processes and kill them after 5 minutes or something. Let me know if you want me to code this in.

src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
```diff
@@ -548,24 +548,33 @@ public static ExecCredentialResponse ExecuteExternalCommand(ExternalExecution co
...
             catch (Exception ex)
             {
                 throw new KubeConfigException($"external exec failed due to: {ex.Message}");
             }
 
-            var stdout = process.StandardOutput.ReadToEnd();
-            var stderr = process.StandardError.ReadToEnd();
-            if (string.IsNullOrWhiteSpace(stderr) == false)
+            // Wait for a maximum of 5 seconds, if a response takes longer probably something went wrong...
+            if (!process.WaitForExit(5000) && string.IsNullOrWhiteSpace(stderr))
```
brendandburns, line 569: We should perhaps wait even longer if the terminal is interactive. Something like this:

https://stackoverflow.com/questions/45393001/determine-if-net-core-console-application-is-running-in-user-interactive-mode

It feels like 5s is pretty fast to get to a web browser and get logged in.

brendandburns:
This looks reasonable to me. Thanks for the PR!

src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
IvanJosipovic: When Kubelogin asks for an interactive prompt, it will return an error. However, the process will not be killed.

IvanJosipovic:
I'm not sure why this test (k8s.Tests.AuthTests.ExternalToken) is failing, its working locally for me.

image

src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
```diff
@@ -557,19 +558,51 @@ public static ExecCredentialResponse ExecuteExternalCommand(ExternalExecution co
...
+            var stderrBuilder = new StringBuilder();
+
+            var stdOutTask = process.StandardOutput.ReadToEndAsync(); // Assumes process exits
+
+            var buffer = new char[1024];
+
+            var stopwatch = Stopwatch.StartNew();
+
+            // Wait for a maximum of 5 seconds, if a response takes longer probably something went wrong...
+            while (!stdOutTask.IsCompleted && stopwatch.Elapsed < TimeSpan.FromSeconds(5))
```
tg123, line 570: you can do the pattern like below without have to loop and sleep, which will always cause thread to sleep 500ms

```
t = task.create { readstdout and stderr }
task.waitany(delay(timeout), t)
```

tg123, line 570: could you please move timeout to config

IvanJosipovic: I tried that logic last night, however, it introduced another issue and it still needed a loop. Let me bring it back and see if I might have missed something as it was rater late.

tg123: add ct to readasync

IvanJosipovic: @tg123, any suggestion where to place it?

src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
```diff
@@ -557,19 +558,32 @@ public static ExecCredentialResponse ExecuteExternalCommand(ExternalExecution co
...
+                if (stderrBuilder.Length > 0)
+                {
+                    throw new KubeConfigException($"external exec failed due to: {stderrBuilder}");
+                }
+            }
 
             try
             {
-                var responseObject = KubernetesJson.Deserialize<ExecCredentialResponse>(stdout);
+                var responseObject = KubernetesJson.Deserialize<ExecCredentialResponse>(stdOutTask.Result);
```
tg123, line 586: `.Result` vs `getawait()`

see https://stackoverflow.com/questions/17284517/is-task-result-the-same-as-getawaiter-getresult

for better explanation

src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
```diff
@@ -557,19 +558,32 @@ public static ExecCredentialResponse ExecuteExternalCommand(ExternalExecution co
...
+                throw new KubeConfigException("external exec failed due to timeout");
             }
+            else if (result == 0)
+            {
+                var stderrBuilder = new StringBuilder();
+                stderrBuilder.Append(buffer, 0, stdErrorTask.Result);
 
-            // Wait for a maximum of 5 seconds, if a response takes longer probably something went wrong...
-            process.WaitForExit(5);
+                if (stderrBuilder.Length > 0)
```
tg123, line 578: here my suggestion is to ignore stderr
throw only when cannot deserialize the stdout

the exec may write some debug log to stderr

src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
```diff
@@ -557,19 +558,32 @@ public static ExecCredentialResponse ExecuteExternalCommand(ExternalExecution co
...
-            var stdout = process.StandardOutput.ReadToEnd();
-            var stderr = process.StandardError.ReadToEnd();
-            if (string.IsNullOrWhiteSpace(stderr) == false)
+
+            var stdOutTask = process.StandardOutput.ReadToEndAsync(); // Assumes process exits
+
+            var buffer = new char[4096];
+            var stdErrorTask = process.StandardError.ReadAsync(buffer, 0, buffer.Length); // Assumes process will continue
+
+            var result = Task.WaitAny(new Task[] { stdErrorTask, stdOutTask }, TimeSpan.FromSeconds(10));
```
tg123, line 567: 10 sec to config?

tg123: what about `KubernetesClientConfiguration`

IvanJosipovic: I added a static property to KubernetesClientConfiguration.

src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
```diff
@@ -28,6 +29,11 @@ public partial class KubernetesClientConfiguration
         // For testing
         internal static string KubeConfigEnvironmentVariable { get; set; } = "KUBECONFIG";
 
+        /// <summary>
+        /// Exec process timeout in seconds
+        /// </summary>
+        public static int ExecTimeout { get; set; } = 10;
```
tg123, line 35: any reason why this is not per inst var?

IvanJosipovic: @tg123 currently the ExecuteExternalCommand command which uses it is Static.

src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
IvanJosipovic: I think logging of non errors to stderr is a bad practice.

My testing of kubelogin shows the following workflow, and I modeled the current logic to follow.

kubelogin drawio


Also looking at the google auth plugin appears to be similar, errors are only printed on error conditions.
https://github.com/kubernetes/cloud-provider-gcp/blob/master/cmd/gke-gcloud-auth-plugin/cred.go#L145

src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
tg123: oh i see, what about make ExecuteExternalCommand non static?
it is weird to this value static

IvanJosipovic: Its static as its used outside here, https://github.com/kubernetes-client/csharp/blob/e42290d6f7c7fbb515cca119e8c3f7507bd00ce2/src/KubernetesClient/Authentication/ExecTokenProvider.cs#L46

src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
tg123: just double checked client-go

https://github.com/kubernetes/client-go/blob/6c514ab3ffc5f500fd373f09956780c0e3bbe028/plugin/pkg/client/auth/exec/exec.go#L450

the impl is to ignore and redirect stderr to os.stderr and deserialize the stdout directly

src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
tg123: oh i see, that is difficult to change, lets keep it `static`

ps i do not think anyone is going to modify it

src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
IvanJosipovic: I'm not sure how to ignore the stderr as we need its data, and also block the call and wait for stdout. The current model works well with my testing.
See recording [Video.webm](https://user-images.githubusercontent.com/9521987/234685591-bc3a70de-0099-4355-bb9c-547c022ca3bb.webm)

Any suggestions?

tg123: ignore here means do not `if (stderrBuilder.Length > 0)`
just leave it there, only throw it if  cannot deserialize stdout

BTW this kubeui is so cool

IvanJosipovic:
I think this approach closely resembles the Go client.

src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
```diff
@@ -551,25 +561,25 @@ public static ExecCredentialResponse ExecuteExternalCommand(ExternalExecution co
...
-                throw new KubeConfigException($"external exec failed due to: {stderr}");
-            }
-
-            // Wait for a maximum of 5 seconds, if a response takes longer probably something went wrong...
-            process.WaitForExit(5);
-
             try
             {
-                var responseObject = KubernetesJson.Deserialize<ExecCredentialResponse>(stdout);
+                if (!process.WaitForExit((int)TimeSpan.FromSeconds(ExecTimeout).TotalMilliseconds))
```
tg123, line 577:
```suggestion
                if (!process.WaitForExit(TimeSpan.FromSeconds(ExecTimeout))
```

src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
```diff
@@ -28,6 +28,16 @@ public partial class KubernetesClientConfiguration
         // For testing
         internal static string KubeConfigEnvironmentVariable { get; set; } = "KUBECONFIG";
 
+        /// <summary>
+        ///     Exec process timeout in seconds
+        /// </summary>
+        public static int ExecTimeout { get; set; } = 120;
```
tg123, line 34: or make timespan here

tg123 (commented):
suggest use timespan instead of second 
then LGTM

tg123 (approved):
/LGTM

PR #1274, opened 2023-04-20
feat: Add a parameter to the GenericClient class that configures the …

### LINKED ISSUE #1273: GenericClient shouldn't dispose underlying client
**Describe the bug**
When the GenericClient is disposed, it also disposes the underlying Kubernetes client.

I think we should disable this functionality or make it opt in as the GenericClient is resource specific.

https://github.com/kubernetes-client/csharp/blob/729b10c7318793d1c455f7267a6b11d6aed8fed5/src/KubernetesClient/GenericClient.cs#L154

IvanJosipovic:
Fixes #1273

### DIFF
--- a/src/KubernetesClient/GenericClient.cs
+++ b/src/KubernetesClient/GenericClient.cs
@@ -11,6 +11,7 @@ namespace k8s
         private readonly string group;
         private readonly string version;
         private readonly string plural;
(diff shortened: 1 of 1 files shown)


tg123 (approved):
/LGTM

PR #1282, opened 2023-04-27
Support TLS Server Name overrides in kubeconfig

cb:
The client should support tls-server-name just like client-go and kubectl.  See https://github.com/kubernetes/kubernetes/pull/88769

### DIFF
--- a/src/KubernetesClient/Kubernetes.cs
+++ b/src/KubernetesClient/Kubernetes.cs
@@ -149,6 +149,11 @@ namespace k8s
                 await Credentials.ProcessHttpRequestAsync(httpRequest, cancellationToken).ConfigureAwait(false);
             }
 
+            if (!string.IsNullOrWhiteSpace(TlsServerName))
+            {
+                httpRequest.Headers.Host = TlsServerName;
+            }
+
             // Send Request
             cancellationToken.ThrowIfCancellationRequested();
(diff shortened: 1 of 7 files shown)


src/KubernetesClient/Kubernetes.cs
```diff
@@ -149,6 +149,11 @@ protected virtual async Task<HttpResponseMessage> SendRequestRaw(string requestC
                 await Credentials.ProcessHttpRequestAsync(httpRequest, cancellationToken).ConfigureAwait(false);
             }
 
+            if (!string.IsNullOrWhiteSpace(TlsServerName))
+            {
+                httpRequest.Headers.Host = TlsServerName;
```
tg123, line 154: is the header here enough?
no change to `tls check` func?

cb: That's it.  The Host header is used by the framework to set the proper SNI for TLS, and all other TLS checks honor the SNI extension.  You can test by adding a entry in /etc/hosts or %systemroot%\system32\drivers\etc\hosts for a hostname not in the api-server's SANs.  It should fail.  Then add a tls-server-name entry in the config for 'kubernetes' or any valid SAN in the api-server's cert and it should work.

tg123 (approved):

PR #1288, opened 2023-05-03
Fix failing nuget action

m3nax:
Fix cleanup of old nuget packages.

Last run: https://github.com/kubernetes-client/csharp/actions/runs/4848894956


tg123 (approved):
/LGTM

PR #1304, opened 2023-05-18
feat: enable multiple http2 connections

IvanJosipovic:
In my KubeUI project, I create hundreds of Informers (Lists + Watches), one for each resource type. After a period of time, the Informers would stop working and start throwing errors that wouldn't recover. See https://github.com/IvanJosipovic/KubeUI/issues/458. In my testing enabling EnableMultipleHttp2Connections solved this problem.

Docs for this property, https://learn.microsoft.com/en-us/dotnet/api/system.net.http.socketshttphandler.enablemultiplehttp2connections


tg123:
/LGTM

thanks

tg123 (approved):

PR #1332, opened 2023-07-07
fix: yaml with merge

### LINKED ISSUE #1331: Unable to de-serialize knative yamls
**Describe the bug**
Unable to deserialize
parallels.flows.knative.dev
pingsources.sources.knative.dev

From, https://github.com/knative/eventing/releases/download/knative-v1.10.1/eventing-crds.yaml

```yaml
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  name: parallels.flows.knative.dev
  labels:
    knative.dev/crd-install: "true"
    duck.knative.dev/addressable: "true"
    app.kubernetes.io/version: "1.10.1"
    app.kubernetes.io/name: knative-eventing
spec:
  group: flows.knative.dev
  versions:
    - name: v1
      served: true
      storage: true
      subresources:
        status: {}
      schema:
        openAPIV3Schema:
          description: 'Parallel defines conditional branches that will be wired in series through Channels and Subscriptions.'
          type: object
          properties:
            spec:
              description: Spec defines the desired state of the Parallel.
              type: object
              properties:
                branches:
                  description: Branches is the list of Filter/Subscribers pairs.
                  type: array
                  items:
                    type: object
                    x-kubernetes-preserve-unknown-fields: true
                    properties:
                      delivery:
                        description: Delivery is the delivery specification for events to the subscriber This includes things like retries, DLQ, etc.
                        type: object
                        properties:
                          backoffDelay:
                            description: 'BackoffDelay is the delay before retrying. More information on Duration format: - https://www.iso.org/iso-8601-date-and-time-format.html - https://en.wikipedia.org/wiki/ISO_8601  For linear policy, backoff delay is backoffDelay*<numberOfRetries>. For exponential policy, backoff delay is backoffDelay*2^<numberOfRetries>.'
                            type: string
                          backoffPolicy:
                            description: BackoffPolicy is the retry backoff policy (linear, exponential).
                            type: string
                          deadLetterSink:
                            description: DeadLetterSink is the sink receiving event that could not be sent to a destination.
                            type: object
                            properties: &addressableProperties
                              ref:
                                description: Ref points to an Addressable.
                                type: object
                                properties:
                                  apiVersion:
                                    description: API version of the referent.
                                    type: string
                                  kind:
                                    description: 'Kind of the referent. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
                                    type: string
                                  name:
                                    description: 'Name of the referent. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names'
                                    type: string
                                  namespace:
                                    description: 'Namespace of the referent. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/ This is optional field, it gets defaulted to the object holding it if left out.'
                                    type: string
                              uri:
                                description: URI can be an absolute URL(non-empty scheme and non-empty host) pointing to the target or a relative URI. Relative URIs will be resolved using the base URI retrieved from Ref.
                                type: string
                          retry:
                            description: Retry is the minimum number of retries the sender should attempt when sending an event before moving it to the dead letter sink.
                            type: integer
                            format: int32
                        x-kubernetes-preserve-unknown-fields: true # This is necessary to enable the experimental feature delivery-timeout
                      filter:
                        description: Filter is the expression guarding the branch
                        type: object
                        properties:
                          !!merge <<: *addressableProperties
                      reply:
                        description: Reply is a Reference to where the result of Subscriber of this case gets sent to. If not specified, sent the result to the Parallel Reply
                        type: object
                        properties:
                          !!merge <<: *addressableProperties
                      subscriber:
                        description: Subscriber receiving the event when the filter passes
                        type: object
                        properties:
                          !!merge <<: *addressableProperties
                channelTemplate:
                  description: ChannelTemplate specifies which Channel CRD to use. If left unspecified, it is set to the default Channel CRD for the namespace (or cluster, in case there are no defaults for the namespace).
                  type: object
                  properties:
                    apiVersion:
                      description: 'APIVersion defines the versioned schema of this representation of an object. Servers should convert recognized schemas to the latest internal value, and may reject unrecognized values. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources'
                      type: string
                    kind:
                      description: 'Kind is a string value representing the REST resource this object represents. Servers may infer this from the endpoint the client submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
                      type: string
                    spec:
                      description: Spec defines the Spec to use for each channel created. Passed in verbatim to the Channel CRD as Spec section.
                      type: object
                      x-kubernetes-preserve-unknown-fields: true
                reply:
                  description: Reply is a Reference to where the result of a case Subscriber gets sent to when the case does not have a Reply
                  type: object
                  properties:
                    !!merge <<: *addressableProperties
            status:
              description: Status represents the current state of the Parallel. This data may be out of date.
              type: object
              properties:
                address:
                  type: object
                  properties:
                    url:
                      type: string
                annotations:
                  description: Annotations is additional Status fields for the Resource to save some additional State as well as convey more information to the user. This is roughly akin to Annotations on any k8s resource, just the reconciler conveying richer information outwards.
                  type: object
                  x-kubernetes-preserve-unknown-fields: true
                branchStatuses:
                  description: BranchStatuses is an array of corresponding to branch statuses. Matches the Spec.Branches array in the order.
                  type: array
                  items:
                    type: object
                    properties:
                      filterChannelStatus:
                        description: FilterChannelStatus corresponds to the filter channel status.
                        type: object
                        properties: &channelProperties
                          channel:
                            description: Channel is the reference to the underlying channel.
                            type: object
                            properties: &referentProperties
                              apiVersion:
                                description: API version of the referent.
                                type: string
                              fieldPath:
                                description: 'If referring to a piece of an object instead of an entire object, this string should contain a valid JSON/Go field access statement, such as desiredState.manifest.containers[2]. For example, if the object reference is to a container within a pod, this would take on a value like: "spec.containers{name}" (where "name" refers to the name of the container that triggered the event) or if no container name is specified "spec.containers[2]" (container with index 2 in this pod). This syntax is chosen only to have some well-defined way of referencing a part of an object.'
                                type: string
                              kind:
                                description: 'Kind of the referent. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
                                type: string
                              name:
                                description: 'Name of the referent. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names'
                                type: string
                              namespace:
                                description: 'Namespace of the referent. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/'
                                type: string
                              resourceVersion:
                                description: 'Specific resourceVersion to which this reference is made, if any. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#concurrency-control-and-consistency'
                                type: string
                              uid:
                                description: 'UID of the referent. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#uids'
                                type: string
                          ready:
                            description: ReadyCondition indicates whether the Channel is ready or not.
                            type: object
                            x-kubernetes-preserve-unknown-fields: true
                            properties: &readyConditionProperties
                              message:
                                description: A human readable message indicating details about the transition.
                                type: string
                              reason:
                                description: The reason for the condition's last transition.
                                type: string
                              severity:
                                description: Severity with which to treat failures of this type of condition. When this is not specified, it defaults to Error.
                                type: string
                              status:
                                description: Status of the condition, one of True, False, Unknown.
                                type: string
                              type:
                                description: Type of condition.
                                type: string
                      filterSubscriptionStatus:
                        description: FilterSubscriptionStatus corresponds to the filter subscription status.
                        type: object
                        properties:
                          ready:
                            description: ReadyCondition indicates whether the Subscription is ready or not.
                            type: object
                            properties:
                              !!merge <<: *readyConditionProperties
                          subscription:
                            description: Subscription is the reference to the underlying Subscription.
                            type: object
                            properties:
                              !!merge <<: *referentProperties
                      subscriberSubscriptionStatus:
                        description: SubscriptionStatus corresponds to the subscriber subscription status.
                        type: object
                        properties:
                          ready:
                            description: ReadyCondition indicates whether the Subscription is ready or not.
                            type: object
                            properties:
                              !!merge <<: *readyConditionProperties
                          subscription:
                            description: Subscription is the reference to the underlying Subscription.
                            type: object
                            properties:
                              !!merge <<: *referentProperties
                conditions:
                  description: Conditions the latest available observations of a resource's current state.
                  type: array
                  items:
                    type: object
                    properties:
                      !!merge <<: *readyConditionProperties
                ingressChannelStatus:
                  description: IngressChannelStatus corresponds to the ingress channel status.
                  type: object
                  properties:
                    !!merge <<: *channelProperties
                observedGeneration:
                  description: ObservedGeneration is the 'Generation' of the Service that was last processed by the controller.
                  type: integer
                  format: int64
      additionalPrinterColumns:
        - name: URL
          type: string
          jsonPath: .status.address.url
        - name: Age
          type: date
          jsonPath: .metadata.creationTimestamp
        - name: Ready
          type: string
          jsonPath: ".status.conditions[?(@.type==\"Ready\")].status"
        - name: Reason
          type: string
          jsonPath: ".status.conditions[?(@.type==\"Ready\")].reason"
  names:
    kind: Parallel
    plural: parallels
    singular: parallel
    categories:
      - all
      - knative
      - flows
  scope: Namespaced
```

```yaml
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  labels:
    eventing.knative.dev/source: "true"
    duck.knative.dev/source: "true"
    knative.dev/crd-install: "true"
    app.kubernetes.io/version: "1.10.1"
    app.kubernetes.io/name: knative-eventing
  annotations:
    # TODO add schemas and descriptions
    registry.knative.dev/eventTypes: |
      [
        { "type": "dev.knative.sources.ping" }
      ]
  name: pingsources.sources.knative.dev
spec:
  group: sources.knative.dev
  versions:
    - &version
      name: v1beta2
      served: true
      storage: false
      subresources:
        status: {}
      schema:
        openAPIV3Schema:
          type: object
          description: 'PingSource describes an event source with a fixed payload produced on a specified cron schedule.'
          properties:
            spec:
              type: object
              description: 'PingSourceSpec defines the desired state of the PingSource (from the client).'
              properties:
                ceOverrides:
                  description: 'CloudEventOverrides defines overrides to control the output format and modifications of the event sent to the sink.'
                  type: object
                  properties:
                    extensions:
                      description: 'Extensions specify what attribute are added or overridden on the outbound event. Each `Extensions` key-value pair are set on the event as an attribute extension independently.'
                      type: object
                      additionalProperties:
                        type: string
                      x-kubernetes-preserve-unknown-fields: true
                contentType:
                  description: 'ContentType is the media type of `data` or `dataBase64`. Default is empty.'
                  type: string
                data:
                  description: 'Data is data used as the body of the event posted to the sink. Default is empty. Mutually exclusive with `dataBase64`.'
                  type: string
                dataBase64:
                  description: "DataBase64 is the base64-encoded string of the actual event's body posted to the sink. Default is empty. Mutually exclusive with `data`."
                  type: string
                schedule:
                  description: 'Schedule is the cron schedule. Defaults to `* * * * *`.'
                  type: string
                sink:
                  description: 'Sink is a reference to an object that will resolve to a uri to use as the sink.'
                  type: object
                  properties:
                    ref:
                      description: 'Ref points to an Addressable.'
                      type: object
                      properties:
                        apiVersion:
                          description: 'API version of the referent.'
                          type: string
                        kind:
                          description: 'Kind of the referent. More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds'
                          type: string
                        name:
                          description: 'Name of the referent. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names'
                          type: string
                        namespace:
                          description: 'Namespace of the referent. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/ This is optional field, it gets defaulted to the object holding it if left out.'
                          type: string
                    uri:
                      description: 'URI can be an absolute URL(non-empty scheme and non-empty host) pointing to the target or a relative URI. Relative URIs will be resolved using the base URI retrieved from Ref.'
                      type: string
                timezone:
                  description: 'Timezone modifies the actual time relative to the specified timezone. Defaults to the system time zone. More general information about time zones: https://www.iana.org/time-zones List of valid timezone values: https://en.wikipedia.org/wiki/List_of_tz_database_time_zones'
                  type: string
            status:
              type: object
              description: 'PingSourceStatus defines the observed state of PingSource (from the controller).'
              properties:
                annotations:
                  description: 'Annotations is additional Status fields for the Resource to save some additional State as well as convey more information to the user. This is roughly akin to Annotations on any k8s resource, just the reconciler conveying richer information outwards.'
                  type: object
                  x-kubernetes-preserve-unknown-fields: true
                ceAttributes:
                  description: 'CloudEventAttributes are the specific attributes that the Source uses as part of its CloudEvents.'
                  type: array
                  items:
                    type: object
                    properties:
                      source:
                        description: 'Source is the CloudEvents source attribute.'
                        type: string
                      type:
                        description: 'Type refers to the CloudEvent type attribute.'
                        type: string
                conditions:
                  description: 'Conditions the latest available observations of a resource''s current state.'
                  type: array
                  items:
                    type: object
                    required:
                      - type
                      - status
                    properties:
                      lastTransitionTime:
                        description: 'LastTransitionTime is the last time the condition transitioned from one status to another. We use VolatileTime in place of metav1.Time to exclude this from creating equality.Semantic differences (all other things held constant).'
                        type: string
                      message:
                        description: 'A human readable message indicating details about the transition.'
                        type: string
                      reason:
                        description: 'The reason for the condition''s last transition.'
                        type: string
                      severity:
                        description: 'Severity with which to treat failures of this type of condition. When this is not specified, it defaults to Error.'
                        type: string
                      status:
                        description: 'Status of the condition, one of True, False, Unknown.'
                        type: string
                      type:
                        description: 'Type of condition.'
                        type: string
                observedGeneration:
                  description: 'ObservedGeneration is the "Generation" of the Service that was last processed by the controller.'
                  type: integer
                  format: int64
                sinkUri:
                  description: 'SinkURI is the current active sink URI that has been configured for the Source.'
                  type: string
      additionalPrinterColumns:
        - name: Sink
          type: string
          jsonPath: .status.sinkUri
        - name: Schedule
          type: string
          jsonPath: .spec.schedule
        - name: Age
          type: date
          jsonPath: .metadata.creationTimestamp
        - name: Ready
          type: string
          jsonPath: ".status.conditions[?(@.type=='Ready')].status"
        - name: Reason
          type: string
          jsonPath: ".status.conditions[?(@.type=='Ready')].reason"
    - !!merge <<: *version
      name: v1
      served: true
      storage: true
      # v1 schema is identical to the v1beta2 schema
  names:
    categories:
      - all
      - knative
      - sources
    kind: PingSource
    plural: pingsources
    singular: pingsource
  scope: Namespaced
  conversion:
    strategy: Webhook
    webhook:
      conversionReviewVersions: ["v1", "v1beta1"]
      clientConfig:
        service:
          name: eventing-webhook
          namespace: knative-eventing
```

11.0.36

**Dotnet Runtime Version**
e.g. net7

**To Reproduce**

**Expected behavior**
Yamls should de-serialize

IvanJosipovic:
Fixes #1331

I needed to change CommonDeserializerBuilder to a method, without this, I kept on getting duplicate key errors.

### DIFF
--- a/src/KubernetesClient.Models/KubernetesYaml.cs
+++ b/src/KubernetesClient.Models/KubernetesYaml.cs
@@ -15,23 +15,21 @@ namespace k8s
     /// </summary>
     public static class KubernetesYaml
     {
-        private static DeserializerBuilder CommonDeserializerBuilder()
-        {
-            return new DeserializerBuilder()
+        private static DeserializerBuilder CommonDeserializerBuilder =>
+            new DeserializerBuilder()
                 .WithNamingConvention(CamelCaseNamingConvention.Instance)
                 .WithTypeConverter(new IntOrStringYamlConverter())
                 .WithTypeConverter(new ByteArrayStringYamlConverter())
                 .WithTypeConverter(new ResourceQuantityYamlConverter())
                 .WithAttemptingUnquotedStringTypeDeserialization()
                 .WithOverridesFromJsonPropertyAttributes();
-        }
 
         private static readonly IDeserializer StrictDeserializer =
-            CommonDeserializerBuilder()
+            CommonDeserializerBuilder
             .WithDuplicateKeyChecking()
             .Build();
         private static readonly IDeserializer Deserializer =
-            CommonDeserializerBuilder()
+            CommonDeserializerBuilder
             .IgnoreUnmatchedProperties()
             .Build();
         private static IDeserializer GetDeserializer(bool strict) => strict ? StrictDeserializer : Deserializer;
--- a/src/KubernetesClient.Models/KubernetesYaml.cs
+++ b/src/KubernetesClient.Models/KubernetesYaml.cs
@@ -15,20 +15,23 @@ namespace k8s
     /// </summary>
     public static class KubernetesYaml
     {
-        private static readonly DeserializerBuilder CommonDeserializerBuilder =
-            new DeserializerBuilder()
+        private static DeserializerBuilder CommonDeserializerBuilder()
+        {
+            return new DeserializerBuilder()
                 .WithNamingConvention(CamelCaseNamingConvention.Instance)
                 .WithTypeConverter(new IntOrStringYamlConverter())
                 .WithTypeConverter(new ByteArrayStringYamlConverter())
                 .WithTypeConverter(new ResourceQuantityYamlConverter())
                 .WithAttemptingUnquotedStringTypeDeserialization()
                 .WithOverridesFromJsonPropertyAttributes();
+        }
+
         private static readonly IDeserializer StrictDeserializer =
-            CommonDeserializerBuilder
+            CommonDeserializerBuilder()
             .WithDuplicateKeyChecking()
             .Build();
         private static readonly IDeserializer Deserializer =
-            CommonDeserializerBuilder
+            CommonDeserializerBuilder()
             .IgnoreUnmatchedProperties()
             .Build();
         private static IDeserializer GetDeserializer(bool strict) => strict ? StrictDeserializer : Deserializer;
@@ -154,7 +157,7 @@ namespace k8s
             typeMap?.ToList().ForEach(x => mergedTypeMap[x.Key] = x.Value);
 
             var types = new List<Type>();
-            var parser = new Parser(new StringReader(content));
+            var parser = new MergingParser(new Parser(new StringReader(content)));
             parser.Consume<StreamStart>();
             while (parser.Accept<DocumentStart>(out _))
             {
@@ -162,7 +165,7 @@ namespace k8s
                 types.Add(mergedTypeMap[dict["apiVersion"] + "/" + dict["kind"]]);
             }
 
-            parser = new Parser(new StringReader(content));
+            parser = new MergingParser(new Parser(new StringReader(content)));
             parser.Consume<StreamStart>();
             var ix = 0;
             var results = new List<object>();
@@ -205,12 +208,14 @@ namespace k8s
 
         public static TValue Deserialize<TValue>(string yaml, bool strict = false)
         {
-            return GetDeserializer(strict).Deserialize<TValue>(yaml);
+            using var reader = new StringReader(yaml);
+            return GetDeserializer(strict).Deserialize<TValue>(new MergingParser(new Parser(reader)));
         }
 
         public static TValue Deserialize<TValue>(Stream yaml, bool strict = false)
         {
-            return GetDeserializer(strict).Deserialize<TValue>(new StreamReader(yaml));
+            using var reader = new StreamReader(yaml);
+            return GetDeserializer(strict).Deserialize<TValue>(new MergingParser(new Parser(reader)));
         }
 
         public static string SerializeAll(IEnumerable<object> values)
--- a/tests/KubernetesClient.Tests/KubernetesYamlTests.cs
+++ b/tests/KubernetesClient.Tests/KubernetesYamlTests.cs
@@ -901,5 +901,199 @@ spec:
             Assert.Equal(12.0, Assert.IsType<float>(v1AlphaFoo.Spec["float"]), 3);
             Assert.Equal(content.Replace("\r\n", "\n"), KubernetesYaml.SerializeAll(objs).Replace("\r\n", "\n"));
         }
+
+        [Fact]
+        public void LoadFromStringCRDMerge()
+        {
+            var content = @"apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  labels:
+    eventing.knative.dev/source: ""true""
+    duck.knative.dev/source: ""true""
+    knative.dev/crd-install: ""true""
+    app.kubernetes.io/version: ""1.10.1""
+    app.kubernetes.io/name: knative-eventing
+  annotations:
+    # TODO add schemas and descriptions
+    registry.knative.dev/eventTypes: |
+      [
+        { ""type"": ""dev.knative.sources.ping"" }
+      ]
+  name: pingsources.sources.knative.dev
+spec:
+  group: sources.knative.dev
+  versions:
+    - &version
+      name: v1beta2
+      served: true
+      storage: false
+      subresources:
+        status: {}
+      schema:
+        openAPIV3Schema:
+          type: object
+          description: 'PingSource describes an event source with a fixed payload produced on a specified cron schedule.'
+          properties:
+            spec:
+              type: object
+              description: 'PingSourceSpec defines the desired state of the PingSource (from the client).'
+              properties:
+                ceOverrides:
+                  description: 'CloudEventOverrides defines overrides to control the output format and modifications of the event sent to the sink.'
+                  type: object
+                  properties:
+                    extensions:
+                      description: 'Extensions specify what attribute are added or overridden on the outbound event. Each `Extensions` key-value pair are set on the event as an attribute extension independently.'
+                      type: object
+                      additionalProperties:
+                        type: string
+                      x-kubernetes-preserve-unknown-fields: true
+                contentType:
+                  description: 'ContentType is the media type of `data` or `dataBase64`. Default is empty.'
+                  type: string
+                data:
+                  description: 'Data is data used as the body of the event posted to the sink. Default is empty. Mutually exclusive with `dataBase64`.'
+                  type: string
+                dataBase64:
+                  description: ""DataBase64 is the base64-encoded string of the actual event's body posted to the sink. Default is empty. Mutually exclusive with `data`.""
+                  type: string
+                schedule:
+                  description: 'Schedule is the cron schedule. Defaults to `* * * * *`.'
+                  type: string
+                sink:
+                  description: 'Sink is a reference to an object that will resolve to a uri to use as the sink.'
+                  type: object
+                  properties:
+                    ref:
+                      description: 'Ref points to an Addressable.'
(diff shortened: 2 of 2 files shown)


tg123:
/LGTM

thanks!

src/KubernetesClient.Models/KubernetesYaml.cs
```diff
@@ -15,14 +15,15 @@ namespace k8s
     /// </summary>
     public static class KubernetesYaml
     {
-        private static readonly DeserializerBuilder CommonDeserializerBuilder =
+        private static DeserializerBuilder CommonDeserializerBuilder =>
```
tg123, line 18: have to `new` every time?

IvanJosipovic: Yea, without this the tests fail. This code is only ran 2x, once to new up the StrictDeserializer and once for the normal Deserializer.
image

tg123 (approved):
/LGTM

PR #1348, opened 2023-07-25
CStatus should be status otherwise it doesn't work at all to load 'st…

lknite:
When trying to load some 'status' items for the first time realized 'CStatus' in the provided example CustomResourceDefinition should have actually been 'status'.  After correcting the typo status items were then available.


tg123 (approved):
/LGTM

Issue #1350, opened 2023-07-26
Timestamp deserialization fails on well-formatted timestamps depending on number of nanosecond digits


anekdoti:
**Describe the bug**
Timestamp deserialization is more limited than Kubernetes API conventions allow. I.e., there are timestamps that are fine according to Kubernetes API conventions (e.g., `lastTransitionTimestamp` in status conditions) that can not be deserialized using the C# Kubernetes client but instead raise an exception like shown below. 

More precisely, the C# Kubernetes client only allows either exactly 6 digits for nanoseconds or none at all, whereas an arbitrary number < 10 is actually possible.

**Kubernetes C# SDK Client Version**
Observed on `10.0.31`, but according to the implementation of `KubernetesDateTimeOffsetConverter` on the master branch we expect the bug to occur on master as well.

**Server Kubernetes Version**
`1.24.14`

**Dotnet Runtime Version**
net6

**To Reproduce**
Create a Kubernetes resource with a status condition where the `lastTransitionTimestamp` has e.g. 9 nanosecond positions.
Attempt to read the resource with the generic client of the C# Kubernetes Client. The exception shown below will be thrown.

**Expected behavior**
The Kubernetes resource should be read properly, including the `lastTransitionTimestamp` of the status condition.

**Where do you run your app with Kubernetes SDK (please complete the following information):**
 The Kubernetes SDK is used in a container running the image `mcr.microsoft.com/dotnet/sdk:6.0-alpine3.16` in Kubernetes `1.24.14`.

**Additional context**
According to the [Kubernetes API conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties), the lastTransitionTimestamp should be deserializable as a Golang [Time](https://pkg.go.dev/time#Time) struct. The deserialization of a string into such a struct follows the conventions of [RFC3339](https://www.rfc-editor.org/rfc/rfc3339.html) and [its implementation](https://cs.opensource.google/go/go/+/refs/tags/go1.20.6:src/time/format.go;l=1494) in particular allows an arbitrary number of nanosecond digits from 0 to 9.

We actually faced this problem in practice trying to read a `KafkaConnector` resource by the [Strimzi](https://strimzi.io/) operator that contained a status condition with the `lastTransitionTimestamp` as shown in the exception below. According to the Kubernetes API conventions, this timestamp is well-formatted and should be deserialized correctly.

[...]

PR #1351, opened 2023-07-27
support for rfc3339 nano

tg123:
#1350

### DIFF
--- a/tests/KubernetesClient.Tests/KubernetesJsonTests.cs
+++ b/tests/KubernetesClient.Tests/KubernetesJsonTests.cs
@@ -5,6 +5,7 @@ namespace k8s.Tests;
 
 public class KubernetesJsonTests
 {
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("Build", "CA1812:'KubernetesJsonTests.RfcTime' is an internal class that is apparently never instantiated. If so, remove the code from the assembly. If this class is intended to contain only static members, make it 'static' (Module in Visual Basic). (https://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1812)", Justification = "json type")]
     private class RfcTime
     {
         public DateTime rfc3339 { get; set; }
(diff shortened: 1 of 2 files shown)


tests/KubernetesClient.Tests/KubernetesJsonTests.cs
```diff
@@ -41,8 +42,33 @@ public void RFC3339()
         }
         */
 
-        var json = "{\"rfc3339\":\"2009-11-10T23:00:00Z\",\"rfc3339micro\":\"2009-11-10T23:00:00.000000Z\",\"rfc3339nano\":\"2009-11-10T13:00:00.000000000Z\"}\r\n";
+        // second changed from go playground
+        var json = "{\"rfc3339\":\"2009-11-10T23:45:12Z\",\"rfc3339micro\":\"2009-11-10T23:45:12.123456Z\",\"rfc3339nano\":\"2009-11-10T13:24:56.123456789Z\"}\r\n";
```
ndallmeyer, line 47: Please consider the following datetimes as test-cases, that are valid with regard to Kubernetes-Api, but all lead to a FormatException, because they not match 3, 6, or 9 digits:

"2009-11-10T13:24:56.12345678Z"
"2009-11-10T13:24:56.12345Z"
"2009-11-10T13:24:56.1234Z"
"2009-11-10T13:24:56.12Z"
"2009-11-10T13:24:56.1Z"

tg123: any reason to support those format that server will never output?

anekdoti: Hello @tg123. As already described in issue https://github.com/kubernetes-client/csharp/issues/1350, timestamps in the Kubernetes API are implemented by the Golang type [`time.Time`](https://pkg.go.dev/time#Time) which implicitly defines what is and is not allowed in terms of nanosecond digits.

Parsing a string as this type allows an arbitrary number of nanosecond digits from 0-9. And in fact, also deserializing an instance of this type may lead to an arbitrary number of nanosecond digits from 0-9. Consider [this example](https://go.dev/play/p/VWdgDKRBWHH).

I currently see nothing that explicitly excludes that this actually happens in practice and see the burden of proof for the contrary more on your side.

tg123: i do not think it should be supported through
why not if just one line change

anekdoti: I don't understand your second line (why not if just one line change). For the first line: I think it doesn't matter what we think about whether arbitrary nanosecond digits should be supported or not - all what matters is that the Kubernetes API permits it. And as long as the C# Kubernetes Client does not support it, it will fail sooner or later in a very unpleasant and subtil way. I think an a bit more complex implementation of the deserialization is a small price for avoiding this.

anekdoti: Sorry for the misunderstanding! Your solution is actually very simple and elegant. Thank you :-)

ndallmeyer (approved):

anekdoti (approved):

tg123:
ping @brendanburns for approval

PR #1370, opened 2023-08-14
Added support of nine digit nanoseconds format.

ndallmeyer:
Added support for up to nine digit nanoseconds format. The nanoseconds are truncated to 7 decimal places. See https://github.com/kubernetes-client/csharp/issues/1350


tg123:
could you please merge my testcase

ndallmeyer:
Thanks, based on your tests I've implemented tests that cover all edge cases.

src/KubernetesClient.Models/KubernetesJson.cs
```diff
@@ -26,18 +26,49 @@ public override void Write(Utf8JsonWriter writer, TimeSpan value, JsonSerializer
 
         private sealed class KubernetesDateTimeOffsetConverter : JsonConverter<DateTimeOffset>
         {
-            private const string SerializeFormat = "yyyy'-'MM'-'dd'T'HH':'mm':'ss.ffffffK";
-            private const string Iso8601Format = "yyyy'-'MM'-'dd'T'HH':'mm':'ssK";
+            private const string RFC3339NanoFormat = "yyyy-MM-dd'T'HH':'mm':'ss.fffffffK";
+
+            private static string FormatDateTimeOffsetToSevenDigitsNanoseconds(string dateTime)
+            {
+                var isUTC = dateTime.EndsWith("Z");
```
tg123, line 33: too complex
why not just trim extra input?

ndallmeyer: As already pointed out by @anekdoti in https://github.com/kubernetes-client/csharp/issues/1350 the lastTransitionTimestamp should be deserializable as a Golang Time struct that follows RFC3339 where an arbitrary number of nansecond digits from [0 to 9 is valid](https://cs.opensource.google/go/go/+/refs/tags/go1.20.6:src/time/format.go;l=1494). Therefore trimming alone would not help here.
This method allows an arbitrary number of nanosecond digits from 0 to 9.

tg123:
check my impl in #1351

anekdoti:
Hello @tg123,

anekdoti:
Hello @tg123 ,

thank you for your PR https://github.com/kubernetes-client/csharp/pull/1351. 

It solves the specific case that made me open issue https://github.com/kubernetes-client/csharp/issues/1350 where we observed an error with a timestamp with 9 nanosecond digits. 

However, as also described in my issue, timestamps can have **an arbitrary number (up to 9) of nanosecond digits** by the Golang implementation. For this reason, your PR only solves certain special cases whereas the PR on hand by @ndallmeyer solves the general case as by Kubernetes API conventions. 

To avoid further and foreseeable problems with deserializing timestamps, I would therefore strongly suggest to follow the general approach by @ndallmeyer .

tg123:
@anekdoti could you please add bad test cases?
also please make pipeline happy

PR #1372, opened 2023-08-15
Avoid additional serialization/deserialization of response bodies

### LINKED ISSUE #1371: Performance optimization when using `GenericClient`
**Describe the bug**

The `GenericClient` serializes reponse body's (which is already parsed JsonElement) to string and then parses them again via `KubernetesJson.Deserialize<>`.

Adding a new overload to `KubernetesJson.Deserialize<>` which accepts `JsonElement` and replacing the `response.Body.ToString()` calls with `(JsonElement)response.Body` would eliminate the need to deserialize the body twice.

`11.0.21`

**Server Kubernetes Version**
`1.25.9`

**Dotnet Runtime Version**
`net6.0`

prochnowc:
Fixes #1371 

This PR adds additional overloads with generic type argument to all operations which return `object`.

### DIFF
--- a/src/KubernetesClient.Models/KubernetesJson.cs
+++ b/src/KubernetesClient.Models/KubernetesJson.cs
@@ -92,6 +92,10 @@ namespace k8s
             return JsonSerializer.Deserialize<TValue>(json, jsonSerializerOptions ?? JsonSerializerOptions);
         }
 
+        public static TValue Deserialize<TValue>(JsonElement json, JsonSerializerOptions jsonSerializerOptions = null)
+        {
+            return json.Deserialize<TValue>(jsonSerializerOptions ?? JsonSerializerOptions);
+        }
 
         public static string Serialize(object value, JsonSerializerOptions jsonSerializerOptions = null)
         {
--- a/src/KubernetesClient.Models/KubernetesJson.cs
+++ b/src/KubernetesClient.Models/KubernetesJson.cs
@@ -92,11 +92,6 @@ namespace k8s
             return JsonSerializer.Deserialize<TValue>(json, jsonSerializerOptions ?? JsonSerializerOptions);
         }
 
-        public static TValue Deserialize<TValue>(JsonElement json, JsonSerializerOptions jsonSerializerOptions = null)
-        {
-            return json.Deserialize<TValue>(jsonSerializerOptions ?? JsonSerializerOptions);
-        }
-
         public static string Serialize(object value, JsonSerializerOptions jsonSerializerOptions = null)
         {
             return JsonSerializer.Serialize(value, jsonSerializerOptions ?? JsonSerializerOptions);
--- a/src/LibKubernetesGenerator/templates/Operations.cs.template
+++ b/src/LibKubernetesGenerator/templates/Operations.cs.template
@@ -83,5 +83,72 @@ public partial class AbstractKubernetes : I{{name}}Operations
         {{/IfReturnType operation "stream"}}
         return result;
     }
+
+    {{#IfReturnType operation "object"}}
+    /// <inheritdoc/>
+    async Task<HttpOperationResponse<T>> I{{name}}Operations.{{GetMethodName operation "WithHttpMessagesAsync"}}<T>(
+{{#operation.parameters}}
+        {{GetDotNetType .}} {{GetDotNetName .}},
+{{/operation.parameters}}
+        IReadOnlyDictionary<string, IReadOnlyList<string>> customHeaders,
+        CancellationToken cancellationToken)
+    {
+        using var cts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
+        cts.CancelAfter(HttpClientTimeout);
+        {{#IfParamContains operation "watch"}}
+        if (watch == true)
+        {
+            cts.CancelAfter(Timeout.InfiniteTimeSpan);
+        }
+        {{/IfParamContains operation "watch"}}
+        cancellationToken = cts.Token;
+
+        {{#operation.parameters}}
+        {{#isRequired}}
+        if ({{GetDotNetName name}} == null)
+        {
+            throw new ArgumentNullException("{{GetDotNetName name}}");
+        }
+        {{/isRequired}}
+        {{/operation.parameters}}
+
+        // Construct URL
+        var url = $"{{ToInterpolationPathString path}}";
+        {{#IfGroupPathParamContainsGroup path}}
+        url = url.Replace("apis//", "api/");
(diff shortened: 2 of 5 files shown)


prochnowc:
The e2e test worked fine on my local Maschine.

src/KubernetesClient.Models/KubernetesJson.cs
```diff
@@ -92,6 +92,10 @@ public static TValue Deserialize<TValue>(Stream json, JsonSerializerOptions json
             return JsonSerializer.Deserialize<TValue>(json, jsonSerializerOptions ?? JsonSerializerOptions);
         }
 
+        public static TValue Deserialize<TValue>(JsonElement json, JsonSerializerOptions jsonSerializerOptions = null)
```
tg123, line 95: is it ok to make this internal?

prochnowc: IMHO this should be kept public, otherwise it will not be possible to make use of it when using `ICustomObjectsOperations`.

tg123: no worried, it by default
```
[assembly: InternalsVisibleTo("KubernetesClient")]
```

this api looked unreasonable to people who do not have context. so let keep it secret

prochnowc: To clarify:

Lets suppose in my application I want to read some custom object, I'am still forced to use double serialization/deserialization without this being public:

```csharp
var result = await kubernetes.CustomObjects.ListClusterCustomObjectAsync(/*...*/);
var customObj = KubernetesJson.Deserialize<V1MyCustomObj>(result.ToString());
```

a possible work around would be to cast the result to `JsonElement` and invoke `Deserialize` on it:

```csharp
var result = await kubernetes.CustomObjects.ListClusterCustomObjectAsync(/*...*/);
var customObj = ((JSonElement)result).Deserialize<V1MyCustomObj>();
```

sadly this would not use the custom convertes registered in `KubernetesJson`.

Another way would be to make `JsonSerializerOptions` in `KubernetesJson` public and make the new `Deserialize` method internal, which would enable app's to use the API this way:

```csharp
var result = await kubernetes.CustomObjects.ListClusterCustomObjectAsync(/*...*/);
var customObj = ((JSonElement)result).Deserialize<V1MyCustomObj>(KubernetesJson.JsonSerializerOptions);
```

prochnowc: @tg123 which option do you prefer ?

prochnowc: I've changed the code generation to generate methods for CustomObjects to include a type parameter.
IMHO `GenericClient` could be marked as obsolete.

prochnowc: @tg123 Can you please review again? Thanks!

tg123: This is awesome, i will take a deeper look after runrise Thanks

src/LibKubernetesGenerator/templates/Operations.cs.template
```diff
@@ -83,5 +83,72 @@ public partial class AbstractKubernetes : I{{name}}Operations
         {{/IfReturnType operation "stream"}}
         return result;
     }
+
+    {{#IfReturnType operation "object"}}
```
tg123, line 87: i understand that the template engine is ****
but anyway to method body?

prochnowc: I'm sorry but i dont really understand your comment.

prochnowc: The code is generated as I would expect it.

tg123: this part

```
using var cts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
        cts.CancelAfter(HttpClientTimeout);
        {{#IfParamContains operation "watch"}}
        if (watch == true)
        {
            cts.CancelAfter(Timeout.InfiniteTimeSpan);
        }
        {{/IfParamContains operation "watch"}}
        cancellationToken = cts.Token;

        {{#operation.parameters}}
        {{#isRequired}}
        if ({{GetDotNetName name}} == null)
        {
            throw new ArgumentNullException("{{GetDotNetName name}}");
        }
        {{/isRequired}}
        {{/operation.parameters}}

        // Construct URL
        var url = $"{{ToInterpolationPathString path}}";
        {{#IfGroupPathParamContainsGroup path}}
        url = url.Replace("apis//", "api/");
        {{/IfGroupPathParamContainsGroup}}
        {{#IfListNotEmpty operation.parameters}}
        var q = new QueryBuilder();
        {{#operation.parameters}}
        {{#IfKindIs . "query"}}
        q.Append("{{name}}", {{GetDotNetName name}});
        {{/IfKindIs . "query"}}
        {{/operation.parameters}}
        url += q.ToString();
        {{/IfListNotEmpty operation.parameters}}

        // Create HTTP transport
        {{#IfParamContains operation "body"}}
        var httpResponse = await SendRequest(url, HttpMethods.{{Method}}, customHeaders, body, cancellationToken);
        {{/IfParamContains operation "body"}}
        {{#IfParamDoesNotContain operation "body"}}
        var httpResponse = await SendRequest<object>(url, HttpMethods.{{Method}}, customHeaders, null, cancellationToken);
        {{/IfParamDoesNotContain operation "body"}}
        // Create Result
        var httpRequest = httpResponse.RequestMessage;
        var result = await CreateResultAsync<T>(
            httpRequest,
            httpResponse,
            {{#IfParamContains operation "watch"}}
            watch,
            {{/IfParamContains operation "watch"}}
            {{#IfParamDoesNotContain operation "watch"}}
            false,
            {{/IfParamDoesNotContain operation "watch"}}
            cancellationToken);
        return result;
```

prochnowc: Yes, but what is wrong with this part?

tg123: nothing wrong but is there any way to share method body with https://github.com/kubernetes-client/csharp/blob/700c422fd4f4630d2d43b3d6bf176ff9331317ee/src/LibKubernetesGenerator/templates/Operations.cs.template#L13C6-L13C6

duplicate code

prochnowc: Changed code generation to not generate duplicate code.

tg123 (approved):
/LGTM

Issue #1340, opened 2023-07-13
Invalid url in GetServiceAccountIssuerOpenIDConfigurationAsync


molszews:
**Describe the bug**
Calling `GetServiceAccountIssuerOpenIDConfigurationAsync` results in url request to `/.well-known/openid-configuration/` path which is rejected by k8s API (403). I've noticed that removing trailing `/` from path, i.e. using `/.well-known/openid-configuration` is accepted by k8s API and responds with correct data (tested with curl and manually issuing http request from container).

edit: I think the same applies to `GetServiceAccountIssuerOpenIDKeysetAsync` method
edit2: looks like return type of the method (`string`) is invalid as well as it causes deserialization issues:
```
'System.Text.Json.JsonException: The JSON value could not be converted to System.String. Path: $ | LineNumber: 0 | BytePositionInLine: 1.
 ---> System.InvalidOperationException: Cannot get the value of a token type 'StartObject' as a string.
```

**Kubernetes C# SDK Client Version**
`11.0.36`

**Server Kubernetes Version**
`1.26.3`

**Dotnet Runtime Version**
.net7

**To Reproduce**
```
var config = KubernetesClientConfiguration.InClusterConfig();
var client = new Kubernetes(config);
            
var openIdConnectConfigurationRaw = await client.GetServiceAccountIssuerOpenIDConfigurationAsync(default);
```

**Expected behavior**
`openIdConnectConfigurationRaw` should contain JSON document contents.

**Where do you run your app with Kubernetes SDK (please complete the following information):**
 - OS: Linux
 - Environment: container
 - Cloud: Azure, AKS

**Additional context**
Sample stack trace:
```
2023-07-13 09:00:38.689Z k8s.Autorest.HttpOperationException: Operation returned an invalid status code 'Forbidden', response body {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"forbidden: User \"system:serviceaccount:portalpr10490:portal-users-service-sa\" cannot get path \"/.well-known/openid-configuration/\"","reason":"Forbidden","details":{},"code":403}
2023-07-13 09:00:38.689Z 
2023-07-13 09:00:38.689Z    at k8s.Kubernetes.SendRequestRaw(String requestContent, HttpRequestMessage httpRequest, CancellationToken cancellationToken)
[...]

PR #1373, opened 2023-08-16
update swagger gen from 1.28.0

tg123:


tg123:
include fix to #1340 with https://github.com/kubernetes-client/gen/pull/248

PR #1377, opened 2023-08-20
Conditionally redirect stderr for exec auth provider

bianyifan:
External command authentication provider may prompt user interaction in stderr (e.g. [Azure/kubelogin device code flow](https://github.com/Azure/kubelogin/blob/e5dd8912ed0b6e4a4f0eda8f8cc8012c26d66b6c/pkg/token/devicecode.go#L46C55-L46C55)).
If stderr is redirected, users won't be able to see the prompt in the console.

Globally searching for `ExecStdError` didn't yield any meaningful result, so I would just remove that event handler and disable stderr redirection.

### DIFF
--- a/src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
+++ b/src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
@@ -494,7 +494,7 @@ namespace k8s
             throw new KubeConfigException("Refresh not supported.");
         }
 
-        public static Process CreateRunnableExternalProcess(ExternalExecution config)
+        public static Process CreateRunnableExternalProcess(ExternalExecution config, EventHandler<DataReceivedEventArgs> captureStdError = null)
         {
             if (config == null)
             {
@@ -535,7 +535,7 @@ namespace k8s
             }
 
             process.StartInfo.RedirectStandardOutput = true;
-            process.StartInfo.RedirectStandardError = true;
+            process.StartInfo.RedirectStandardError = captureStdError != null;
             process.StartInfo.UseShellExecute = false;
             process.StartInfo.CreateNoWindow = true;
 
@@ -560,14 +560,15 @@ namespace k8s
                 throw new ArgumentNullException(nameof(config));
             }
 
-            var process = CreateRunnableExternalProcess(config);
+            var captureStdError = ExecStdError;
+            var process = CreateRunnableExternalProcess(config, captureStdError);
 
             try
             {
                 process.Start();
-                if (ExecStdError != null)
+                if (captureStdError != null)
                 {
-                    process.ErrorDataReceived += (s, e) => ExecStdError.Invoke(s, e);
+                    process.ErrorDataReceived += captureStdError.Invoke;
                     process.BeginErrorReadLine();
                 }
             }


IvanJosipovic:
See this PR on why this was added, https://github.com/kubernetes-client/csharp/pull/1267#discussion_r1176035653

bianyifan:
I can understand back in time the stderr messages are catched and being used in exception messages, but that got changed in #1267; nothing reads redirected stderr stream now and prompts in stderr got lost.
I've also checked golang impl, golang is directly passing stderr to os.stderr.

And for error handling, I believe we should check `process.ExitCode` after `process.WaitForExit`, before trying to deserialize stdout.

IvanJosipovic:
> I can understand back in time the stderr messages are catched and being used in exception messages, but that got changed in #1267; nothing reads redirected stderr stream now and prompts in stderr got lost. I've also checked golang impl, golang is directly passing stderr to os.stderr.
> 
> And for error handling, I believe we should check `process.ExitCode` after `process.WaitForExit`, before trying to deserialize stdout. I could send another PR for that.

It has been a little time since that PR, however, I believe that approach would cause the process to hang as Kubelogin waits for you to log into a http based url and won't exit until the process is killed or the login has occurred.


Grab kubelogin and try https://github.com/Azure/kubelogin/releases

kubelogin get-token --environment AzurePublicCloud --server-id 6dae42f8-4368-4678-94ff-3960e28e3630 --client-id 80faf920-1908-4b52-b5ef-a8e7bedfc67a --tenant-id b82075ce-f897-4df8-8624-47b71a1fd251 --login devicecode


Maybe we could set RedirectStandardError to true only if ExecStdError is not null?

tg123:
seems a customizable `exec auth provider` would be better

bianyifan:
> It has been a little time since that PR, however, I believe that approach would cause the process to hang as Kubelogin waits for you to log into a http based url and won't exit until the process is killed or the login has occurred.
Exactly, I am not able to login without a prompt and kubelogin keeps waiting for user interaction.

> Maybe we could set RedirectStandardError to true only if ExecStdError is not null?
> seems a customizable exec auth provider would be better
Let me try summarize:
- Users can opt into capturing stderr with `ExecStdError`.
- If user opted into capturing stderr, we would redirect the stderr to `ExecStdError`. Otherwise, send stderr directly to console.
- Also add exit code check right after `if (!process.WaitForExit((int)(ExecTimeout.TotalMilliseconds)))`, so users could get a rough idea of what happened even without stderr.
Thanks for your input and I will make an update!

bianyifan:
Updated.
However I don't think using a global event handler to process stderr message is a great design: if I am reading multiple different kubeconfig files at the same time, it would be tricky to identify where each log comes from.

bianyifan:
> lets do a proposal to remove static?

I'd like to understand the purpose of capturing stderr. If the purpose is just to include stderr output into exception messages in non-user-interactive scenarios, ideally we could just use `System.Environment.UserInteractive` as a switch, if `UserInteractive = false`, capture the stderr and include it in the thrown exception; if `UserInteractive = true`, disable `RedirectStandardError`.
[Python impl is doing it this way](https://github.com/kubernetes-client/python/blob/76ec6f5ce5572f66d82717e296480dc42584f0b7/kubernetes/base/config/exec_provider.py#L74C36-L74C36).
Unfortunately this only works on Windows. On [UNIX or browser `UserInteractive` is always true](https://github.com/dotnet/runtime/blob/8b274d3dd883bdedd3800cff0ac2faed7a2a1529/src/libraries/System.Private.CoreLib/src/System/Environment.UnixOrBrowser.cs#L15).

So, currently I am thinking:
1. Remove [ExecuteExternalCommand from KubernetesClientConfiguration](https://github.com/bianyifan/csharp/blob/16a1e41bd8d98ccc70330ad7ede4a80b54a875c1/src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs#L468-L469). The initial token retrieval can be handled by `ExecTokenProvider` as well.
2. Currently the call chain is `BuildConfigFrom*` -> `GetKubernetesClientConfiguration` -> `InitializeContext` -> `SetUserDetails` -> `new ExecTokenProvider`. We could add an optional parameter `bool? allowInteractiveAuthentication = null` at each level. If `allowInteractiveAuthentication ?? (RuntimeInformation.IsOSPlatform(OSPlatform.Windows) && Environment.UserInteractive)`, stderr will be redirected and included in exception message.

If this plan looks good to you, I can start another PR for that.

IvanJosipovic:
One more thing to note, when the exception occurs, we need to keep the kubelogin process around if it doesn't exit on its own, otherwise the login process will not complete. However there is a concern that we could spawn many zombie processes.

bianyifan:
> One more thing to note, when the exception occurs, we need to keep the kubelogin process around if it doesn't exit on its own, otherwise the login process will not complete. However there is a concern that we could spawn many zombie processes.

I believe this is only a concern in the try-block around `process.Start` and the if-block of `WaitForExt(int)`, right?
For try-block around `process.Start`, if exception happens there, it could be process didn't start at all, or failed to redirect stderr to event handler.
If process didn't even start, we have nothing to clean up.
If stderr redirection failed, or the process didn't exit within the timeout, I believe we should terminate the process since we already considered the authentication attempt as failed, there's no use keeping it around.

tg123:
any additional comments? @IvanJosipovic

tg123:
good to merge @IvanJosipovic ?

tg123 (approved):
/LGTM

bianyifan:
Thanks @tg123 ! I will work on removing static stderr handler as we have discussed before soon.

PR #1403, opened 2023-09-19
removing line breaks

ceneafar:


tg123:
please sign cla. 
so just formatting?

PR #1437, opened 2023-10-25
full clone cacert from config object

### LINKED ISSUE #1436: KubernetesClientConfiguration is not thread-safe
**Describe the bug**
KubernetesClientConfiguration is mutated by the `Kubernetes.Dispose()` Method as the `Kubernetes` constructor just copies the reference to the `SslCaCerts` property and `Kubernetes.Dispose()` disposes that collection, which is actually coming from the `KubernetesClientConfiguration` instance.

See https://github.com/kubernetes-client/csharp/blob/f55d1faf312b4b6fdbf3f96d2508fc1868bf1397/src/KubernetesClient/Kubernetes.ConfigInit.cs#L26
And https://github.com/kubernetes-client/csharp/blob/f55d1faf312b4b6fdbf3f96d2508fc1868bf1397/src/KubernetesClient/Kubernetes.cs#L213-L222

`12.1.1`

**Server Kubernetes Version**
`v1.27.3`

**Dotnet Runtime Version**
`net7`

**To Reproduce**
* Load a `KubernetesClientConfiguration` instance which contains  `SslCaCerts` entries
* Pass that instance to the `Kubernetes` constructor
* Dispose the `Kubernetes` instance
* Check the `KuberenetesClientConfiguration.SSlCaCerts` collection
  * The collection is now empty

**Expected behavior**
The `KubernetesClientConfiguration` should not be modified. Maybe expose a Dispose method on the `KubernetesClientConfiguration` to dispose the `CaCerts`.

Alternatively document explicitly that a `KubernetesClientConfiguration` can be used only by exactly on `Kubernetes` instance and must be abandoned after passing it to a single `Kubernetes` constructor.

tg123:
fix #1436

### DIFF
--- a/src/KubernetesClient/Kubernetes.ConfigInit.cs
+++ b/src/KubernetesClient/Kubernetes.ConfigInit.cs
@@ -23,16 +23,21 @@ namespace k8s
         {
             Initialize();
             ValidateConfig(config);
-            CaCerts = config.SslCaCerts;
+
+            if (config.SslCaCerts != null)
+            {
+                CaCerts = new X509Certificate2Collection(config.SslCaCerts.Select(c => new X509Certificate2(c)).ToArray());
+            }
+
             SkipTlsVerify = config.SkipTlsVerify;
             TlsServerName = config.TlsServerName;
             CreateHttpClient(handlers, config);
             InitializeFromConfig(config);
             HttpClientTimeout = config.HttpClientTimeout;
             jsonSerializerOptions = config.JsonSerializerOptions;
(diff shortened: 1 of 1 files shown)


PSanetra (commented):
As far as I see there is still an issue with this implementation. The certificates are now copied from the `KubernetesClientConfiguration` instance to each consuming `KubernetesClient` instance. Each `KubernetesClient` instance will dispose their copied `X509Certificate2` instances on their `Dispose()` call, but the `config.SslCaCerts` will get never disposed as `KubernetesClientConfiguration` does not implement `IDisposable`.

PR #1449, opened 2023-11-10
Fix possible resource leak on `KubernetesClientConfiguration` without introducing `IDisposable`

### LINKED ISSUE #1446: KubernetesClientConfiguration does not implement IDisposable.
@brendandburns @tg123 As far as I see there is still an issue with this implementation. The certificates are now copied from the `KubernetesClientConfiguration` instance to each consuming `KubernetesClient` instance. Each `KubernetesClient` instance will dispose their copied `X509Certificate2` instances on their `Dispose()` call, but the `config.SslCaCerts` will get never disposed as `KubernetesClientConfiguration` does not implement `IDisposable`.

_Originally posted by @PSanetra in https://github.com/kubernetes-client/csharp/pull/1437#pullrequestreview-1720240995_

PSanetra:
Fix #1446 without introducing `IDisposable` on `KubernetesClientConfiguration`

Breaking Changes:
  * Using `KubernetesClientConfiguration.SslCaCerts` or `KubernetesClientConfiguration.LoadSslCaCerts()` requires the user to dispose the returned `X509Certificate2Collection` after usage.
  * Setter of `KubernetesClientConfiguration.SslCaCerts` was removed.

### DIFF
--- a/src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
+++ b/src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
@@ -4,7 +4,6 @@ using k8s.KubeConfigModels;
 using System.Diagnostics;
 using System.Net;
 using System.Runtime.InteropServices;
-using System.Security.Cryptography.X509Certificates;
 
 namespace k8s
 {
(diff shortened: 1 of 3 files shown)


tg123:
prefer idisposable which does not introduce breaking change

PSanetra:
@tg123 Introducing the `IDisposable` interface is a breaking change as well as it requires to introduce code changes in many code bases. All code bases which instantiated a `KubernetesClientConfiguration` somewhere must dispose the `KubernetesClientConfiguration`.

In comparison the proposed approach in this PR will probably affect only very few code bases. Only those which did directly use `SslCaCerts`. I guess this will be mostly just this project internally.

PSanetra:
@tg123 Can you have another look at this PR?

Issue #1346, opened 2023-07-25
Kubelogin not supported ?


Raulster24:
**Describe the bug**
Kube config file containing kubelogin commands does not work and throws timeout exception

**Kubernetes C# SDK Client Version**
11.0.44

**Server Kubernetes Version**
1.25.5

**Dotnet Runtime Version**
net6

**KubeConfig**
```
kind: Config
preferences: {}
users:
- name: clusterUser*****
  user:
    exec:
      apiVersion: client.authentication.k8s.io/v1beta1
      args:
      - get-token
      - --login
      - azurecli
      - --server-id
      - *******
      command: kubelogin
      env: null
      provideClusterInfo: false
```
Please confirm if it is not supported, if not supported then is there any workaround without using kubectl proxy or generating token

tg123:
kubelogin is supported 
may i know the details of exceptions

Raulster24:
`k8s.Exceptions.KubeConfigException
  HResult=0x80131500
  Message=external exec failed due to uncaught exception: k8s.Exceptions.KubeConfigException: external exec failed due to timeout
   at k8s.KubernetesClientConfiguration.ExecuteExternalCommand(ExternalExecution config)
  Source=KubernetesClient
  StackTrace:
   at k8s.KubernetesClientConfiguration.ExecuteExternalCommand(ExternalExecution config) in k8s\KubernetesClientConfiguration.cs:line 481
   at k8s.KubernetesClientConfiguration.SetUserDetails(K8SConfiguration k8SConfig, Context activeContext) in k8s\KubernetesClientConfiguration.cs:line 329
   at k8s.KubernetesClientConfiguration.InitializeContext(K8SConfiguration k8SConfig, String currentContext) in k8s\KubernetesClientConfiguration.cs:line 201
   at k8s.KubernetesClientConfiguration.GetKubernetesClientConfiguration(String currentContext, String masterUrl, K8SConfiguration k8SConfig) in k8s\KubernetesClientConfiguration.cs:line 175
   at k8s.KubernetesClientConfiguration.<BuildConfigFromConfigFileAsync>d__19.MoveNext() in k8s\KubernetesClientConfiguration.cs:line 138
   at k8s.KubernetesClientConfiguration.BuildDefaultConfig() in k8s\KubernetesClientConfiguration.cs:line 110
   at simple.PodList.<Main>d__1.MoveNext() in C:\Users\rsrivastava\source\repos\KubernetesClient\Program.cs:line 46
[...]

PR #1457, opened 2023-11-18
Fix: Exec provider potential deadlock

avin3sh:
The current use of `WaitForExit()` _before_ trying to read standard output stream is problematic and may cause deadlock scenario. `WaitForExit()` should be generally called after all other methods have been called on Process ([ref](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.process.waitforexit?view=netstandard-2.0)).

In existing scenario, if the child process has already written enough output to the stream, the parent process will be keep on waiting for the child process to exit and the child process would not exit as the stream has not been read from completely.

### DIFF
--- a/src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
+++ b/src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
@@ -553,6 +553,12 @@ namespace k8s
                     throw new KubeConfigException("external exec failed due to timeout");
                 }
 
+                // Force flush the output buffer to avoid case of missing data
+                if (ExecTimeout != Timeout.InfiniteTimeSpan)
+                {
+                    process.WaitForExit();
+                }
+
                 var responseObject = KubernetesJson.Deserialize<ExecCredentialResponse>(output.ToString());
 
                 if (responseObject == null || responseObject.ApiVersion != config.ApiVersion)
--- a/src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
+++ b/src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
@@ -537,12 +537,13 @@ namespace k8s
 
             try
             {
-                if (!process.WaitForExit((int)(ExecTimeout.TotalMilliseconds)))
+                var responseObject = KubernetesJson.Deserialize<ExecCredentialResponse>(process.StandardOutput.ReadToEnd());
+
+                if (!process.WaitForExit((int)ExecTimeout.TotalMilliseconds))
                 {
                     throw new KubeConfigException("external exec failed due to timeout");
                 }
 
-                var responseObject = KubernetesJson.Deserialize<ExecCredentialResponse>(process.StandardOutput.ReadToEnd());
                 if (responseObject == null || responseObject.ApiVersion != config.ApiVersion)
                 {
                     throw new KubeConfigException(
--- a/src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
+++ b/src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
@@ -5,6 +5,7 @@ using System.Diagnostics;
 using System.Net;
 using System.Runtime.InteropServices;
 using System.Security.Cryptography.X509Certificates;
+using System.Text;
 
 namespace k8s
 {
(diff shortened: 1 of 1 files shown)


IvanJosipovic:
Hey,

I think this PR might cause some issues. The ReadToEnd() call is blocking and will not end if only an STD Error is returned as the timeout is ignored.

See this comment and thread on why this was built this way, https://github.com/kubernetes-client/csharp/pull/1267#discussion_r1176035653

avin3sh:
Let me go through some test scenarios and get back with any tweaks if required. Exec hanging is what motivated this PR :)

avin3sh:
> The ReadToEnd() call is blocking and will not end if only an STD Error is returned as the timeout is ignored

I think I understood the concern pointed here. The TimeOut was indeed not being honored. @IvanJosipovic what do you think of the current design ? I had read the stream asynchronously, I don't like the idea of using global event handler here but we are already doing it for exec error, so did not want to deviate by introducing thread. 

The workflow diagram in https://github.com/kubernetes-client/csharp/pull/1267#discussion_r1176035653 was very handy and I looked at the KubeUI usage to understand the requirement better :)

rrudduck:
I can confirm that these changes fix the deadlock issue I was experiencing. Prior to the changes in this PR, KubernetesClientConfiguration.BuildDefaultConfig() would hang and eventually throw an exception. With these changes, it works as expected.

avin3sh:
@IvanJosipovic  FWIW, I have been using changes from this PR extensively for past couple of days, and I haven't run into the issue. The linked runtime issue does suggest "correct" way to do this - i.e. calling the other overload afterwards. I will (1) verify this is indeed a problem here, and if it so (2) include the additional changes and verify the stability.

There is also https://github.com/dotnet/runtime/issues/32456 which I will ensure we don't step into from my changes

Let me know if you have any other concerns.

avin3sh:
@tg123 @brendandburns @IvanJosipovic and anyone following

I am done with my changes and have been using it locally for a while - I haven't faced any problems and the issue with Exec freezing until the timeout has gone away. I do not have any more changes to make.

src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs
```diff
@@ -537,12 +538,29 @@ public static ExecCredentialResponse ExecuteExternalCommand(ExternalExecution co
...
+                if (!process.WaitForExit((int)ExecTimeout.TotalMilliseconds))
                 {
                     throw new KubeConfigException("external exec failed due to timeout");
                 }
 
-                var responseObject = KubernetesJson.Deserialize<ExecCredentialResponse>(process.StandardOutput.ReadToEnd());
+                // Force flush the output buffer to avoid case of missing data
+                if (ExecTimeout != Timeout.InfiniteTimeSpan)
+                {
+                    process.WaitForExit();
```
tg123, line 559: any detail here?
got confused? why wait again

avin3sh: This is necessary when `WaitForExit(Int32)` has been called with async event handlers - an additional call to `WaitForExit()` overload is required to ensure all the processing has completed else output data can be incomplete.

See:  https://github.com/dotnet/runtime/issues/18789#issuecomment-265359492
and https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.process.waitforexit?view=netstandard-2.0#system-diagnostics-process-waitforexit

>This overload ensures that all processing has been completed, including the handling of asynchronous events for redirected standard output. You should use this overload after a call to the [WaitForExit(Int32)](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.process.waitforexit?view=netstandard-2.0#system-diagnostics-process-waitforexit(system-int32)) overload when standard output has been redirected to asynchronous event handlers.

tg123:
overall LGTM
@IvanJosipovic @brendanburns any more suggestions?

avin3sh:
> @IvanJosipovic @brendanburns any more suggestions?

Happy new year all! 😄 Friendly ping for comments/approval 🤞

avin3sh:
I am using my local build of the kubernetes client for monitoring our Windows Kubernetes Cluster. Our dependency is increasing and it would be nice to see this fix in a production release. I also believe this fix potentially resolves https://github.com/kubernetes-client/csharp/issues/1346.
@tg123 please let me know if there is anything else needed in my PR :)

tg123:
@avin3sh lets allow @IvanJosipovic @brendanburns more time to approve

PR #1497, opened 2024-01-13
introduce slngen and remove stale proj

tg123:
* use slngen to manage projects, this would ensure all code under exmaple to be built
 * remove stale project


roblappcoupa:
I am confused as to why there is no `.sln` file. How are we supposed to open the source in an IDE? On a fresh clone on a Windows 10 machine with .NET 8.0, I ran
```
dotnet msbuild /t:slngen
```
but I get errors that the `slgen` target isn't found. Because there is no documentation on how to customize the HttpClient(s) used by this library and I need to figure out how to replace my old Microsoft.Rest Interceptor, I have to try and go through the code so I can figure out how to try and do this.

PR #1509, opened 2024-01-30
Let `Microsoft.VisualStudio.SlnGen` dependency development only

WeihanLi:
This seemed should be a development dependency only, currently, it would be a dependency for the packages

image

Configure `Microsoft.VisualStudio.SlnGen` `PrivateAssets` to let it be development only

https://learn.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files#controlling-dependency-assets

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -38,6 +38,6 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.VisualStudio.SlnGen" Version="11.1.0" />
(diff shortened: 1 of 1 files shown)


tg123 (approved):
/LGTM

rwkarg:
When should a release with this change be expected to be available?

tg123:
pushed 13.0.23

WeihanLi:
> pushed 13.0.23

@tg123 seemed the nuget package not pushed

PR #1529, opened 2024-02-28
fix docfx build

WeihanLi:
use `dotnet build -c Release` for the build step

action run: https://github.com/WeihanLi/kubernetes-client-csharp/actions/runs/8074136853

### DIFF
--- a/.github/workflows/docfx.yaml
+++ b/.github/workflows/docfx.yaml
@@ -38,7 +38,7 @@ jobs:
           8.0.x
 
     - name: Build
-      run: dotnet build
+      run: dotnet build -c Release
 
     - uses: nunit/docfx-action@v3.1.0
       name: Build Documentation


.github/workflows/docfx.yaml
```diff
@@ -34,11 +34,11 @@ jobs:
       with:
         dotnet-version: |
           6.0.x
-          7.0.x
+          7.0.x
```
WeihanLi, line 37: no idea why this diff, seemed no changes, just updated the following build

tg123:
not due to docfx?
last time i checked the pipeline, it was caused by net6 not supported

tg123 (approved):
/LGTM

WeihanLi:
> not due to docfx?

Seeing this warning, tried with `-c Release`

https://github.com/kubernetes-client/csharp/actions/runs/8064137407/job/22027395641

image

PR #1530, opened 2024-02-29
SlnGen instructions are potentially incomplete

### LINKED ISSUE #1528: SlnGen instructions are potentially incomplete
**Describe the bug**
Fresh pull of the repository is unable to generate a solution

**To Reproduce**

Forked and pulled master branch.
`dotnet msbuild /t:slngen`
--> error : You must specify the path of a project or directory containing just a project in order to generate a solution.

**Expected behavior**
Code can be loaded in Visual Studio to enable contributions.

Visual Studio

**Additional context**
It's unclear if any additional steps are required to create a solution and start contributing to this project. 
The current SlnGen dependency and the sparse documentation on how to use this tooling seems to be limiting the potential contribution base for this project.

I'm not immediately clear on the need for SlnGen given the relatively small number of projects and what appears to be a single solution that would be generated anyway. SlnGen appears to be a tool for very large code bases with very many projects, but with features that only require some small portion of those projects loaded to work on them. That does not appear to be the case with this code base.

tg123:
Fixes #1528

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ dotnet add package KubernetesClient
 ## Generate with Visual Studio
 
 ```
-dotnet msbuild /t:slngen
+dotnet msbuild /Restore /t:SlnGen kubernetes-client.proj
 ```
 
 ## Authentication/Configuration


rwkarg:
I can confirm that the new documented instructions work for me.

Issue #1525, opened 2024-02-17
Prometheus-net is an unnecessary dependency


rwkarg:
**Describe the bug**
This package references prometheus-net for metrics but that is not a core function of this package.
This can lead to problems where a user has an older version of Prometheus server running on a version that can't accept metrics from the version of prometheus-net reference by this package. This has happened with an older (but not too old) version of Promethus and prometheus-net referenced in KubernetesClient 8.x. The new prometheus-net version includes exemplars which leads to the Prometheus server not accepting the published metrics. Users can't force using an older version of prometheus-net explicitly because it is binary incompatible with the version KubernetesClient was built with and will lead to runtime exceptions and crashing.

Ideally, there would have been a separate package for KubernetesClient.Prometheus that could optionally pulled in to get metrics.

Now however, there are .NET Metrics available without needing to pull in any additional dependency (other than the System package System.Diagnostics.DiagnosticSource) and then users of KubernetesClient are free to use anything that integrates with that common interface (`dotnet-counters`, OpenTelemetry w/ Prometheus exporter, OpenTelemetry w/ OTLP exporter, directly with a MeterListener). This allows for a no-dependency way to collect metrics and a common interface to achieve the existing behavior (Prometheus exposition) or a variety of other exports that the users' applications are already using. This also means that the user is in control of the version of any additional, non-core dependencies like exporting metric data. Additionally, users that do not have a framework to store metrics would not need to pull in any unnecessary packages when all they want to do is work with a Kubernetes cluster.

Example impact:
**Kubernetes C# SDK Client Version**
e.g. `8.x`

**To Reproduce**
Run a Prometheus 2.42.0 server
https://github.com/prometheus/prometheus/releases/tag/v2.42.0
Run a test app using KubernetesClient 8.x+ with any exposed metrics
Scrape metrics from the test app

**Expected behavior**
Prometheus metrics should be stored by the Prometheus server

**Actual behavior**
No metrics are stored by the Prometheus server because that version of Prometheus server does not accept metrics that includes exemplars, but the version of prometheus-net that is required to be used by KubernetesClient 8.x+ includes exemplars in the data it exposes.

rwkarg:
Related to https://github.com/kubernetes-client/csharp/discussions/898

rwkarg:
It actually looks like this is not used by default, but an optional piece of functionality that needs to be opted in to. But it still always has the `prometheus-net` reference.
Add a Prometheus handler. #591 

[...]

PR #1534, opened 2024-03-04
Adding back PrometheusHandler to avoid breaking change

rwkarg:
This adds a PrometheusHandler implementation to mimic the prior prometheus-net based implementation.

Note that the old implementation incorrectly created some labels/tags with a name of "method" but the value as the Verb of the request. This incorrect behavior is maintained and noted.

The end result is that the same named metrics will be produced. The prometheus-net library reads values from System.Diagnostics.Metrics already so these will be picked up by anyone that is already using this handler.

Ultimately, users should migrate to using the built in HttpClient metrics that covers these custom metrics, uses standard semantic conventions for metric names, and has the correct labels/tags added to metrics.

### DIFF
--- a/src/KubernetesClient/KubernetesClient.csproj
+++ b/src/KubernetesClient/KubernetesClient.csproj
@@ -10,6 +10,7 @@
     </PropertyGroup>
 
     <ItemGroup>
+        <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="7.0.0" />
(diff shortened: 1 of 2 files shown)


rwkarg:
@brendandburns When you get a chance can you see if this addresses the concerns of [#1525?](https://github.com/kubernetes-client/csharp/issues/1525#issuecomment-1965538653) ?

brendandburns:
/lgtm
/approve

Thanks!

Issue #1537, opened 2024-03-05
KubernetesClientConfiguration.LoadKubeConfig is not thread-safe due to YamlDotNet Deserializer


nathanwoctopusdeploy:
**Describe the bug**

`KubernetesClientConfiguration.LoadKubeConfig` is not thread-safe due to YamlDotNet Deserializer being used as a static

PR with a test showing the issue https://github.com/kubernetes-client/csharp/pull/1536

```
   Exception during deserialization
   
   at YamlDotNet.Serialization.ValueDeserializers.NodeValueDeserializer.DeserializeValue(IParser parser, Type expectedType, SerializerState state, IValueDeserializer nestedObjectDeserializer)
   at YamlDotNet.Serialization.ValueDeserializers.AliasValueDeserializer.DeserializeValue(IParser parser, Type expectedType, SerializerState state, IValueDeserializer nestedObjectDeserializer)
   at YamlDotNet.Serialization.Deserializer.Deserialize(IParser parser, Type type)
   at YamlDotNet.Serialization.Deserializer.Deserialize[T](IParser parser)
   at k8s.KubernetesYaml.Deserialize[TValue](String yaml, Boolean strict) in F:\Code\kubernetes-client\src\KubernetesClient\KubernetesYaml.cs:line 207
   at k8s.KubernetesYaml.<LoadFromStreamAsync>d__11`1.MoveNext() in F:\Code\kubernetes-client\src\KubernetesClient\KubernetesYaml.cs:line 181
   at k8s.KubernetesClientConfiguration.<LoadKubeConfigAsync>d__31.MoveNext() in F:\Code\kubernetes-client\src\KubernetesClient\KubernetesClientConfiguration.ConfigFile.cs:line 639
   at k8s.KubernetesClientConfiguration.<LoadKubeConfigAsync>d__36.MoveNext() in F:\Code\kubernetes-client\src\KubernetesClient\KubernetesClientConfiguration.ConfigFile.cs:line 713
   at k8s.KubernetesClientConfiguration.LoadKubeConfig(FileInfo[] kubeConfigs, Boolean useRelativePaths) in F:\Code\kubernetes-client\src\KubernetesClient\KubernetesClientConfiguration.ConfigFile.cs:line 695
[...]

PR #1536, opened 2024-03-05
YamlDotNot is not thread-safe

nathanwoctopusdeploy:
This PR demonstrates that `KubernetesClientConfiguration.LoadKubeConfig` is not thread-safe due to YamlDotNot Deserializers.

```
   Exception during deserialization
   
   at YamlDotNet.Serialization.ValueDeserializers.NodeValueDeserializer.DeserializeValue(IParser parser, Type expectedType, SerializerState state, IValueDeserializer nestedObjectDeserializer)
   at YamlDotNet.Serialization.ValueDeserializers.AliasValueDeserializer.DeserializeValue(IParser parser, Type expectedType, SerializerState state, IValueDeserializer nestedObjectDeserializer)
   at YamlDotNet.Serialization.Deserializer.Deserialize(IParser parser, Type type)
   at YamlDotNet.Serialization.Deserializer.Deserialize[T](IParser parser)
   at k8s.KubernetesYaml.Deserialize[TValue](String yaml, Boolean strict) in F:\Code\kubernetes-client\src\KubernetesClient\KubernetesYaml.cs:line 207
   at k8s.KubernetesYaml.<LoadFromStreamAsync>d__11`1.MoveNext() in F:\Code\kubernetes-client\src\KubernetesClient\KubernetesYaml.cs:line 181
   at k8s.KubernetesClientConfiguration.<LoadKubeConfigAsync>d__31.MoveNext() in F:\Code\kubernetes-client\src\KubernetesClient\KubernetesClientConfiguration.ConfigFile.cs:line 639
   at k8s.KubernetesClientConfiguration.<LoadKubeConfigAsync>d__36.MoveNext() in F:\Code\kubernetes-client\src\KubernetesClient\KubernetesClientConfiguration.ConfigFile.cs:line 713
   at k8s.KubernetesClientConfiguration.LoadKubeConfig(FileInfo[] kubeConfigs, Boolean useRelativePaths) in F:\Code\kubernetes-client\src\KubernetesClient\KubernetesClientConfiguration.ConfigFile.cs:line 695
   at k8s.Tests.KubernetesClientConfigurationTests.<>c__DisplayClass48_0.<LoadKubeConfigShouldBeThreadSafe>g__LoadKubeConfig|3() in F:\Code\kubernetes-client\tests\KubernetesClient.Tests\KubernetesClientConfigurationTests.cs:line 691

   # INNER EXCEPTION

[...]

PR #1544, opened 2024-04-03
Lock usage of YamlDotNet Serializer and Deserializer

alxmitch:
YamlDotNet `ISerializer` / `IDeserializer` is [not thread safe](https://github.com/aaubry/YamlDotNet/discussions/844). This PR prevents concurrent use of the static serializers and deserializer in `KubernetesYaml`.

Addresses https://github.com/kubernetes-client/csharp/issues/1537

I have verified locally that this resolves the failing test in the [linked PoC](https://github.com/kubernetes-client/csharp/pull/1536) for the original issue, but have not included that test here as it didn't always reliably reproduce the issue. Keen for advice on appropriate testing around this change.

### DIFF
--- a/src/KubernetesClient/KubernetesYaml.cs
+++ b/src/KubernetesClient/KubernetesYaml.cs
@@ -12,6 +12,9 @@ namespace k8s
     /// </summary>
     public static class KubernetesYaml
     {
+        private static readonly object DeserializerLockObject = new object();
+        private static readonly object SerializerLockObject = new object();
+
         private static DeserializerBuilder CommonDeserializerBuilder =>
             new DeserializerBuilder()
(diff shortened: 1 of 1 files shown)


tg123 (approved):

leonard520:
Hi, I am hitting the issue in our environment, and it is great to see the fix. May I know when will be a new release to include this fix? Thanks!

tg123:
13.0.37 is now published

shenglol:
I'm curious, why did we opt for lock instead of ThreadLocal? We're using this library within an ASP.NET Core service that handles concurrent requests. The concern with lock is the potential performance overhead it introduces.

tg123:
@shenglol for now, it is a quick, dirty and safest solution.

could you please send a pr of suggesting optimization?

PR #1549, opened 2024-04-19
migrating to record for data objects

tg123:
https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/record

grants users move flexibility to use data objects, for example, built-in deconstruction.


tg123:
still thinking if make `Immutability` by default

Issue #1539, opened 2024-03-10
Memory Leak with aot package ListDeployments and ListStatefulSet


guillaume-chervet:
**Describe the bug**
SlimFaas is calling ListDeployments and ListStatefulSet every Second (in the futur we will use watch). I think of memory leak because memory increase without ending.
Previous version worked on production without any crash with 180Mo of RAM set and the same code. The only change is the switch to AOT packages.

If I remove calls, the memory keep very stable.

https://github.com/AxaFrance/SlimFaas/blob/main/src/SlimFaas/ReplicasSynchronizationWorker.cs
https://github.com/AxaFrance/SlimFaas/blob/main/src/SlimFaas/Kubernetes/KubernetesService.cs

All RAM goes down in the graph bellow are OOM Killed.
![image](https://github.com/kubernetes-client/csharp/assets/52236059/141a4e93-d19d-4d64-b741-08881ea50245)

**Kubernetes C# SDK Client Version**
"KubernetesClient.Aot" Version="13.0.12"

**Dotnet Runtime Version**
.NET 8

WeihanLi:
Is it possible to analyze some memory info from a dump?

Think the `HttpRequestMessage`/`HttpResponseMessage` and the `HttpContent` should be disposed, maybe we could try to add `using` for them firstly

![image](https://github.com/kubernetes-client/csharp/assets/7604648/d799a286-b16b-4f0f-a06e-d8990f261485)

guillaume-chervet:
Hi fixed it in my side by this commit https://github.com/AxaFrance/SlimFaas/commit/d71b2ba065f416ae22bd0340407c928e9226e051

But it is still a bug it seem dispose does not dispose well.

I will try to memory dump.

tg123:
yes dump please
i will try to repo by calling list

guillaume-chervet:
I had a add week,  i will try to do it next week!

tg123:
i ran the load test last weekend for whole afternoon, but did not see any leak.
code is simple

for(;;) { listdeployment; sleep 5s}

tg123:
my test code
```
using k8s;

var config = KubernetesClientConfiguration.BuildDefaultConfig();
IKubernetes client = new Kubernetes(config);
Console.WriteLine("Starting Request!");


for (; ;){
    var list = client.AppsV1.ListDeploymentForAllNamespaces();
    foreach (var item in list.Items)
    {
        Console.WriteLine(item.Metadata.Name);
[...]

PR #1552, opened 2024-04-29
Add missing Dispose() calls

brianpursley:
While not recommended, if you repeatedly create a Kubernetes client inside of a loop, there appears to be a memory leak, even if you dispose it.

This PR adds the following to fix this:

* Adds `.Dispose()` for `FirstMessageHandler`
* Adds `.Dispose()` for `HttpClientHandler`
* A few other minor changes in the `Dispose()` method
* ~~Adds `Using` for `httpRequest` (while unrelated to the Kubernetes client lifecycle, I believe this should also be disposed)~~

Possible fix for #1539 (I suspect that issue is not related only to aot).

## Test Results

I tested this using the following console program and running in Rider using memory profiling (sampled allocations):
```
using k8s;

var config = KubernetesClientConfiguration.BuildConfigFromConfigFile();

for (var i = 0; i < 1000; i++)
{
    using var client = new Kubernetes(config);
    var list = client.CoreV1.ListNamespacedPod("kube-system");
    Console.WriteLine(string.Join(", ", list.Items.Select(x => x.Metadata.Name)));
}
```

### Before the fix
Memory increases slowly, eventually reaching ~163MB by the end of the program:
BeforeFix1

### After the fix
Memory holds stead around ~123MB:
AfterFix1

### DIFF
--- a/src/KubernetesClient/Kubernetes.cs
+++ b/src/KubernetesClient/Kubernetes.cs
@@ -209,6 +209,7 @@ namespace k8s
 
                 // Dispose the client
                 HttpClient?.Dispose();
+                HttpClient = null;
 
                 // Dispose the certificates
                 if (CaCerts is not null)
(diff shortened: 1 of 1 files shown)


brianpursley:
Hmm, i think I can't do this:
```
using var httpRequest = new HttpRequestMessage
```

... because this is not awaited ...
```
return SendRequestRaw("", httpRequest, cancellationToken);
```

`SendRequest` should probably use async/await, but I think that is a different problem.  I will update this PR to remove that change.

tg123:
Good catch 

/LGTM

tg123 (approved):

PR #1554, opened 2024-04-30
#1553 optimise dependencies

### LINKED ISSUE #1553: Remove polyfil package from net 7/8
### Describe the feature.
I want my application's and libraries to be as small as possible and the minimum amount of dependencies.

#### Is your feature related to a problem? 
Additional dependencies are being added to my projects.

#### Describe the requested feature

When the framework is able to natively provide a dependency ie System.Diagnostics.DiagnosticSource that should be utilised rather than an explicit dependency. This will be the case for the net 7+ TFM.
        
#### Describe alternatives you've considered
Accept the additional dependencies

### Additional Context

_No response_

thompson-tomo:
Have gone and eliminated polyfils packages where possible as in those cases the dependencies are already natively being provided by the framework.

Closes #1553

### DIFF
--- a/src/KubernetesClient.Classic/KubernetesClient.Classic.csproj
+++ b/src/KubernetesClient.Classic/KubernetesClient.Classic.csproj
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;net48</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;net48;netstandard2.1;net6.0</TargetFrameworks>
     <RootNamespace>k8s</RootNamespace>
   </PropertyGroup>
 
@@ -12,6 +12,9 @@
     <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="8.0.0" />
     <PackageReference Include="Fractions" Version="7.3.0" />
     <PackageReference Include="YamlDotNet" Version="15.1.0" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'net48' or '$(TargetFramework)' == 'netstandard2.0' or '$(TargetFramework)' == 'netstandard2.1'">
     <PackageReference Include="System.Text.Json" Version="7.0.3" />
   </ItemGroup>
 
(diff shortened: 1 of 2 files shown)


src/KubernetesClient.Classic/KubernetesClient.Classic.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;net48</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;net48;netstandard2.1;net6.0</TargetFrameworks>
```
tg123, line 4: why netstandard2.1 and net6 here?
classic will support netstandard2.0 only

to use net6, please use main sdk

src/KubernetesClient/KubernetesClient.csproj
```diff
@@ -7,7 +7,6 @@
     </PropertyGroup>
 
     <ItemGroup>
-        <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="7.0.0" />
```
tg123, line 10: see #1534

WeihanLi: The `UpDownCounter` introduced in 7.0, `net6.0` framework reference does not have this reference, so we have to reference the package at least for `net6.0`, maybe a condition could be added like `Condition="'$(TargetFramework)'='net6.0'"`, safe to remove for `net8.0`

thompson-tomo: good pickup

src/KubernetesClient.Classic/KubernetesClient.Classic.csproj
thompson-tomo: have removed the netstandard 2.1 as didn't provide benefit.

src/KubernetesClient/KubernetesClient.csproj
WeihanLi: seemed fix not pushed

src/KubernetesClient.Classic/KubernetesClient.Classic.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;net48</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;net48;net6.0</TargetFrameworks>
```
tg123, line 4: could you please elaborate why net6 is added here?

PR #1561, opened 2024-06-07
Update KubernetesClient DiagnosticSource dependency

WeihanLi:
`System.Diagnostics.DiagnosticSource` has been included in the framework reference, no need to reference for net8

fixes https://github.com/kubernetes-client/csharp/issues/1553

### DIFF
--- a/src/KubernetesClient/KubernetesClient.csproj
+++ b/src/KubernetesClient/KubernetesClient.csproj
@@ -7,7 +7,7 @@
     </PropertyGroup>
 
     <ItemGroup>
-        <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="7.0.0" />
+        <PackageReference Condition="'$(TargetFramework)'=='net6.0'" Include="System.Diagnostics.DiagnosticSource" Version="7.0.0" />
         <PackageReference Include="System.IdentityModel.Tokens.Jwt" Version="7.1.2" />
         <PackageReference Include="IdentityModel.OidcClient" Version="5.2.1" />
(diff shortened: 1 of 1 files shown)


src/KubernetesClient/KubernetesClient.csproj
```diff
@@ -7,7 +7,7 @@
     </PropertyGroup>
 
     <ItemGroup>
-        <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="7.0.0" />
```
tg123, line 10: so this is net6 only?

we already moved to net8 lol

WeihanLi: Yeah, the `System.Diagnostics.DiagnosticSource` is included in the shared framework, so there's no need to reference this package for most cases.

While we're using the `UpDownCounter` in the code

https://github.com/kubernetes-client/csharp/blob/2c2bd6a4a388f630f9aac84cbe4b8a65f0096ee1/src/KubernetesClient/PrometheusHandler.cs#L33

The `UpDownCounter` has been introduced since .NET 7
see the API: https://apisof.net/catalog/c7f52491f5c19b75d271677fe9123886

The shared framework for .NET 6 only has the 6.0.x reference which does not support this `UpDownCounter`, so we had to add a reference explicitly.

For .NET 7 and later frameworks, we could use the framework reference directly, so no need to add the reference for `System.Diagnostics.DiagnosticSource` when targeting `net8.0` I think

tg123 (approved):
/LGTM

PR #1577, opened 2024-08-24
update `InternalsVisibleTo` usage

WeihanLi:
- update `InternalsVisibleTo` usage, use MS Build item instead
- remove unnecessary items


tg123 (approved):
/LGTM

PR #1565, opened 2024-06-12
init cpm support

WeihanLi:
Use the nuget CPM feature to manage the nuget package versions centrally

https://learn.microsoft.com/en-us/nuget/consume-packages/central-package-management

fixes #1564

PR #1580, opened 2024-09-07
fix package version mismatch

### LINKED ISSUE #1575: Nuget package is versioned differently to GitHub
**Describe the bug**
The latest version in this repo is `14.0.8`, however in NuGet, it's tagged as `14.0.9`.

e.g. `14.0.8`

NuGet: https://www.nuget.org/packages/KubernetesClient#versions-body-tab
image

GitHub: https://github.com/kubernetes-client/csharp/releases/tag/v14.0.8

tg123:
fix #1575

regression introduced by #1565

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -1,5 +1,24 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
     <ItemGroup>
+        <PackageReference Include="StyleCop.Analyzers">
+            <PrivateAssets>all</PrivateAssets>
+            <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
+        </PackageReference>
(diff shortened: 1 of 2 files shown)


WeihanLi:
any idea why this caused the breaking?

tg123:
i did git bisect, 
create a tag for each local build test, found that the pr breaks it

i have no idea why

WeihanLi:
wired that from my test, the same version produced for both release and nuget package from the same commit

- https://github.com/WeihanLi/kubernetes-client-csharp/actions/runs/10757117841/job/29830774075
- https://github.com/WeihanLi/kubernetes-client-csharp/actions/runs/10757209841/job/29830988800

WeihanLi:
@tg123 would you mind trying to trigger a release to verify whether we still have this issue, maybe we could wait until the v1.31 is ready

tg123:
sure i did the test locally, seems you are right, it is due to tag not created

PR #1588, opened 2024-09-18
Enable JSON Patch in AOT

### LINKED ISSUE #1587: KubernetesJson.Serialize fails due to null JsonSerializerInfo for V1Patch
**Describe the bug**
It looks like `SourceGenerationContext` lacks JSON serialization metadata for the `V1Patch` type, which causes patch requests to fail in AOT scenarios.

`15.0.1`

**Server Kubernetes Version**
`1.30.0`

**Dotnet Runtime Version**
`net9.0`, though this is likely reproducible across AOT compatible TFM's

**To Reproduce**
Attempt to patch some k8s config using one of the `Patch...` methods on the Kubernetes client that takes `V1Patch` while using the `Kubernetes.Aot` package.

**Expected behavior**
Configuration to update successfully.

 - OS: Windows
 - Environment: Local (Desktop, ARM Laptop)
 - Cloud: No, local cluster

**Additional context**

I've fixed this in a local version of the package by just adding a second `JsonSerializerContext` just for `V1Patch`, and in `KubernetesJson.Serialize` checking both contexts for type info ([commit](https://github.com/kubernetes-client/csharp/compare/master...hwoodiwiss:csharp:fix-v1-patch)):
```csharp
var info = SourceGenerationContext.Default.GetTypeInfo(value.GetType()) ?? LocalSgContext.Default.GetTypeInfo(value.GetType());
```

I appreciate this is a suboptimal solution. Unfortunately I get build errors adding `V1Patch` to `SourceGenerationContext`.

hwoodiwiss:
Updates `KubernetesJson.Serialize` with a special case for seializing `V1Patch` when the content is string, as based on the examples and the converter implementation, the expectation in this case is that the conent already be JSON.

fixes #1587

### DIFF
--- a/src/KubernetesClient.Aot/KubernetesJson.cs
+++ b/src/KubernetesClient.Aot/KubernetesJson.cs
@@ -91,6 +91,11 @@ namespace k8s
 
         public static string Serialize(object value, JsonSerializerOptions jsonSerializerOptions = null)
         {
+            if (value is V1Patch { Content: string jsonValue })
+            {
+                return jsonValue;
+            }
+
             var info = SourceGenerationContext.Default.GetTypeInfo(value.GetType());
             return JsonSerializer.Serialize(value, info);
         }
--- a/examples/patch-aot/Program.cs
+++ b/examples/patch-aot/Program.cs
@@ -1,0 +1,32 @@
+using k8s;
+using k8s.Models;
+
+var config = KubernetesClientConfiguration.BuildConfigFromConfigFile();
+IKubernetes client = new Kubernetes(config);
+Console.WriteLine("Starting Request!");
+
+var pod = client.CoreV1.ListNamespacedPod("default").Items.First();
+var name = pod.Metadata.Name;
+PrintLabels(pod);
+
+var patchStr = @"
+{
+    ""metadata"": {
+        ""labels"": {
+            ""test"": ""test""
+        }
+    }
+}";
+
+client.CoreV1.PatchNamespacedPod(new V1Patch(patchStr, V1Patch.PatchType.MergePatch), name, "default");
+PrintLabels(client.CoreV1.ReadNamespacedPod(name, "default"));
+
+static void PrintLabels(V1Pod pod)
+{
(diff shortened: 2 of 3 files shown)


src/KubernetesClient.Aot/KubernetesJson.cs
```diff
@@ -91,6 +91,11 @@ public static TValue Deserialize<TValue>(Stream json, JsonSerializerOptions json
 
         public static string Serialize(object value, JsonSerializerOptions jsonSerializerOptions = null)
         {
+            if (value is V1Patch { Content: string jsonValue })
```
tg123, line 94: i did not get it, is it still some dynamic thing?

hwoodiwiss: I don't quite understand the question?

In the usages I've seen, the Content property of the V1Patch object is always set as the json string value of the patch request body. This pattern, broken down, checks:

value is of type V1Patch, and is not null
value.Content is of type string
Then it assigns the value of value.Content to jsonValue, however this is all essentially just syntactic sugar, so all of that is resolved at compile time into a slightly uglier string of conditions.

tg123: let me ask this way?
is it a breaking change or different behavior from non-aot version?
the assumption is that V1Patch must be json string?

hwoodiwiss: Ahhh, got you, this isn't a breaking behaviour, and I would say it's restoring some functionality that exists in the non-aot version to the aot version. 

I don't think full parity can be achieved without changing the API surface.

tg123: could you please also provide an example of aot json, thne i think i am good with this pr

hwoodiwiss: I've added that, mostly copying the existing patch example

tg123 (approved):

PR #1603, opened 2024-12-30
generate api v1.32

tg123:


tg123:
not working with https://github.com/kubernetes-client/gen/pull/269

PR #1609, opened 2025-01-20
Actually assign loaded cert when generating pfx

rmja:
Fix regression in .net 9 when migrating to X509CertificateLoader.

### DIFF
--- a/src/KubernetesClient/CertUtils.cs
+++ b/src/KubernetesClient/CertUtils.cs
@@ -81,7 +81,7 @@ namespace k8s
                 if (config.ClientCertificateKeyStoreFlags.HasValue)
                 {
 #if NET9_0_OR_GREATER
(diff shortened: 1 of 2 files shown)


rmja:
I found that the test that would have caught this was being skipped for Windows. This is now fixed. I believe that the cla is signed, right?

tg123:
/LGTM

anyway to make this covered in ut?

tg123 (approved):

tg123 (approved):

rmja:
> /LGTM
> 
> anyway to make this covered in ut?

This is covered now in the tests! The problem is that the tests that would have detected this were disabled on Windows, and the issue only shows on Windows. 

Looking forward to a release with this fix, as Windows compatability is currently broken.

PR #1621, opened 2025-04-12
Add tests to #1618

brendandburns:
I couldn't push a commit to  #1618 directly. @tg123 feel free to take my commit and add it to your PR and I will close this one. Or alternately we can merge this one.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -29,6 +29,7 @@
     <PackageVersion Include="System.Reactive" Version="6.0.1" />
     <PackageVersion Include="System.Text.Json" Version="9.0.0" />
     <PackageVersion Include="Vecc.YamlDotNet.Analyzers.StaticGenerator" Version="16.3.0" />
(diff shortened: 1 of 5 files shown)


tg123:
seems the c# sdk had a wrong oidc impl
`refresh_token` must be set to use oidc auth type
https://github.com/kubernetes-client/csharp/blob/675de3da2ada6dbc9d7870a29829036781b73c91/src/KubernetesClient/KubernetesClientConfiguration.ConfigFile.cs#L403

so it is likely no one ever use this feature.
the current one follows same impl with refresh_token support only.


what about removing oidc support and related dependencies?
should align with go-client imho

tg123:
/LGTM

lets merge this first to remove dependency, as it at least does not introduce any regression

brendandburns:
Sounds good to remove the dependency.

tg123:
nuget cred expired :) @brendandburns

PR #1625, opened 2025-04-22
Remove IdentityModel package versions from Directory.Packages.props

eerhardt:
These shouldn't be needed anymore after #1621 

cc @tg123


tg123 (approved):
/LGTM

Issue #1622, opened 2025-04-16
Proposal: Improve KubernetesClient API Organization for Better Discoverability


ayrloong:
## Background & Problem Statement

While working with the KubernetesClient (.NET client library), I noticed that all API methods are currently grouped under a single flat hierarchy. This design makes it challenging to discover specific methods, especially when working with different resource types across API groups/versions.

For example, in the official Go client ([client-go](https://github.com/kubernetes/client-go)), methods are logically organized with clear resource hierarchy:
```go
clientset.CoreV1().Pods("default").Get(context.TODO(), "example-xxxxx", metav1.GetOptions{})
```

However, in the current .NET implementation, similar operations lack this intuitive grouping:
```csharp
// Current structure
var pods = kubernetes.CoreV1.ListNamespacedPodWithHttpMessagesAsync(namespace);
var services = kubernetes.CoreV1.ListServiceForAllNamespacesWithHttpMessagesAsync(namespace);
```

This flat structure leads to:
- Difficulty discovering related methods
- Long method lists in IDE auto-completion
- Inconsistent with Kubernetes API grouping conventions

## Proposed Improvement

**Adopt a hierarchical organization pattern similar to client-go's Clientset structure:**

```csharp
// Proposed structure
var pods = await kubernetes.CoreV1.Pods.ListAsync(namespace);
var deployments = await kubernetes.AppsV1.Deployments.ListAsync(namespace);
```

## Key Benefits
1. **Improved Discoverability**  
   Methods are grouped by API group/version and resource types

2. **Consistency**  
   Aligns with Kubernetes API structure and client-go patterns

3. **Type Safety**  
   Stronger typing through dedicated resource classes

4. **Scalability**  
   Easier to maintain and extend for future API versions

## Implementation Approach

1. **Hierarchical Structure:**
   ```
   KubernetesClient
   ├── CoreV1
   │   ├── Pods
   │   ├── Services
   │   └── Namespaces
   ├── AppsV1
   │   ├── Deployments
   │   └── StatefulSets
   └── NetworkingV1
       ├── Ingresses
       └── NetworkPolicies
   ```

2. **Backward Compatibility**  
   Maintain existing flat methods (marked obsolete) while introducing new hierarchy

3. **Resource-Specific Classes**  
   Create dedicated classes for each resource type with relevant operations:
   ```csharp
   public class PodOperations {
       Task<List<Pod>> ListAsync(string namespace);
[...]

PR #1627, opened 2025-04-28
Implementing a more modular API

ayrloong:
This is the basic implementation based on this issue https://github.com/kubernetes-client/csharp/issues/1622

@tg123 Can you help me check if the overall structure is in compliance?


examples/clientset/Program.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    internal class Program
+    {
+        private static async Task Main(string[] args)
+        {
+            var config = KubernetesClientConfiguration.BuildConfigFromConfigFile();
+            IKubernetes client = new Kubernetes(config);
+
+            ClientSet clientSet = new ClientSet(client);
+            var list = await clientSet.CoreV1.Pod.ListNamespacedPodAsync("default").ConfigureAwait(false);
```
tg123, line 16: ListAsync?

i prefer the style in your proposal, it is simpler

src/KubernetesClient/Kubernetes.WebSocket.cs
```diff
@@ -18,19 +18,19 @@ public partial class Kubernetes
         /// <inheritdoc/>
         public Task<WebSocket> WebSocketNamespacedPodExecAsync(string name, string @namespace = "default",
             string command = null, string container = null, bool stderr = true, bool stdin = true, bool stdout = true,
-            bool tty = true, string webSocketSubProtol = null, Dictionary<string, List<string>> customHeaders = null,
+            bool tty = true, string webSocketSubProtocol = null, Dictionary<string, List<string>> customHeaders = null,
```
tg123, line 21: reviewed, just a typo

src/LibKubernetesGenerator/KubernetesClientSourceGenerator.cs
```diff
@@ -21,46 +21,38 @@ private static IContainer BuildContainer(OpenApiDocument swagger)
             builder.RegisterType<ClassNameHelper>()
                 .WithParameter(new NamedParameter(nameof(swagger), swagger))
                 .AsSelf()
-                .AsImplementedInterfaces()
-                ;
+                .AsImplementedInterfaces();
```
tg123, line 24: is it possible to keep the original style, 
put `;` in the last line in case something new in the future

src/LibKubernetesGenerator/templates/Client.cs.template
```diff
@@ -0,0 +1,154 @@
...
+/// </summary>
+public partial class {{name}}Client : ResourceClient
+{
+       public {{name}}Client(IKubernetes kubernetes) : base(kubernetes)
+       {
+       }
+
+   {{for api in apis }}
+    {{if IfReturnType api.operation "void"}}
+    private async Task<HttpOperationResponse> {{name}}Client_{{GetMethodName api.operation "WithHttpMessagesAsync"}}(
```
tg123, line 22: is it possible to delegate here to older api
or make older api calling newer api?

better to have single entrypoint

tg123 (commented):
adding [deprecated] to older style?

src/LibKubernetesGenerator/KubernetesClientSourceGenerator.cs
ayrloong: Already restored

examples/clientset/Program.cs
ayrloong: I used x-kuberneter-action to define the method name. Can you help me see if there is any problem with this?

ayrloong:
> adding [deprecated] to older style?

I think we can do both at the same time.

ayrloong:
> adding [deprecated] to older style?
I think it can be marked as obsolete/deprecated after a few versions

ayrloong:
Currently, the `WithHttpMessagesAsync` method remains unimplemented. Implementing it at this stage could introduce excessive complexity to the entire class. Based on current observations, `WithHttpMessagesAsync` appears to be primarily utilized by the `Watch` operation.

I've been tracking the Kubernetes OpenAPI Swagger specifications in the GitHub repository and noticed the recent addition of a `Watch` endpoint. However, the official release timeline for this feature remains unclear. If the implementation is significantly delayed, we might need to manually implement a `Watch` method as a temporary workaround.

In the latest commit, I've introduced the Humanizer library to handle pluralization transformations. Given that it appears to provide duplicate functionality with CaseExtensions, I recommend removing CaseExtensions to maintain a single source of truth for string manipulation.

https://raw.githubusercontent.com/kubernetes/kubernetes/refs/heads/master/api/openapi-spec/swagger.json
3f751ac230ed7fe95927fa275bb5863

src/KubernetesClient/ClientSets/ResourceClient.cs
```diff
@@ -0,0 +1,11 @@
+namespace k8s.ClientSets;
+
+public abstract class ResourceClient
+{
+    protected Kubernetes Client { get; }
+
+    public ResourceClient(IKubernetes kubernetes)
+    {
+        Client = (Kubernetes)kubernetes;
```
tg123, line 9: did not quite get here, 
anyway to avoid cast, why not force input `Kubernetes`?

tg123 (commented):
cloud you please also cleanup new style warnings?
i will take a deeper look at large files, please allow me more time to review 

thanks

src/KubernetesClient/ClientSets/ResourceClient.cs
ayrloong: Given that `IKubernetes` is typically registered in the DI container, I relied on dependency injection instead of manually providing `Kubernetes`

ayrloong:
> cloud you please also cleanup new style warnings? i will take a deeper look at large files, please allow me more time to review
> 
> thanks

I'm on vacation this week so might not get to these changes right away. I'll go through them all once I'm back next week. Really appreciate you taking the time to review!

ayrloong:
> cloud you please also cleanup new style warnings? i will take a deeper look at large files, please allow me more time to review
> 
> thanks

Warning code cleared

src/LibKubernetesGenerator/templates/ClientSet.cs.template
```diff
@@ -0,0 +1,23 @@
...
+// </auto-generated>
+
+namespace k8s.ClientSets;
+
+/// <summary>
+/// </summary>
+public partial class ClientSet
+{
+    {{for group in groups}}
+    public {{group}}GroupClient {{group}};
```
tg123, line 14:
```suggestion
    public {{group}}GroupClient {{group}} => {get;}
```

make this readonly

src/LibKubernetesGenerator/templates/GroupClient.cs.template
```diff
@@ -0,0 +1,24 @@
...
+
+namespace k8s.ClientSets;
+
+
+/// <summary>
+/// </summary>
+public partial class {{name}}GroupClient
+{
+    {{for client in clients}}
+    public {{client}}Client {{client}};
```
tg123, line 15:
```suggestion
    public {{client}}Client {{client}} => {get;} = new XXX;
```

no ctor required

src/LibKubernetesGenerator/ClientSetGenerator.cs
```diff
@@ -0,0 +1,104 @@
...
+                {
+                    var groupVersionKindElements = x.Operation?.ExtensionData?["x-kubernetes-group-version-kind"];
+                    var groupVersionKind = (Dictionary<string, object>)groupVersionKindElements;
+
+                    return new { Kind = groupVersionKind?["kind"] as string, Api = x };
+                });
+
+                foreach (var item in apis.GroupBy(x => x.Kind))
+                {
+                    var kind = item.Key.Pluralize();
```
tg123, line 74: how can humanizer be aligned with kubernetes plural name?

`PluralHelper`?

src/LibKubernetesGenerator/GeneralNameHelper.cs
```diff
@@ -162,5 +163,60 @@ public static string GetMethodName(OpenApiOperation watchOperation, string suffi
 
             return methodName;
         }
+
+        public static string GetActionName(OpenApiOperationDescription apiOperation, string resource, string suffix)
+        {
```
tg123, line 168: the func name is a bit confusing with GetMethodName, lets ask chatgpt to give it another name?

tg123 (commented):
Thanks, some comments

and also, could you please add some basic pod CRUD tests in E2E?

src/LibKubernetesGenerator/templates/ClientSet.cs.template
ayrloong: done

src/LibKubernetesGenerator/GeneralNameHelper.cs
ayrloong: I used GPT to generate some names, unfortunately they are terrible. I think we can change `GetMethodName` to `GetOperationId`, maybe it will be clearer.

src/LibKubernetesGenerator/templates/GroupClient.cs.template
ayrloong: Here, we need to pass on `Kubernetes` just like above

ayrloong:
> Thanks, some comments
> 
> and also, could you please add some basic pod CRUD tests in E2E?

Thank you very much for your review. I will add the test code later.

src/LibKubernetesGenerator/ClientSetGenerator.cs
ayrloong: In this line of code `https://github.com/kubernetes-client/csharp/blob/master/src/LibKubernetesGenerator/PluralHelper.cs#L45`, it filters operations only where the OperationId prefix contains `List`.

If some endpoints lack a `List` interface, they cannot generate plural forms. For example, the endpoint api/v1/namespaces/{namespace}/bindings does not have a `List` related interface, which prevents it from generating the correct plural form.

If we redefine the method to remove this filtering condition, it should resolve the issue and align the generated plurals one-to-one with Kubernetes' native plural naming conventions.
cab03ee039ad4192775128f7a177da7

tg123: could you please check here
https://github.com/kubernetes-client/csharp/blob/e6317b857d2a6358eeeb6f3fe76e5bdd46f67bcb/src/LibKubernetesGenerator/templates/ModelExtensions.cs.template#L8C93-L8C107

ayrloong: Just now I used `PluralHelper` to generate plural names again, but found that it was all lowercase, so we may not be able to convert it to `C#` camelCase.

Do you have a better idea? I tried using `PluralHelper` and getting its plural name in `OpenAPI` without success

image

ayrloong: I really think using Humanizer is a good approach. If certain names require special handling:
```csharp
var pluralExceptions = new Dictionary<string, string>
{
    { "Endpoints", "Endpoints" }
};
string Pluralize(string typeName, Dictionary<string, string> exceptions)
{
    if (exceptions.TryGetValue(typeName, out var plural))
    {
        return plural;
    }
    return typeName.Pluralize(); 
} 
```
Such exceptions can also be incorporated.

`client-go` appears to use the same approach, as seen here: https://github.com/kubernetes/code-generator/blob/ae11bdb9bcf55c40306bd2d770b55a2c74239930/pkg/util/plural_exceptions.go#L14-L37.

tg123: i wish to keep single source of truth, not opposed to humanizer

or just use singular as name

src/LibKubernetesGenerator/GeneralNameHelper.cs
```diff
@@ -162,5 +163,29 @@ public static string GetMethodName(OpenApiOperation watchOperation, string suffi
 
             return methodName;
         }
+
+        public static string GetActionName(OpenApiOperation apiOperation, string resource, string suffix)
+        {
+            var operationId = apiOperation.OperationId.ToPascalCase();
+            var replacements = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase)
+            {
+                { "Replace", "Put" },
```
tg123, line 172: may i know why mapping those names?

tg123: lets keep same naming conversion as client-go?
or
keep same as original c# naming (no mapping)?

```
type PodInterface interface {
	Create(ctx context.Context, pod *corev1.Pod, opts metav1.CreateOptions) (*corev1.Pod, error)
	Update(ctx context.Context, pod *corev1.Pod, opts metav1.UpdateOptions) (*corev1.Pod, error)
	// Add a +genclient:noStatus comment above the type to avoid generating UpdateStatus().
	UpdateStatus(ctx context.Context, pod *corev1.Pod, opts metav1.UpdateOptions) (*corev1.Pod, error)
	Delete(ctx context.Context, name string, opts metav1.DeleteOptions) error
	DeleteCollection(ctx context.Context, opts metav1.DeleteOptions, listOpts metav1.ListOptions) error
	Get(ctx context.Context, name string, opts metav1.GetOptions) (*corev1.Pod, error)
	List(ctx context.Context, opts metav1.ListOptions) (*corev1.PodList, error)
	Watch(ctx context.Context, opts metav1.ListOptions) (watch.Interface, error)
	Patch(ctx context.Context, name string, pt types.PatchType, data []byte, opts metav1.PatchOptions, subresources ...string) (result *corev1.Pod, err error)
	Apply(ctx context.Context, pod *applyconfigurationscorev1.PodApplyConfiguration, opts metav1.ApplyOptions) (result *corev1.Pod, err error)
	// Add a +genclient:noStatus comment above the type to avoid generating ApplyStatus().
	ApplyStatus(ctx context.Context, pod *applyconfigurationscorev1.PodApplyConfiguration, opts metav1.ApplyOptions) (result *corev1.Pod, err error)
	UpdateEphemeralContainers(ctx context.Context, podName string, pod *corev1.Pod, opts metav1.UpdateOptions) (*corev1.Pod, error)
	UpdateResize(ctx context.Context, podName string, pod *corev1.Pod, opts metav1.UpdateOptions) (*corev1.Pod, error)

	PodExpansion
}
```

ayrloong: I think `client-go` style is better. If we use it, we only need to map two names `Replace-> Update` `Read-> Get`. Other names are the same as `client-go`.

tg123:
/LGTM

thanks!

tg123 (approved):

PR #1639, opened 2025-07-10
Update codeql-analysis.yml

brendandburns:
Looks like server-2019 is no longer supported.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -17,7 +17,7 @@ on:
 jobs:
   analyze:
     name: Analyze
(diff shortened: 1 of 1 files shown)


tg123 (approved):
/LGTM

.github/workflows/codeql-analysis.yml
```diff
@@ -17,7 +17,7 @@ on:
 jobs:
   analyze:
     name: Analyze
-    runs-on: windows-2019
+    runs-on: windows-2022
```
WeihanLi, line 20: Is it acceptable to use `windows-latest`?

Issue #1602, opened 2024-12-19
AoT version with AKS `kubelogin` throws exception on creating configuration


dn-ark:
**Describe the bug**
Using `KubernetesClient.Aot` throws an exception when trying to create a configuration via `KubernetesClientConfiguration.BuildConfigFromConfigFile()` that utilizes AKS `kubelogin`.

Exception is
```
k8s.Exceptions.KubeConfigException: 'external exec failed due to uncaught exception: System.ArgumentNullException: Value cannot be null. (Parameter 'jsonTypeInfo')
   at System.Text.Json.ThrowHelper.ThrowArgumentNullException(String parameterName)
   at System.Text.Json.JsonSerializer.Deserialize(String json, JsonTypeInfo jsonTypeInfo)
   at k8s.KubernetesJson.Deserialize[TValue](String json, JsonSerializerOptions jsonSerializerOptions)
   at k8s.KubernetesClientConfiguration.ExecuteExternalCommand(ExternalExecution config)'
```

**Kubernetes C# SDK Client Version**
15.0.1

**Server Kubernetes Version**
1.30.6

**Dotnet Runtime Version**
.NET 9

**To Reproduce**
1. Use [kubelogin example](https://github.com/kubernetes-client/csharp/tree/master/examples/aks-kubelogin) as a reference
2. First, confirm that regular `KubernetesClient` version 15.0.1 works correct when calling `KubernetesClientConfiguration.BuildConfigFromConfigFile()`
3. Now change the package to `KubernetesClient.Aot` version 15.0.1 and try again - get an exception

**Expected behavior**
Configuration should be created successfully.

**KubeConfig**
Please refer to the linked sample but here's relevant `users` section from my config specifically:
```yaml
users:
- name: clusterUser_<name>
  user:
    exec:
      apiVersion: client.authentication.k8s.io/v1beta1
      args:
      - get-token
      - --environment
      - AzurePublicCloud
      - --server-id
      - <server_id>
      - --client-id
      - <client_id>
      - --tenant-id
      - <tenant_id>
      - --login
      - devicecode
      command: kubelogin
      env: null
      installHint: |2

        kubelogin is not installed which is required to connect to AAD enabled cluster.

        To learn more, please go to https://aka.ms/aks/kubelogin
      interactiveMode: IfAvailable
      provideClusterInfo: false
```

**Where do you run your app with Kubernetes SDK:**
 - OS: Windows 11 Pro
[...]

PR #1643, opened 2025-08-02
fix ExecCredentialResponse deserializion in aot

k7o:
- Fix for #1602 
- Made 'ExecCredentialResponse' deserializable in aot so tokens can be extracted initializing Kubernetes object


brendandburns:
This looks good to me, but I defer to @tg123 he's the AoT expert.

PR #1654, opened 2025-08-29
generate base on 1.34

tg123:
- `IBasicKubernetes` removed
 - generated code formatted for better src view in ide


tg123:
@brendandburns yes please,

PR #1657, opened 2025-09-03
fix(yaml): byte[] serialization

### LINKED ISSUE #1650: KubernetesYaml.Serialize() producing invalid Secret data by converting byte[] to plain strings
**Describe the bug**
KubernetesYaml.Serialize() currently converts byte[] fields into plain strings. This causes issues for Secret manifests, since the data field ends up containing string content instead of base64. As a result, the generated manifests can’t be applied directly with kubectl — they only round-trip cleanly if you deserialize them back through KubernetesYaml.Deserialize() first.

```
var cfg = KubernetesClientConfiguration.BuildDefaultConfig();
var client = new Kubernetes(cfg);
var secret = client.ReadNamespacedSecret("default-token-gvs7k", "default");

var yaml = KubernetesYaml.Serialize(secret);

File.WriteAllText("secret.yaml", yaml);

kubectl apply -f .\secret.yaml
Error from server (BadRequest): error when creating ".\\secret.yaml": Secret in version "v1" cannot be handled as a Secret: illegal base64 data at input byte 0
```

17.0.4

**Dotnet Runtime Version**
e.g. net9

**To Reproduce**
Serialize Secret with KuberntesYaml.Serialize and apply with Kubectl

**Expected behavior**
KuberntesYaml.Serialize should output a valid Yaml that can be applied with other Kubernetes tools.

If the strings are moved from the `data` field to `stringData` then it would be a valid Secret.

IvanJosipovic:
Fixes #1650

### DIFF
--- a/src/KubernetesClient/KubernetesYaml.cs
+++ b/src/KubernetesClient/KubernetesYaml.cs
@@ -92,18 +92,8 @@ namespace k8s
             public void WriteYaml(IEmitter emitter, object value, Type type, ObjectSerializer serializer)
             {
                 var obj = (byte[])value;
-                var strValue = Encoding.UTF8.GetString(obj);
-
-                // Check if the string is multi-line by looking for a newline character.
-                var scalarStyle = strValue.Contains('\n') ? ScalarStyle.Literal : ScalarStyle.Any;
-
-                emitter.Emit(new Scalar(
-                    AnchorName.Empty,
-                    TagName.Empty,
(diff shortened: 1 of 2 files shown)


tg123:
i wonder if this will break configmap?

tg123:
thanks for confirming, could you please also add testcase to cover configmap?

tg123 (approved):
/LGTM

Issue #1662, opened 2025-09-09
TODO list


tg123:
- [ ] remove websocket, using http2
- [x] generate WatchXXX for easier api
- [ ] remove remaining autorest package
- [ ] refactor KubernetesClientConfiguration and Kubernetes, too many duplicate fields
- [ ] ask chatgpt to impl kubectl

PR #1667, opened 2025-09-15
feat: enhance Kubernetes client with watch functionality

lqlive:
Base on https://github.com/kubernetes-client/csharp/issues/1662
 @tg123  Hi ,This is my initial implementation. Could you help me see how much it differs from your ideas?

During my implementation, I thought of two issues:

1.Can the `ListNamespacedPodWithHttpMessagesAsync` method be set as protected or private to prevent external access?

2.Replace the tuple `(WatchEventType, V1Pod)` in `IAsyncEnumerable<(WatchEventType, V1Pod)>` with a new type, such as:

``` csharp
public readonly struct ResourceEvent<TResource>(EventType eventType, TResource value, TResource oldValue = default!)
{
    public EventType EventType { get; } = eventType;
    public TResource OldValue { get; } = oldValue;
    public TResource Value { get; } = value;
}

public enum EventType
{
    Unknown = 0,
    Created = 1,
    Updated = 2,
    Deleted = 3
}
```

### DIFF
--- a/examples/watch/Program.cs
+++ b/examples/watch/Program.cs
@@ -1,5 +1,4 @@
 using k8s;
-using k8s.Models;
 using System;
 using System.Threading;
 using System.Threading.Tasks;
@@ -8,9 +7,10 @@ var config = KubernetesClientConfiguration.BuildConfigFromConfigFile();
 
 IKubernetes client = new Kubernetes(config);
 
-var podlistResp = client.CoreV1.ListNamespacedPodWithHttpMessagesAsync("default", watch: true);
+var podlistResp = client.CoreV1.WatchListNamespacedPodAsync("default");
+
 // C# 8 required https://docs.microsoft.com/en-us/archive/msdn-magazine/2019/november/csharp-iterating-with-async-enumerables-in-csharp-8
-await foreach (var (type, item) in podlistResp.WatchAsync<V1Pod, V1PodList>().ConfigureAwait(false))
+await foreach (var (type, item) in podlistResp.ConfigureAwait(false))
 {
     Console.WriteLine("==on watch event==");
     Console.WriteLine(type);
@@ -22,14 +22,24 @@ await foreach (var (type, item) in podlistResp.WatchAsync<V1Pod, V1PodList>().Co
 void WatchUsingCallback(IKubernetes client)
 #pragma warning restore CS8321 // Remove unused private members
 {
-    var podlistResp = client.CoreV1.ListNamespacedPodWithHttpMessagesAsync("default", watch: true);
-    using (podlistResp.Watch<V1Pod, V1PodList>((type, item) =>
+    using var podlistResp = client.CoreV1.WatchListNamespacedPod("default");
+    podlistResp.OnEvent += (type, item) =>
     {
         Console.WriteLine("==on watch event==");
         Console.WriteLine(type);
         Console.WriteLine(item.Metadata.Name);
         Console.WriteLine("==on watch event==");
-    }))
+    };
+    podlistResp.OnError += (error) =>
+    {
+        Console.WriteLine("==on watch error==");
+        Console.WriteLine(error.Message);
+        Console.WriteLine("==on watch error==");
+    };
+    podlistResp.OnClosed += () =>
+    {
+        Console.WriteLine("==on watch closed==");
+    };
     {
         Console.WriteLine("press ctrl + c to stop watching");
 
--- a/src/KubernetesClient.Aot/KubernetesClient.Aot.csproj
+++ b/src/KubernetesClient.Aot/KubernetesClient.Aot.csproj
@@ -89,8 +89,8 @@
     <Compile Include="..\KubernetesClient\IKubernetes.Exec.cs" />
     <Compile Include="..\KubernetesClient\Kubernetes.Exec.cs" />
 
-    <!-- <Compile Include="..\KubernetesClient\Watcher.cs" /> -->
-    <!-- <Compile Include="..\KubernetesClient\WatcherExt.cs" /> -->
+    <Compile Include="..\KubernetesClient\Watcher.cs" />
+    <Compile Include="..\KubernetesClient\WatcherExt.cs" />
     <Compile Include="..\KubernetesClient\LineSeparatedHttpContent.cs" />
 
     <Compile Include="..\KubernetesClient\Exceptions\KubeConfigException.cs" />
--- a/src/LibKubernetesGenerator/ParamHelper.cs
+++ b/src/LibKubernetesGenerator/ParamHelper.cs
@@ -3,6 +3,7 @@ using NSwag;
 using Scriban.Runtime;
 using System;
 using System.Linq;
+using System.Collections.Generic;
 
 namespace LibKubernetesGenerator
 {
@@ -21,6 +22,8 @@ namespace LibKubernetesGenerator
         {
             scriptObject.Import(nameof(GetModelCtorParam), new Func<JsonSchema, string>(GetModelCtorParam));
             scriptObject.Import(nameof(IfParamContains), IfParamContains);
+            scriptObject.Import(nameof(FilterParameters), FilterParameters);
+            scriptObject.Import(nameof(GetParameterValueForWatch), new Func<OpenApiParameter, bool, string, string>(GetParameterValueForWatch));
         }
 
         public static bool IfParamContains(OpenApiOperation operation, string name)
@@ -37,6 +40,23 @@ namespace LibKubernetesGenerator
             }
 
             return found;
+        }
+
+        public static IEnumerable<OpenApiParameter> FilterParameters(OpenApiOperation operation, string excludeParam)
+        {
+            return operation.Parameters.Where(p => p.Name != excludeParam);
+        }
+
+        public string GetParameterValueForWatch(OpenApiParameter parameter, bool watch, string init = "false")
+        {
+            if (parameter.Name == "watch")
+            {
+                return watch ? "true" : "false";
+            }
+            else
+            {
+                return generalNameHelper.GetDotNetNameOpenApiParameter(parameter, init);
+            }
         }
 
         public string GetModelCtorParam(JsonSchema schema)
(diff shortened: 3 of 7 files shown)


src/LibKubernetesGenerator/templates/OperationsExtensions.cs.template
```diff
@@ -154,5 +155,76 @@ public static partial class {{name}}OperationsExtensions
...
+public static Watcher<{{GetReturnType api.operation "T"}}> Watch{{GetOperationId api.operation ""}}(
+    this I{{name}}Operations operations,
+{{ for parameter in $filteredParams}}
+    {{GetDotNetTypeOpenApiParameter parameter}} {{GetDotNetNameOpenApiParameter parameter "true"}},
+{{end}}
+    Action<WatchEventType, {{GetReturnType api.operation "T"}}> onEvent = null,
+    Action<Exception> onError = null,
+    Action onClosed = null)
+{
+    if (onEvent == null) throw new ArgumentNullException(nameof(onEvent));
```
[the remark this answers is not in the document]
lqlive: image
Because the parameters before  onEvent have default values, onEvent must be given a default value

src/LibKubernetesGenerator/ParamHelper.cs
```diff
@@ -39,6 +42,23 @@ public static bool IfParamContains(OpenApiOperation operation, string name)
             return found;
         }
 
+        public static IEnumerable<OpenApiParameter> FilterParameters(OpenApiOperation operation, string excludeParam)
+        {
+            return operation.Parameters.Where(p => p.Name != excludeParam);
+        }
+
+        public string GetParameterValueForWatch(OpenApiParameter parameter, bool watch, string init = "false")
```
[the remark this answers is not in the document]
lqlive: Here it is just to pass `generalNameHelper.GetDotNetNameOpenApiParameter(parameter, init);`

src/KubernetesClient.Aot/KubernetesClient.Aot.csproj
```diff
@@ -89,8 +89,8 @@
     <Compile Include="..\KubernetesClient\IKubernetes.Exec.cs" />
     <Compile Include="..\KubernetesClient\Kubernetes.Exec.cs" />
 
-    <!-- <Compile Include="..\KubernetesClient\Watcher.cs" /> -->
-    <!-- <Compile Include="..\KubernetesClient\WatcherExt.cs" /> -->
+    <Compile Include="..\KubernetesClient\Watcher.cs" />
```
tg123: why this was removed?

examples/watch/Program.cs
```diff
@@ -22,19 +22,18 @@
...
-    var podlistResp = client.CoreV1.ListNamespacedPodWithHttpMessagesAsync("default", watch: true);
-    using (podlistResp.Watch<V1Pod, V1PodList>((type, item) =>
+    using var podlistResp = client.CoreV1.WatchListNamespacedPod("default", onEvent: (type, item) =>
     {
         Console.WriteLine("==on watch event==");
         Console.WriteLine(type);
         Console.WriteLine(item.Metadata.Name);
         Console.WriteLine("==on watch event==");
-    }))
+    });
```
tg123: are you sure the ; should be here?

tg123 (commented):
could you please also change e2e test and watcher related test?

and lets mark direct call to watch() deprecated, will be removed next version

src/KubernetesClient.Aot/KubernetesClient.Aot.csproj
lqlive: image  
If it's not removed, the solution can't be built

examples/watch/Program.cs
lqlive: restored ,because I removed the `using ()`

src/LibKubernetesGenerator/TypeHelper.cs
```diff
@@ -246,12 +246,14 @@ string toType()
 
                     break;
                 case "T":
-                    // Return single item type from list type (e.g., V1Pod from V1PodList)
-                    return !string.IsNullOrEmpty(t) && t.EndsWith("List", StringComparison.Ordinal)
-                        ? t.Substring(0, t.Length - 4)
-                        : t;
+                    var itemType = TryGetItemTypeFromSchema(response);
```
lqlive, line 248: image

I pulled the latest code and found that the conditions here cannot cover this `ResourceClaim`, so I changed the logic here

src/KubernetesClient.Aot/KubernetesClient.Aot.csproj
tg123: ah sorry i remember, seems watch has some issue with aot

src/LibKubernetesGenerator/templates/IOperations.cs.template
```diff
@@ -25,6 +25,9 @@ public partial interface I{{name}}Operations
     /// <param name="cancellationToken">
     /// A <see cref="CancellationToken"/> which can be used to cancel the asynchronous operation.
     /// </param>
+    {{if IfParamContains api.operation "watch"}}  
+    [Obsolete("This method will be deprecated in future versions. Please use the Watch method instead.")]
```
tg123: sorry my bad
i mean add [Obsolete] to 

watcherext.Watch(this xxx)

src/LibKubernetesGenerator/TypeHelper.cs
tg123: could you please elaborate on this? i am lost

lqlive: This method is primarily used to retrieve the object name of the Item within a List object. Previously, I used a substring approach to extract the object name of the List. However, this approach encounters issues under the `ResourceClaim` resource because the generated Item object name is `Resourcev1ResourceClaim`, making the substring method less suitable in this case.

src/LibKubernetesGenerator/templates/IOperations.cs.template
lqlive: done

src/KubernetesClient.Aot/KubernetesClient.Aot.csproj
lqlive: So is there anything else that needs to be done here?

src/LibKubernetesGenerator/TypeHelper.cs
lqlive: I haven't come up with a better method for this for the time being. Do you have any better suggestions?

src/KubernetesClient.Aot/KubernetesClient.Aot.csproj
tg123: there is a `<DefineConstants>$(DefineConstants);K8S_AOT</DefineConstants>`
lets skip watch in generator code
`#if !K8S_AOT`

lqlive: Done, thanks

src/LibKubernetesGenerator/TypeHelper.cs
tg123: got it thanks

src/KubernetesClient/WatcherExt.cs
```diff
@@ -71,6 +73,7 @@ public static Watcher<T> Watch<T, L>(
         /// <param name="onError">a callback when any exception was caught during watching</param>
         /// <param name="cancellationToken">cancellation token</param>
         /// <returns>IAsyncEnumerable of watch events</returns>
+        [Obsolete("This method will be deprecated in future versions.")] 
         public static IAsyncEnumerable<(WatchEventType, T)> WatchAsync<T, L>(
             this Task<HttpOperationResponse<L>> responseTask,
```
tg123, line 78: i think we should not filter `bool watch` in list at the moment
otherwise, this WatchExt cant work anymore for old code

so what do you think, make those `internal` or keep compatible until next version?

src/LibKubernetesGenerator/TypeHelper.cs
```diff
@@ -283,5 +293,26 @@ public static bool IfType(JsonSchemaProperty property, string type)
...
+            if (listSchema?.Properties?.TryGetValue("items", out var itemsProperty) != true)
+            {
+                return null;
+            }
+
+            if (itemsProperty.Reference != null)
+            {
+                return classNameHelper.GetClassNameForSchemaDefinition(itemsProperty.Reference);
+            }
+
```
tg123: could you please fix those warnings

tg123 (commented):
overall lgtm

lets close on weather hide `bool watch` or `watcherext`, then we can merge

src/KubernetesClient/WatcherExt.cs
lqlive: I think it's best to just change it to `internal` so we don't forget to remove it in a future version.

src/LibKubernetesGenerator/TypeHelper.cs
lqlive: Done

src/KubernetesClient/WatcherExt.cs
tg123: cc @brendanburns  for ideas
`internal` it will create a big breaking change and users have to to migrate their code to new `WatchXXX `
do you think if it is nessary to wait one more big version?

lqlive: > cc @brendanburns for ideas `internal` it will create a big breaking change and users have to to migrate their code to new `WatchXXX ` do you think if it is nessary to wait one more big version?

It seems that @ is wrong. It should be @brendandburns

tg123: ping @brendandburns

brendandburns: I would prefer to keep it compatible for one version to give people a chance to migrate and then we can remove/update.

Does that work?

lqlive: Restored, thanks

PR #1701, opened 2025-12-11
feat: add non-generic Deserialize method to KubernetesJson

kimpenhaus:
this PR adds a non-generic Deserialise method to `KubernetesJson.cs`

**Reason:**

based on this [discussion](https://github.com/kubernetes-client/csharp/discussions/1683) adding a non generic `Deserialize` method gives the opportunity to use the same json serialisation mechanism everywhere - when using KubernetesClient package.

### DIFF
--- a/src/KubernetesClient/KubernetesJson.cs
+++ b/src/KubernetesClient/KubernetesJson.cs
@@ -4,6 +4,7 @@ using System.Text.RegularExpressions;
 using System.Xml;
 
 #if NET8_0_OR_GREATER
+using System.Diagnostics.CodeAnalysis;
 using System.Text.Json.Serialization.Metadata;
 #endif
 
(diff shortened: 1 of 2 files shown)


tg123:
what about 
```
        var elm = KubernetesJson.Deserialize<JsonElement>(json);
        var result = elm.Deserialize(typeof(V1Secret));
```

i felt uncomfortable about the warning in sdk

src/KubernetesClient/KubernetesJson.cs
```diff
@@ -134,6 +134,26 @@ public static void AddJsonOptions(Action<JsonSerializerOptions> configure)
...
+        /// An object of type <paramref name="returnType"/> deserialized from the JSON string.
+        /// </returns>
+        public static object Deserialize(string json, Type returnType, JsonSerializerOptions jsonSerializerOptions = null)
+        {
+#if NET8_0_OR_GREATER
+            var info = (jsonSerializerOptions ?? JsonSerializerOptions).GetTypeInfo(returnType);
+            return JsonSerializer.Deserialize(json, info);
+#else
+            return JsonSerializer.Deserialize(json, returnType, jsonSerializerOptions ?? JsonSerializerOptions);
+#endif
```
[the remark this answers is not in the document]
kimpenhaus: updated the title as that was outdated - my fault.

tests/KubernetesClient.Tests/KubernetesJsonTests.cs
```diff
@@ -201,4 +201,27 @@ public void DateTimeNonRfc3339FormatIsRejected()
...
+                              "apiVersion": "v1",
+                              "kind": "Secret",
+                              "metadata": {
+                                "name": "typed-secret"
+                              },
+                              "type": "Opaque"
+                            }
+                            """;
+
+#pragma warning disable CA2263
```
tg123, line 219: could you please help me understand, the api will result a warning? user must disable it manually?

kimpenhaus: @tg123 [CA2263](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2263) `Prefer generic overload when type is known` would cause a warning - but we want to test explicitly the non-generic overload here. that's why the warning needs to be suppressed with this test.

kimpenhaus:
this is like 2 lines of code (beside the unit-test) which performance way better then doing a double deserialization like suggested [here](https://github.com/kubernetes-client/csharp/pull/1701#issuecomment-3692451471) - there is no impact on existing code

it's kind of disappointing...

PR #1799, opened 2026-06-08
Add Windows e2e tests via kind-in-WSL

tg123:
Adds a new `e2e classic (wsl)` CI job that runs the `KubernetesClient.Classic` (.NET Framework 4.8) end-to-end tests against a real single-node Kubernetes cluster on Windows.

**Approach**
- The `kindest/node` image (which bundles containerd, the kubelet, and the K8s control-plane components) is imported into WSL2 as a systemd-enabled distro using `tg123/oci-to-wsl` (SHA-pinned).
- `.github/e2e/bootstrap-kind.sh` brings up the single-node control plane inside the WSL distro: it sets up cgroupfs (no systemd cgroup delegation needed), patches out the kindest/node kubelet drop-in that assumes a Docker cgroup namespace, runs `kubeadm init`, untaints the node, and applies the bundled kindnet CNI.
- The Windows host obtains a kubeconfig pointing at the WSL VM's IP. Because .NET Framework 4.8 does not support custom trust stores, `certificate-authority-data` is replaced with `insecure-skip-tls-verify` (CI-only ephemeral cluster).
- Only `tests/E2E.Classic.Tests` (net48) runs in this job. The existing Linux `e2e` job continues to cover the modern (net8/9/10) `E2E.Tests` and `E2E.Aot.Tests` suites.

**Job requirements**
- `runs-on: windows-2025` (WSL2 only ships on Windows Server 2025; `windows-latest` may still be Server 2022 with WSL1).

**New test project**
- `tests/E2E.Classic.Tests` mirrors the pattern used by `tests/E2E.Aot.Tests`: a thin csproj that links the shared `MinikubeFactAttribute.cs` and `Onebyone.cs` and ships its own minimal `MinikubeTests.cs` (the full E2E.Tests suite depends on packages that don't support net48, such as `LeaderElection`, `JsonPatch.Net`, and `Kubectl`).
- The project is excluded from the root traversal build on non-Windows runners.

The CLA-signing commit history was preserved by squashing into a single commit authored by `Boshi LIAN <bolian@microsoft.com>`.


Minikube.runsettings
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
```
tg123: @copilot I don't think this file is necessary

testEnvironments.json
```diff
@@ -0,0 +1,17 @@
+﻿{
+  "version": "1",
```
tg123: @copilot Why this file? rm?

